PR #11, opened 2017-10-06
api: add seeding endpoint

diegodelemos:
Signed-off-by: Diego Rodriguez <diego.rodriguez@cern.ch>

### DIFF
--- a/reana_server/rest/analyses.py
+++ b/reana_server/rest/analyses.py
@@ -21,8 +21,6 @@
 
 """Reana-Server Ping-functionality Flask-Blueprint."""
 
-import json
-
 from flask import current_app as app
 from flask import Blueprint, jsonify, request
 
@@ -34,13 +32,13 @@ rwc_api_client = create_openapi_client('reana-workflow-controller')
 
 @blueprint.route('/analyses', methods=['GET'])
 def get_analyses():  # noqa
-    r"""Get all analyses.
+    r"""Get all current analyses in REANA.
 
     ---
     get:
-      summary: Returns list of all analyses.
+      summary: Returns list of all current analyses in REANA.
       description: >-
-        This resource return all analyses in JSON format.
+        This resource return all current analyses in JSON format.
       produces:
        - application/json
       responses:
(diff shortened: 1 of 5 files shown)


reana_server/rest/analyses.py
```diff
@@ -182,17 +180,94 @@ def create_analysis():  # noqa
...
+
+@blueprint.route('/analyses/<analysis_id>/workspace', methods=['POST'])
+def seed_analysis(analysis_id):  # noqa
+    r"""Seed analysis with files.
+
+    ---
+    post:
+      summary: Seeds the analysis workspace with the provided file.
+      description: >-
+        This resource expects a file to be placed in the analysis workspace
```
hjhsalo: Maybe 
```
This resource expects a file that is to be placed in the analysis workspace identified by the UUID analysis_id.
```
(to be placed --> is to be placed)

hjhsalo:
This is not really part of the PR but in [reana_server/api_client.py#L35](https://github.com/diegodelemos/reana-server/blob/55d43b200549c1f99959f56623a5c8be09bdffd4/reana_server/api_client.py#L35) there is:

```
spec_file_path = os.path.join(
    pkg_resources.resource_filename('reana_server',
                                    'openapi_connections'),
  spec_file)
```

It is PEP8-compliant, but I missed the `spec_file` part while skimming through the file so maybe it should changed a bit to improve readibility?
Something like this perhaps?:
```
spec_file_path = os.path.join(
                    pkg_resources.
                    resource_filename(
                        'reana_server',
                        'openapi_connections'),
                    spec_file)
```

or

```
spec_file_path = os.path.join(
            pkg_resources.
            resource_filename(
                'reana_server',
                'openapi_connections'),
            spec_file)
```

hjhsalo:
Maybe change this in [reana_server/rest/analyses.py#L115](https://github.com/diegodelemos/reana-server/blob/55d43b200549c1f99959f56623a5c8be09bdffd4/reana_server/rest/analyses.py#L115):
```
This resource is expecting JSON data with all the necessary
information to instantiate a yadage workflow.
```
(informations --> information)

hjhsalo:
In [reana_server/rest/analyses.py#L33](https://github.com/diegodelemos/reana-server/blob/55d43b200549c1f99959f56623a5c8be09bdffd4/reana_server/rest/analyses.py#L33):

```
@blueprint.route('/analyses', methods=['GET'])
def get_analyses():  # noqa
    r"""Get all analyses.
    ---
    get:
      summary: Returns list of all analyses.
      description: >-
        This resource return all analyses in JSON format.
```

Term *all analyses* is not very clear. 
But to be honest, I don't know if this is any better:
```Returns list of all analyses (running, finished, waiting) in REANA cluster.```

Or maybe it can be opened in the description part just below this one.

Are there more states for an analysis than running, finished, waiting?

PR #24, opened 2018-01-11
api: add download analysis output files endpoint

### LINKED ISSUE #21: api: add download analysis output files endpoint
Implement endpoint `GET` `/api/analyses/:analysis_uuid/workspace/outputs/:file_path` following previous PRs style (see [here](https://github.com/reanahub/reana-server/pull/17/files))
- [x] Copy last `workflow-controller` OpenAPI spec.
- [x] Use bravado client to make the call. Remember to handle exceptions properly as described in https://github.com/reanahub/reana-server/issues/18.
- [x] Test creating some files under `<analysis_dir>/outputs/` and then download some using `curl` call direct to `server`.

diegodelemos:
* (closes #21).

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -417,6 +417,73 @@
         "summary": "Seeds the analysis workspace with the provided file."
       }
     },
+    "/api/analyses/{analysis_id}/workspace/outputs/{file_name}": {
+      "get": {
+        "description": "This resource is expecting a workflow UUID and a file name to return its content.",
+        "operationId": "get_analysis_outputs_file",
+        "parameters": [
+          {
+            "description": "Required. Organization which the analysis belongs to.",
+            "in": "query",
+            "name": "organization",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "description": "Required. UUID of analysis owner.",
+            "in": "query",
+            "name": "user",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "description": "Required. analysis UUID.",
+            "in": "path",
+            "name": "analysis_id",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "description": "Required. Name (or path) of the file to be downloaded.",
+            "in": "path",
+            "name": "file_name",
+            "required": true,
+            "type": "string"
+          }
+        ],
+        "produces": [
+          "multipart/form-data"
+        ],
+        "responses": {
+          "200": {
+            "description": "Requests succeeded. The file has been downloaded.",
+            "schema": {
+              "type": "file"
+            }
+          },
+          "400": {
+            "description": "Request failed. The incoming data specification seems malformed."
+          },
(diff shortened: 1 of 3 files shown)


docs/openapi.json
```diff
@@ -417,6 +417,73 @@
...
+            "description": "Requests succeeded. The file has been downloaded.",
+            "schema": {
+              "type": "file"
+            }
+          },
+          "400": {
+            "description": "Request failed. The incoming data specification seems malformed."
+          },
+          "404": {
+            "description": "Request failed. User doesn't exist.",
```
tiborsimko: Better `file_name` does not exist for output file handling endpoints...

reana_server/rest/analyses.py
```diff
@@ -488,3 +491,94 @@ def set_analysis_status(analysis_id):  # noqa
...
+            file_name=file_name).result()
+
+        return send_file(
+            io.BytesIO(http_response.raw_bytes),
+            attachment_filename=file_name,
+            mimetype='multipart/form-data'), 200
+    except (KeyError, HTTPBadRequest) as e:
+        return jsonify({"message": str(e)}), 400
+    except HTTPError as e:
+        # we have this option, to blindly mirror responses from RWC or
```
diegodelemos: @tiborsimko @hjhsalo any opinion here? I would rather prefer to explicitly specify exceptions so you can directly check and update the `docstring`. If we just redirect what comes from RWC we would have to check RWC `openapi` spec to write the `docstring` (which is likely to not going to happen)

tiborsimko: I'd probably favour explicit over implicit...

reana_server/openapi_connections/reana_workflow_controller.json
```diff
@@ -454,6 +454,223 @@
...
+        "produces": [
+          "multipart/form-data"
+        ],
+        "responses": {
+          "200": {
+            "description": "Requests succeeded. The list of input files has been retrieved.",
+            "schema": {
+              "items": {
+                "properties": {
+                  "last-modified": {
```
hjhsalo: Not to be added for this PR, but maybe checksumming (MD5, SHA1) would be a nice improvement at some point.

reana_server/rest/analyses.py
```diff
@@ -20,9 +20,12 @@
 # submit itself to any jurisdiction.
 
 """Reana-Server analysis-functionality Flask-Blueprint."""
-from bravado.exception import HTTPForbidden, HTTPBadRequest, HTTPNotFound
+import io
+
+from bravado.exception import (HTTPBadRequest, HTTPError, HTTPForbidden,
```
hjhsalo: `HTTPError` doesn't seem to be used.

diegodelemos: indeed, thanks

PR #25, opened 2018-01-12
dockerfile: add debug flag to install wdb

diegodelemos:


Dockerfile
```diff
@@ -19,9 +19,23 @@
 # submit itself to any jurisdiction.
 
 FROM python:3.5
+
+RUN apt-get update && \
+    apt-get install -y vim emacs-nox
```
diegodelemos: do we consider by default that having some usable editor inside the containers is useful even for PROD?

Dockerfile
```diff
@@ -19,9 +19,19 @@
...
+
 ADD . /code
 WORKDIR /code
-RUN pip install -e .[all]
+
+# Debug off by default
+ARG DEBUG=false
+
+RUN if [ "${DEBUG}" = "true" ]; then pip install wdb; pip install -e .[all]; else pip install .[all]; fi;
+
```
diegodelemos: @hjhsalo @tiborsimko `FLASK_DEBUG` would be missing. Since there is no easy way to make ENV variables conditional inside a `Dockerfile` (and it is also bad practice), we can:
* Set the `FLASK_DEBUG` and any other env variable needed in the build parameter i.e.:
```console
$ docker build . --build-arg DEBUG=true -e FLASK_DEBUG=1 -t reana-server:v0.0.1
```
* Add it to the [add wdb-service for debug support](https://github.com/reanahub/reana-cluster/pull/33/files) PR since we are adding `WDB_*` env variables already (see [here](https://github.com/reanahub/reana-cluster/pull/33/files#diff-d075e9824c5d956940794b6cc18a0af4R26)).

tiborsimko: The build variable technique is fine, we can provide convenience scripts to help in this.  However, it would be good to address this in `reana-cluster` before not too long...

Dockerfile
```diff
@@ -19,9 +19,19 @@
...
+    apt-get install -y vim emacs-nox
+
 ADD . /code
 WORKDIR /code
-RUN pip install -e .[all]
+
+# Debug off by default
+ARG DEBUG=false
+
+RUN if [ "${DEBUG}" = "true" ]; then pip install wdb; pip install -e .[all]; else pip install .[all]; fi;
```
tiborsimko: What about using `requirements-dev.txt` so that we could add there `Flask-DebugToolBar` and friends if need be?

tiborsimko:
Please add `include *.txt` to `MANIFEST.in` for the new `requirements-dev.txt` file.

PR #29, opened 2018-01-22
api: add seed/list analysis code endpoints

diegodelemos:
* Includes single endpoint for file seeding and retrieving. Amends current
  code accordingly.

### DIFF
--- a/reana_server/rest/analyses.py
+++ b/reana_server/rest/analyses.py
@@ -216,9 +216,9 @@ def create_analysis():  # noqa
         return jsonify({"message": str(e)}), 500
 
 
-@blueprint.route('/analyses/<analysis_id>/workspace', methods=['POST'])
-def seed_analysis(analysis_id):  # noqa
-    r"""Seed analysis with files.
+@blueprint.route('/analyses/<analysis_id>/workspace/inputs', methods=['POST'])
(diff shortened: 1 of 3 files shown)


reana_server/rest/analyses.py
```diff
@@ -730,10 +894,11 @@ def get_analysis_outputs_list(analysis_id):  # noqa
               }
     """
     try:
-        response, http_response = rwc_api_client.api.get_workflow_outputs(
+        response, http_response = rwc_api_client.api.get_workflow_files(
             user=request.args.get('user'),
             organization=request.args.get('organization'),
-            workflow_id=analysis_id).result()
+            workflow_id=analysis_id,
+            input_type='output').result()
```
hjhsalo: The `get_workflow_files` endpoint is used to retrieve files of a workflow / analysis.
The parameter ` input_type` may not be the most descriptive term to specify that one wants to retrieve all files that have been created by a workflow.
Maybe `file_type` or just `type` would be more descriptive or not less confusing compared to `input_type`, but on the other hand `file_type` or `type` is kind of ambiguous. Sorry I don't have better suggestions...

diegodelemos:
Since this component has no tests, we wait until we test it (from CLI), and we merge it along with https://github.com/reanahub/reana-client/pull/58 (only when it works)

PR #38, opened 2018-03-01
api: support specifying workflow run name

hjhsalo:
* Add support for specifying a name for a workflow
  when creating the workflow.
  (closes reanahub/reana-client#70)

Signed-off-by: Harri Hirvonsalo <harri.hirvonsalo@cern.ch>

### DIFF
--- a/reana_server/rest/analyses.py
+++ b/reana_server/rest/analyses.py
@@ -28,6 +28,7 @@ from bravado.exception import HTTPError
 from flask import current_app as app
 from flask import Blueprint, jsonify, request, send_file
 
+from reana_server.utils import is_uuid_v4
 from ..api_client import create_openapi_client
 
 blueprint = Blueprint('analyses', __name__)
@@ -68,6 +69,8 @@ def get_analyses():  # noqa
               properties:
                 id:
                   type: string
+                name:
+                  type: string
                 organization:
                   type: string
                 status:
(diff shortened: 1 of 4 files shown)


reana_server/rest/analyses.py
```diff
@@ -77,24 +79,28 @@ def get_analyses():  # noqa
               [
                 {
                   "id": "256b25f4-4cfb-4684-b7a8-73872ef455a1",
+                  "name": "name_of_workflow",
```
diegodelemos: Although I guess that you probably discarded using `workflow_name` because it is the name of the parameter itself, this name sounds weird to me, what about using the same as in [`reana-client`](http://reana-client.readthedocs.io/en/latest/gettingstarted.html#create-workflow) `mytest` for one and `mytest-n` for lists?

diegodelemos:
watch out Travis ❗️

reana_server/rest/analyses.py
hjhsalo: Yes WIP

reana_server/rest/analyses.py
```diff
@@ -201,11 +216,17 @@ def create_analysis():  # noqa
         if workflow_engine not in app.config['AVAILABLE_WORKFLOW_ENGINES']:
             raise Exception('Unknown workflow type.')
 
+        workflow_name = request.args.get('workflow_name', '')
+
+        if is_uuid_v4(workflow_name):
+            raise Exception('Workflow name cannot be a valid UUIDv4.')
```
diegodelemos: I think that here we should throw a 400, reusing the message

PR #39, opened 2018-03-06
api: standardize error handling

diegodelemos:
* Standardizes the way of catching `Bravado` exceptions so messages
   are propagated through the REST APIs. (addresses reanahub/reana-client#83).

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -82,11 +82,22 @@
               "type": "array"
             }
           },
+          "400": {
+            "description": "Request failed. The incoming payload seems malformed."
+          },
+          "404": {
+            "description": "Request failed. User does not exist.",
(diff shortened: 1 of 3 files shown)


docs/openapi.json
```diff
@@ -82,6 +82,17 @@
...
             }
           },
+          "400": {
+            "description": "Request failed. The incoming payload seems malformed."
+          },
+          "404": {
+            "description": "Request failed. User does not exist.",
+            "examples": {
+              "application/json": {
+                "message": "User 00000000-0000-0000-0000-000000000000 doesn't exist."
```
dinosk: A cosmetic change I suggest is to write in all messages 'does not' and end with a period

PR #47, opened 2018-06-19
factory: remove session on appcontext teardown

diegodelemos:


diegodelemos:
@dinosk @tiborsimko passing and tested locally
```console
$ for i in $(seq 1 50); do workflow_name=`reana-client workflow create -f reana-serial.yaml` \
  && export REANA_WORKON=$workflow_name \
  && reana-client code upload code && reana-client workflow start ; done
```

Dockerfile
```diff
@@ -24,6 +24,8 @@ ENV TERM=xterm
 RUN apt-get update && \
     apt-get install -y vim-tiny
 
+RUN pip install -e git://github.com/dinosk/reana-commons.git@user-mgmt#egg=reana-commons
```
tiborsimko: `s/dinosk/reanahub/g`

tiborsimko: `user-mgmt` ... I think the necessary changes were merged to `master` everywhere?

Dockerfile
```diff
@@ -32,9 +34,7 @@ ARG DEBUG=false
 
 RUN if [ "${DEBUG}" = "true" ]; then pip install -r requirements-dev.txt; pip install -e .[all]; else pip install .[all]; fi;
 
-RUN adduser --uid 1000 --disabled-password --gecos '' reanauser && \
-    chown -R reanauser:reanauser /code
-USER reanauser
 EXPOSE 5000
 ENV FLASK_APP=/code/reana_server/app.py
-CMD ["flask", "run", "--host=0.0.0.0"]
+CMD flask users create info@reana.io &&\
```
tiborsimko: `users create` needed?

Dockerfile
diegodelemos: 👍 good catch

reana_server/config.py
```diff
@@ -38,3 +38,11 @@
     'serial'
 ]
 """Available workflow engines."""
+
+ORGANIZATIONS = ['atlas', 'alice', 'cms', 'default', 'lhcb']
```
tiborsimko: Are the organizations back?

diegodelemos: Not really. I will remove them

Dockerfile
diegodelemos: It is emulating what `workflow-controller` used to do. If we don't do it here we have to add an additional step to the installation instructions or to `reana-cluster init` cc @dinosk

reana_server/config.py
```diff
@@ -38,3 +38,11 @@
...
+
+
+ADMIN_USER_ID = "00000000-0000-0000-0000-000000000000"
```
tiborsimko: remember to change this later...

Dockerfile
tiborsimko: OK, let's keep it... We'll change things WRT user registration and login later anyway...

reana_server/config.py
diegodelemos: probably we should get it from env and inject it as a secret into the container (means PR on `reana-cluster` and docs update)? `REANA_SERVER_ADMIN_TOKEN`? cc @dinosk

reana_server/rest/users.py
```diff
@@ -0,0 +1,278 @@
...
+    """
+    try:
+        user_email = request.args.get('email')
+        token = request.args.get('token')
+        admin = Session.query(User).filter_by(id_=ADMIN_USER_ID).one_or_none()
+        if token != admin.api_key:
+            return jsonify({"message": "Action not permitted."}), 403
+        user_parameters = dict(api_key=secrets.token_urlsafe())
+        user_parameters['email'] = user_email
+        user = User(**user_parameters)
```
diegodelemos: Aren't we missing the workspace creation here? Like in L70 in `reana_server/cli.py`

PR #50, opened 2018-06-21
rest: remove register_user() method

dinosk:


reana_server/rest/users.py
```diff
@@ -128,7 +128,7 @@ def get_user():  # noqa
 
 
 @blueprint.route('/users', methods=['POST'])
-def create_user():  # noqa
+def create_user(): # noqa
```
diegodelemos: you need the two spaces for pep8

PR #56, opened 2018-07-06
cli: specify token in users create

dinosk:
Addresses https://github.com/reanahub/reana-client/issues/127
Closes https://github.com/reanahub/reana-server/issues/59

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ ENV FLASK_APP=/code/reana_server/app.py
 EXPOSE 5000
 
 CMD flask db init && \
-    flask users create info@reana.io &&\
+    flask users create_default info@reana.io &&\
     uwsgi --module reana_server.app:app \
     --http-socket 0.0.0.0:5000 --master \
     --processes ${UWSGI_PROCESSES} --threads ${UWSGI_THREADS} \
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1085,16 +1085,23 @@
         "operationId": "get_user",
         "parameters": [
           {
-            "description": "Required. The email of the user.",
+            "description": "Not required. The email of the user.",
             "in": "query",
             "name": "email",
-            "required": true,
+            "required": false,
             "type": "string"
           },
           {
             "description": "Not required. UUID of the user.",
             "in": "query",
             "name": "id_",
+            "required": false,
+            "type": "string"
+          },
+          {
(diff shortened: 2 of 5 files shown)


docs/openapi.json
```diff
@@ -1111,27 +1118,37 @@
...
-                  "type": "string"
+              "items": {
+                "properties": {
+                  "email": {
+                    "type": "string"
+                  },
+                  "id_": {
+                    "type": "string"
+                  },
+                  "token": {
```
dinosk: I left the ACCESS_TOKEN as token here, to be included in the global renaming with https://github.com/reanahub/reana-client/issues/128

Dockerfile
```diff
@@ -49,7 +49,7 @@ ENV FLASK_APP=/code/reana_server/app.py
 EXPOSE 5000
 
 CMD flask db init && \
-    flask users create info@reana.io &&\
+    flask users create_default info@reana.io &&\
```
dinosk: Note this will also need to be changed in `reana-cluster` if the CMD for the container is specified there.

reana_server/cli.py
```diff
@@ -97,3 +104,92 @@ def users_create_default(email, organization_name, id_):
...
+    '-ut',
+    '--user-token',
+    help='The REANA_ACCESS_TOKEN of the user.')
+@click.option(
+    '-t',
+    '--token',
+    default=os.environ.get('REANA_TOKEN', None),
+    help='The api key of an administrator.')
+@click.pass_context
+def create_user(ctx, email, user_token, token):
```
diegodelemos: We could create non url safe tokens:
```console
$ flask users create -e diego@cern.ch -t <admin_token> -ut []?#/?][~~~.-#
$ flask users get -t <admin_token>
ID                                     EMAIL                     TOKEN
00000000-0000-0000-0000-000000000000   info@reana.io             <admin_token>
57aa5dd0-94d3-4640-96c7-5a479d4da95b   diego@cern.ch             []?#/?][~~~.-#
```
Shall we check if the token is url safe and if it is not warn the user?

Dockerfile
diegodelemos: Present in `reana-cluster` master see https://github.com/reanahub/reana-cluster/pull/99.

dinosk: It's actually already merged: https://github.com/reanahub/reana-cluster/pull/99/files

PR #62, opened 2018-07-13
global: remove organization

dinosk:
Connects https://github.com/reanahub/reana-commons/issues/25

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -48,11 +48,6 @@ def get_workflows():  # noqa
       produces:
        - application/json
       parameters:
-        - name: organization
-          in: query
-          description: Required. Organization which the workflow belongs to.
-          required: true
-          type: string
         - name: access_token
           in: query
           description: Required. The API access_token of workflow owner.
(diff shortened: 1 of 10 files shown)


reana_server/rest/workflows.py
```diff
@@ -731,10 +678,31 @@ def upload_file(workflow_id_or_name):  # noqa
           description: >-
             Request succeeded. File successfully transferred.
           schema:
-            type: object
-            properties:
-              message:
-                type: string
+            type: file
```
diegodelemos: As mentioned in https://github.com/reanahub/reana-client/pull/134#discussion_r203789606 this returns a message with information [see](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest.py#L446).

tests/conftest.py
```diff
@@ -97,19 +97,7 @@ def app(base_app, db_engine, session):
```
diegodelemos:
```console
$ git grep -i organization tests/conftest.py
tests/conftest.py:# granted to it by virtue of its status as an Intergovernmental Organization or
tests/conftest.py:from reana_commons.models import Base, Organization, User, UserOrganization
tests/conftest.py:        'ORGANIZATIONS': ['default'],
```

PR #64, opened 2018-07-17
cli: rename REANA_TOKEN to REANA_ACCESS_TOKEN

dinosk:
Connects https://github.com/reanahub/reana-client/issues/128

### DIFF
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -54,7 +54,7 @@ def create_user_space(user_id, org):
 
 def get_user_from_token(token):
     """Validate that the token provided is valid."""
-    user = Session.query(User).filter_by(api_key=token).one_or_none()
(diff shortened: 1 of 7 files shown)


reana_server/utils.py
```diff
@@ -63,27 +63,27 @@ def get_user_from_token(token):
 def _get_users(_id, email, user_token, token):
```
diegodelemos: maybe we can rename `token` to `admin_token`, what do you think? Also in `_create_user(..)`

dinosk: Oops, should've changed it as in https://github.com/reanahub/reana-server/pull/64/files#diff-aa882553e3d2579ee378587080f3db34R133, I went with `admin_access_token` and `user_access_token` WDYT @diegodelemos ?

diegodelemos: even better 🍎

PR #65, opened 2018-07-19
api: fixing specification

Sinclert:
I found some typos in the specification of the API.

Additionally, I saw the following inconsistencies:
- Fields regarding the `Organization`, which I think is in process to be removed.
- Some examples contains the `status` = `waiting`, even if the possible states are: `running`, `failed`, `created` and `finished`

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -247,14 +247,14 @@
                   "id": "72e3ee4f-9cd3-4dc7-906c-24511d9f5ee3",
                   "name": "mytest-3",
                   "organization": "default_org",
-                  "status": "waiting",
+                  "status": "created",
                   "user": "00000000-0000-0000-0000-000000000000"
                 },
                 {
                   "id": "c4c0a1a6-beef-46c7-be04-bf4b3beca5a1",
                   "name": "mytest-4",
                   "organization": "default_org",
-                  "status": "waiting",
+                  "status": "created",
                   "user": "00000000-0000-0000-0000-000000000000"
                 }
               ]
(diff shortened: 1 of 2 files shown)


dinosk:
:+1: for the catch. We need to fix the typos in the corresponding docstrings, because the openapi specs get auto generated from them. Here for example: https://github.com/reanahub/reana-server/blob/master/reana_server/rest/workflows.py#L179

And to generate the spec afterwards: https://github.com/reanahub/reana-server/blob/master/run-tests.sh#L24
```
FLASK_APP=reana_server/app.py python ./scripts/generate_openapi_spec.py
```

The organization removal is part of https://github.com/reanahub/reana-server/pull/62/files, the status inconsistencies we indeed need to fix as well :+1:

diegodelemos:
This typos are originated here https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest.py#L86.

Normally what we do is writing API docs in the method docstrings (as you can see in the previous link) and then we do `./run-tests.sh` which behind the scenes calls this:
```console
FLASK_APP=reana_workflow_controller/app.py python ./scripts/generate_openapi_spec.py
```
Which collects docstrings and generates the OpenAPI spec. You will get a message like `OpenAPI specs differ, resolve conflict.`. A file called `temp_openapi.json` is generated, thats the new API documentation, which if you are happy you can:
```console
mv temp_openapi.json docs/openapi.json
```
Once this is done, you would have to copy it ove from `reana-workflow-controller` to here.
```console
cp ~/code/reana/reana-workflow-controller/docs/openapi.json \
   reana_server/openapi_connections/reana_workflow_controller.json
```

PR #93, opened 2018-10-04
api_client: import from reana-commons

dinosk:
* Removes code for instantiating a SwaggerClient, and imports the
  BaseAPIClient class from reana-commons for it.

Connects https://github.com/reanahub/reana/issues/93


tests/conftest.py
```diff
@@ -28,39 +28,45 @@
...
-def base_app(tmp_shared_volume_path):
+def base_app():
     """Flask application fixture."""
     config_mapping = {
+        'AVAILABLE_WORKFLOW_ENGINES': 'serial',
         'SERVER_NAME': 'localhost:5000',
         'SECRET_KEY': 'SECRET_KEY',
         'TESTING': True,
-        'SHARED_VOLUME_PATH': tmp_shared_volume_path,
+        'SHARED_VOLUME_PATH': '/tmp/test',
```
diegodelemos: This is because you need https://github.com/reanahub/reana-pytest-commons/issues/4 before right?

tests/test_views.py
```diff
@@ -22,19 +22,63 @@
...
+            res = client.get(url_for('workflows.get_workflows'),
+                             query_string={"access_token":
+                                           default_user.access_token})
+            assert res.status_code == 200
+
+
+def test_create_workflow(app, default_user):
+    """Test create_workflow view."""
+    with app.test_client() as client:
+        with patch('reana_server.rest.workflows.current_rwc_api_client',
```
diegodelemos: Could we take this patch to a fixture and then yield or to a decorator?  So we save this lines for every test that needs the mocked `httpclient` for API Clients.

tests/conftest.py
```diff
@@ -28,39 +28,45 @@
...
         'SQLALCHEMY_DATABASE_URI':
         'sqlite:///',
         'SQLALCHEMY_TRACK_MODIFICATIONS': False,
+        'COMPONENTS_DATA': COMPONENTS_DATA,
     }
-    app_ = create_app(config_mapping)
-    return app_
+    app = Flask(__name__)
+    app.config.from_mapping(config_mapping)
```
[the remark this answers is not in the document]
dinosk: The factory uses `app.config.from_object()` to load the configuration from `reana_server.config`, it'd have to be changed to `from_mapping`. It can be added as a second parameter, or we could create ConfigObjects as described in http://flask.pocoo.org/docs/1.0/config/#configuration-best-practices

PR #96, opened 2018-10-19
docs: improve sphinxcontrib-openapi rendering

roksys:
* Improves REST API docs rendering. Closes #25

Signed-off-by: Rokas Maciulaitis <rokas.maciulaitis@cern.ch>


docs/conf.py
```diff
@@ -198,3 +210,44 @@
...
+    return module_name, function_name
+
+
+def process_docstring(app, what, name, obj, options, lines):
+    """
+    Deletes unnecessary docstring, saves summary and formats a hyperlink
+    to redocs.
+    """
+    module_name,function_name = get_name(name)
+    if what != "module" and module_name in rest_api_modules:
```
diegodelemos: Shouldn't this in the case of not being a `rest_api_modules`, use the default `process_docstring`?

diegodelemos: Shouldn't this in the case of not being a `rest_api_modules`, use the default `process_docstring`?

PR #98, opened 2018-10-19
tests: views permission and rwc contract

dinosk:
* Adds tests covering user access token validation and contract
  validation for rwc REST API calls.


reana_server/rest/users.py
```diff
@@ -12,6 +12,7 @@
 import traceback
 
 from flask import Blueprint, jsonify, request
+from sqlalchemy.exc import InvalidRequestError
```
roksys: Seems to be unused import.

dinosk: indeed, thanks

PR #102, opened 2018-11-01
api: delete workflow

dinosk:
* ADDS REST endpoint for workflow deletion.
  Propagates call further to reana-worfklow-controller.

Connects https://github.com/reanahub/reana-client/issues/192


dinosk:
This is doing the same propagation to RWC, if it is the time to move functionality here, it can be refactored.

dinosk:
Closing as a new status `deleted` has been added

PR #106, opened 2018-11-12
rest: workflow diff

dinosk:
* Adds REST API endpoint which calls
  reana-workflow-controller to retrieve
  a list of differences between two workflows.

Connects https://github.com/reanahub/reana-client/issues/193

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -1119,3 +1119,125 @@ def get_workflow_parameters(workflow_id_or_name):  # noqa
     except Exception as e:
         logging.error(traceback.format_exc())
         return jsonify({"message": str(e)}), 500
+
+
+@blueprint.route('/workflows/<workflow_id_or_name_a>/diff/'
+                 '<workflow_id_or_name_b>', methods=['GET'])
+def get_workflow_diff(workflow_id_or_name_a, workflow_id_or_name_b):  # noqa
+    r"""Get differences between two workflows.
+
+    ---
+    get:
+      summary: Get diff between two workflows.
+      description: >-
+        This resource shows the differences between
+        the assets of two workflows.
+        Resource is expecting two workflow UUIDs or names.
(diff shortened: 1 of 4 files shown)


reana_server/rest/workflows.py
```diff
@@ -1119,3 +1119,123 @@ def get_workflow_parameters(workflow_id_or_name):  # noqa
...
+        - name: brief
+          in: query
+          description: Optional flag. If set, file contents are examined.
+          required: false
+          type: boolean
+        - name: context_lines
+          in: query
+          description: Optional parameter. Sets number of context lines
+                       for workspace diff output.
+          required: false
```
diegodelemos: Also default value here.

reana_server/rest/workflows.py
```diff
@@ -1119,3 +1119,123 @@ def get_workflow_parameters(workflow_id_or_name):  # noqa
...
+          type: string
```
diegodelemos: Type is integer I guess?

reana_server/rest/workflows.py
```diff
@@ -1119,3 +1119,123 @@ def get_workflow_parameters(workflow_id_or_name):  # noqa
...
+        - name: workflow_id_or_name_b
+          in: path
+          description: Required. Analysis UUID or name of the second workflow.
+          required: true
+          type: string
+        - name: brief
+          in: query
+          description: Optional flag. If set, file contents are examined.
+          required: false
+          type: boolean
```
diegodelemos: We should document here the default value.

reana_server/rest/workflows.py
dinosk: Indeed this is what I tried first, and then after the RC->RS->RWC the type had changed to `string` so I changed it to string. I am not 100% sure, but I believe this happened with the type set to `integer` in both specs.

diegodelemos:
Needs rebase.

PR #111, opened 2018-11-26
api: add remove files from workspace endpoint

diegodelemos:
* Closes reanahub/reana-client#210.


scripts/generate_openapi_spec.py
```diff
@@ -43,7 +42,7 @@ def build_openapi_spec():
 
     package = __title__
     desc = __api_description__
-    ver = __api_version__
```
tiborsimko: :+1:

PR #116, opened 2018-12-17
global: workflow execution scheduler

dinosk:
* Adds scheduler.py with WorkflowExecutionScheduler class,
  responsible for consuming the workflow-submission queue and
  schedule the execution of workflows based on priority and
  system availability.

* Adds WorkflowSubmissionPublisher to do the publishing of the
  incoming workflow submissions to the corresponding message queue.

* Separates start_workflow from set_workflow_status.

Connects https://github.com/reanahub/reana/issues/119

### DIFF
--- a/reana_server/cli.py
+++ b/reana_server/cli.py
@@ -66,6 +66,7 @@ from reana_db.database import Session, init_db
 from reana_db.models import User
 
 from reana_server import config
+from reana_server.scheduler import WorkflowExecutionScheduler
 from reana_server.utils import _create_user, _get_users, create_user_workspace
 
 
@@ -207,3 +208,14 @@ def create_user(ctx, email, user_access_token, admin_access_token):
             click.style('User could not be created: \n{}'
                         .format(str(e)), fg='red'),
             err=True)
+
+
+@click.command('start-scheduler')
+def start_scheduler():
+    """Start a workflow execution scheduler process."""
+    logging.basicConfig(
+        level=logging.DEBUG,
+        format='%(asctime)s - %(threadName)s - %(levelname)s: %(message)s'
+    )
+    scheduler = WorkflowExecutionScheduler()
+    scheduler.run()
--- a/reana_server/scheduler.py
+++ b/reana_server/scheduler.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2018 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA Server Workflow Execution Scheduler."""
+
+import json
+import logging
+
+from reana_commons.config import MQ_DEFAULT_QUEUES
+from reana_commons.consumer import BaseConsumer
+from reana_commons.tasks import reana_ready
+from reana_db.database import Session
+from reana_db.models import Workflow, WorkflowStatus
+
+from reana_server.api_client import current_rwc_api_client
+
+
+class WorkflowExecutionScheduler(BaseConsumer):
+    """Scheduler of workflow execution.
+
+    Class responsible for consuming from the workflow-submission queue
+    and scheduling workflows for execution based on policies and system
+    availability.
+    """
+
+    def __init__(self, **kwargs):
+        """Constructor."""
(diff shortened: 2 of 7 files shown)


reana_server/cli.py
```diff
@@ -207,3 +208,14 @@ def create_user(ctx, email, user_access_token, admin_access_token):
             click.style('User could not be created: \n{}'
                         .format(str(e)), fg='red'),
             err=True)
+
+
+@click.command('start_scheduler')
```
diegodelemos: Just a cosmetic, should we use `-` to separate words in the commands as we do in [`consume-job-queue`](https://github.com/reanahub/reana-cluster/blob/master/reana_cluster/backends/kubernetes/templates/deployments/workflow-controller-template.yaml#L48)? It would affect [r-cluster PR](https://github.com/reanahub/reana-cluster/pull/138/files#diff-553f301b92242889471252a2c1462906R46).

reana_server/scheduler.py
```diff
@@ -0,0 +1,56 @@
...
+        """Implement providing kombu.Consumers with queues/callbacks."""
+        return [Consumer(queues=self.queue, callbacks=[self.on_message],
+                         accept=[self.message_default_format])]
+
+    def on_message(self, workflow_submission, message):
+        """On new workflow_submission event handler."""
+        if reana_ready():
+            message.ack()
+            workflow_submission = json.loads(workflow_submission)
+            print(workflow_submission)
```
diegodelemos: Shall we log here or it is just a forgotten print?

tests/test_scheduler.py
```diff
@@ -0,0 +1,75 @@
...
+# REANA is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with REANA; if not, see <http://www.gnu.org/licenses>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization or
+# submit itself to any jurisdiction.
```
diegodelemos: MIT license.

reana_server/scheduler.py
```diff
@@ -0,0 +1,56 @@
...
+            workflow_submission['status'] = 'start'
```
diegodelemos: I am most probably missing something here, how do we get a `status`? Seeing the call to `publish_workflow_submission` I see only `user_id`, `workflow_id_or_name`, `parameters` being passed.

Dockerfile
```diff
@@ -11,6 +11,8 @@ RUN apt-get update && \
 
 COPY CHANGES.rst README.rst setup.py /code/
 COPY reana_server/version.py /code/reana_server/
+RUN pip install -e git://github.com/dinosk/reana-db.git@119-workflow-submission-queue#egg=reana-db
```
diegodelemos: We can remove this installation as in https://github.com/reanahub/reana-workflow-controller/pull/162/files#r248676641.

reana_server/cli.py
dinosk: Thanks for the link

reana_server/scheduler.py
dinosk: Both!

reana_server/scheduler.py
dinosk: The scheduler only starts workflows, so status is always set to `start`. In brief the flow is: on_message is called when there's a new submission, if the conditions defined in the config are met then set status to `start` and call RWC, otw. requeue the message. Last part may need a bit of refactoring in another iteration because as it is mentioned in the kombu docs: http://docs.celeryproject.org/projects/kombu/en/latest/reference/kombu.message.html#kombu.message.Message.requeue it should not be used as a way to select which messages to process. Welcoming suggestions on different approaches for this task

setup.py
```diff
@@ -97,6 +97,7 @@
         'flask.commands': [
             'db = reana_server.cli:db',
             'users = reana_server.cli:users',
+            'start_scheduler = reana_server.cli:start_scheduler',
```
diegodelemos: `start-scheduler` :)

PR #117, opened 2019-01-17
config: set SHARED_VOLUME_PATH to /var/reana

dinosk:


reana_server/config.py
```diff
@@ -19,4 +19,4 @@
 
 ADMIN_USER_ID = "00000000-0000-0000-0000-000000000000"
 
-SHARED_VOLUME_PATH = os.getenv('SHARED_VOLUME_PATH', '/reana')
+SHARED_VOLUME_PATH = os.getenv('SHARED_VOLUME_PATH', '/var/reana')
```
tiborsimko: FWIW for Minikube there are some nicely pre-defined PV locations such as `/data`
https://github.com/kubernetes/minikube/blob/master/docs/persistent_volumes.md

dinosk: Interesting, `/data` could be a better location, we could do it when also centralizing the root path configuration.

PR #124, opened 2019-02-27
 interactive-execution: predef interactive sessions

diegodelemos:
Closes https://github.com/reanahub/reana-client/issues/258.

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -17,6 +17,7 @@ from bravado.exception import HTTPError
 from flask import Blueprint
 from flask import current_app as app
 from flask import jsonify, request, send_file
+from reana_commons.config import INTERACTIVE_SESSION_TYPES
 from reana_commons.utils import get_workspace_disk_usage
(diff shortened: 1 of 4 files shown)


reana_server/rest/workflows.py
```diff
@@ -1488,28 +1495,30 @@ def open_interactive_session(workflow_id_or_name):  # noqa
...
+                            with one of: [jupyter]."
               }
         500:
           description: >-
             Request failed. Internal controller error.
     """
     try:
         user_id = get_user_from_token(request.args.get('access_token'))
-
+        if interactive_session_type not in INTERACTIVE_SESSION_TYPES:
```
roksys: This check could be also done on the client side, to avoid unnecessary calls.

diegodelemos: It is already done [here](https://github.com/reanahub/reana-client/pull/257/files#diff-19bed1e7334d00a24f059e517c55f01dR887) with Click choices :)

PR #132, opened 2019-04-15
docs: add import & export users cli docs

### LINKED ISSUE #131: docs: document import/export users functionality
Import and export are missing from the CLI docs.

https://reana-server.readthedocs.io/en/latest/cli.html

roksys:
* Closes #131

Signed-off-by: Rokas Maciulaitis <rokas.maciulaitis@cern.ch>

### DIFF
--- a/reana_server/cli.py
+++ b/reana_server/cli.py
@@ -51,6 +51,19 @@ Similarly, to retrieve information for all users:
 .. code-block :: bash
 
     $ flask users get --admin-access-token=<token>
+
+To export, all users in current REANA cluster:
+
+.. code-block :: bash
+
+    # users will be exported in CSV format
+    $ flask export --admin-access-token=<token> > users.csv
+
+To import users to current REANA cluster:
+
+.. code-block :: bash
+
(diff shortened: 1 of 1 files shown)


reana_server/cli.py
```diff
@@ -51,6 +51,19 @@
 .. code-block :: bash
 
     $ flask users get --admin-access-token=<token>
+
+To export, all users in current REANA cluster:
+
+.. code-block :: bash
+
+    # users will be exported in CSV format
+    $ flask export --admin-access-token=<token>
```
diegodelemos: What do you think about this?
```suggestion
    $ flask export --admin-access-token=<token> > users.csv
```
Then we can save having the comment just before.

reana_server/cli.py
```diff
@@ -51,6 +51,19 @@
...
+
+To import users to current REANA cluster:
+
+.. code-block :: bash
+
+    $ flask import --file <users.txt> --admin-access-token=<token>
```
diegodelemos: And then here we can use the same name:

```suggestion
    $ flask import --file users.csv --admin-access-token=<token>
```

Issue #138, opened 2019-05-08
api: add manage secrets endpoints


diegodelemos:
As a continuation of https://github.com/reanahub/reana-client/issues/283, we need to add new endpoints to add, list and delete secrets, which could look like:

```http
GET /api/secrets/
```
```http
POST /api/secrets/<key>
Content-Type: application/json; charset=utf-8
Content-Length: 25

{
  "<key>": "<value>"
}
```

```http
DELETE /api/secrets/<key>
```

The secrets will be created in Kubernetes, using the [Python client](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#create_namespaced_secrets).

PR #139, opened 2019-05-09
api: endpoints to create, list and delete secrets

diegodelemos:
* Adds endpoints to create, list and delete authenticated user
  secrets (addresses #138).

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -3,7 +3,7 @@
   "info": {
     "description": "Submit workflows to be run on REANA Cloud",
     "title": "REANA Server",
-    "version": "0.5.0"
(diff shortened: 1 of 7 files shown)


docs/openapi.json
```diff
@@ -39,6 +39,258 @@
...
+              "application/json": {
+                "message": "Token is not valid"
+              }
+            }
+          },
+          "404": {
+            "description": "Request failed. Secrets do not exist.",
+            "examples": {
+              "application/json": {
+                "message": "The submitted secrets api_key, password, username do not exist."
```
okraskaj: I think it should be a list, shouldn't it?

PR #142, opened 2019-06-20
dockerfile: use shell-oriented DEBUG flag

diegodelemos:


Dockerfile
```diff
@@ -22,13 +22,16 @@ RUN pip install --no-cache-dir requirements-builder && \
 COPY . /code
 
 # Debug off by default
-ARG DEBUG=false
-RUN if [ "${DEBUG}" = "true" ]; then pip install -r requirements-dev.txt; pip install -e .; else pip install .; fi;
+ARG DEBUG=0
+RUN if [ "${DEBUG}" -eq 1 ]; then pip install -r requirements-dev.txt; pip install -e .; else pip install .; fi;
```
tiborsimko: It would be more flexible to use:

```
if [ "${DEBUG}" -gt 0 ]
```

since this opens a theoretical possibility to have various debugging levels, e.g. DEBUG=1 for basic things only, DEBUG=6 for performance, and DEBUG=9 for everything in small detail.

Dockerfile
```diff
@@ -22,13 +22,16 @@ RUN pip install --no-cache-dir requirements-builder && \
...
 
 # Building with locally-checked-out shared modules?
 RUN if test -e modules/reana-commons; then pip install modules/reana-commons --upgrade; fi
 RUN if test -e modules/reana-db; then pip install modules/reana-db --upgrade; fi
 
+# Check if there are broken requirements
+RUN pip check
```
tiborsimko: :+1:

PR #160, opened 2019-07-25
auth: use invenio session cookie to retrieve user

leticiawanderley:


reana_server/rest/ping.py
```diff
@@ -10,7 +10,7 @@
 
 from flask import Blueprint, jsonify
 
-blueprint = Blueprint('ping', __name__)
+blueprint = Blueprint('ping', __name__, url_prefix='/reana-api')
```
diegodelemos: We can't use `/api` because of Invenio's `/api`? If that's the case, we will have to fix the auth on the API application since this is going to break `reana-client`'s calls.

reana_server/rest/secrets.py
```diff
@@ -112,7 +112,10 @@ def add_secrets():  # noqa
               }
     """
     try:
-        user = get_user_from_token(request.args.get("access_token"))
+        if current_user.is_authenticated:
+            user = _get_user_from_invenio_user(current_user.email)
+        else:
+            user = get_user_from_token(request.args.get('access_token'))
```
diegodelemos: _Issue to be created once merged_.

Investigate usage of OAuth2Server to generate tokens so we do not need to generate and maintain our own, see [OAuth2Server CLI `tokens create`](https://github.com/inveniosoftware/invenio-oauth2server/blob/e3515a9d35f01ddae8fa27d2851ef18c80d96bdc/invenio_oauth2server/cli.py#L63)

reana_server/rest/secrets.py
```diff
@@ -112,7 +112,10 @@ def add_secrets():  # noqa
               }
     """
     try:
-        user = get_user_from_token(request.args.get("access_token"))
+        if current_user.is_authenticated:
```
diegodelemos: Can we use the `login_required` decorator? Otherwise, we need to return a 403 if the user is not authenticated nor with the session nor with the token (before we were throwing a [`ValueError` on `get_user_from_token`](https://github.com/reanahub/reana-server/blob/d4cfc0a8fc3930fda3af04647c4bcee0b5a706ba/reana_server/utils.py#L47), catching it and [throwing the 403](https://github.com/reanahub/reana-server/blob/d4cfc0a8fc3930fda3af04647c4bcee0b5a706ba/reana_server/rest/workflows.py#L166)).

tests/test_views.py
```diff
@@ -28,406 +28,444 @@
...
-            assert res.status_code == 403
-
-            res = client.get(
-                url_for("workflows.get_workflows"),
-                query_string={"access_token": default_user.access_token,
-                              "type": "batch"},
-            )
-            assert res.status_code == 200
+        mock_current_user = Mock(is_authenticated=False)
+        with patch('flask_login.utils._get_user',
```
diegodelemos: You can use the `mock.patch.multiple ` so it is a bit less indented :) see in [RWC](https://github.com/reanahub/reana-workflow-controller/blob/44b6d5eb9ed6e915f6e2294710eb6ef8c6fdd50a/tests/test_views.py#L1063-L1066).

reana_server/rest/secrets.py
```diff
@@ -112,7 +112,10 @@ def add_secrets():  # noqa
               }
```
diegodelemos: Originally, we have set the [`access_token` as a required parameter](https://reana-server.readthedocs.io/en/latest/_static/api.html#operation/get_workflows). This is no longer true, can you please update the docstring? This will be applicable for all endpoints which need authentication :).

tests/test_views.py
```diff
@@ -28,406 +28,444 @@
...
+                   Mock(return_value=mock_current_user)):
+            with patch(
+                "reana_server.rest.workflows.current_rwc_api_client",
+                make_mock_api_client("reana-workflow-controller")(),
+            ):
+                res = client.get(
+                    url_for("workflows.get_workflows"),
+                    query_string={"user_id": default_user.id_,
```
diegodelemos: This was not introduced by you, but since you are changing the code, could you remove the `user_id` parameter? [It is no longer in the API](https://reana-server.readthedocs.io/en/latest/_static/api.html#operation/get_workflows). Thanks!

tests/test_views.py
```diff
@@ -28,406 +28,444 @@
...
-            assert res.status_code == 400
+        mock_current_user = Mock(is_authenticated=False)
+        with patch('flask_login.utils._get_user',
+                   Mock(return_value=mock_current_user)):
+            with patch(
+                "reana_server.rest.workflows.current_rwc_api_client",
+                make_mock_api_client("reana-workflow-controller")(),
+            ):
+                # access token needs to be passed instead of user_id
+                res = client.post(
```
diegodelemos: We can remove this `access token needs to be passed instead of user_id ` part of the test.

tests/test_views.py
```diff
@@ -24,376 +24,325 @@
 
 from reana_server.utils import _create_and_associate_reana_user
 
+GET_USER_MOCK = Mock(return_value=Mock(is_authenticated=False))
```
diegodelemos: We could extract this and the mock:
```python
with patch("flask_login.utils._get_user", GET_USER_MOCK):
    ...
```
To a new fixture in `conftest` so the test runs inside this context, for example [here](https://github.com/reanahub/pytest-reana/blob/master/pytest_reana/fixtures.py#L319) we do it. I think it will simplify a bit more the changes.

Issue #155, opened 2019-07-19
gitlab: set webhooks to trigger REANA workflow runs


diegodelemos:
Once we are authorised by the user to modify their project, we should set up the corresponding webhooks to trigger workflow runs when there are new commits. For doing so we will have to add `git refs` to the workflow model https://github.com/reanahub/reana-db/blob/cdeac852419100cda649d914d5ca538ae93753a7/reana_db/models.py#L79.

We should be able to configure if we want to do workflow runs only on **MR commits**, **new commit to master**, or **every commit**.

PR #165, opened 2019-08-05
gitlab: add gitlab endpoints

leticiawanderley:
Addresses #155

Signed-off-by: Leticia Farias Wanderley <leticia.farias.wanderley@cern.ch>

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -14,7 +14,6 @@ from datetime import timedelta
 
 from invenio_app.config import APP_DEFAULT_SECURE_HEADERS
 from invenio_oauthclient.contrib import cern
-
 # Database
 # ========
 #: Database URI including user and password
@@ -30,6 +29,12 @@ AVAILABLE_WORKFLOW_ENGINES = [
 ADMIN_USER_ID = "00000000-0000-0000-0000-000000000000"
 
 SHARED_VOLUME_PATH = os.getenv('SHARED_VOLUME_PATH', '/var/reana')
+
+REANA_URL = os.getenv('REANA_URL', 'reana.io')
+
+REANA_SSO_CERN_CONSUMER_KEY = os.getenv('CERN_CONSUMER_KEY', 'CHANGE_ME')
+
+REANA_SSO_CERN_CONSUMER_SECRET = os.getenv('CERN_CONSUMER_SECRET', 'CHANGE_ME')
 
 
 # Invenio configuration
@@ -86,7 +91,7 @@ SESSION_COOKIE_SECURE = True
 #: provided, the allowed hosts variable is set to localhost. In production it
 #: should be set to the correct host and it is strongly recommended to only
 #: route correct hosts to the application.
-APP_ALLOWED_HOSTS = ['localhost', '127.0.0.1']
+APP_ALLOWED_HOSTS = ['localhost', '127.0.0.1', REANA_URL]
 
 # Security configuration
 # ======================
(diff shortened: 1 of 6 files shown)


reana_server/config.py
```diff
@@ -30,6 +30,12 @@
 
 SHARED_VOLUME_PATH = os.getenv('SHARED_VOLUME_PATH', '/var/reana')
 
+REANA_URL = os.getenv('REANA_URL', 'reana.io')
+
+CERN_CONSUMER_KEY = os.getenv('CERN_CONSUMER_KEY', 'CHANGE_ME')
+
+CERN_CONSUMER_SECRET = os.getenv('CERN_CONSUMER_SECRET', 'CHANGE_ME')
```
diegodelemos: What do you think of prefixing this variable and the previous with `REANA_SSO`? For example, `REANA_SSO_CERN_CONSUMER_KEY`.

reana_server/config.py
```diff
@@ -106,12 +112,19 @@ def _(x):
...
 
 DEBUG = True
 
 SECURITY_PASSWORD_SALT = 'security-password-salt'
 
 SECURITY_SEND_REGISTER_EMAIL = False
+
+# Gitlab Application configuration
+# ================================
+GITLAB_APP_ID = os.getenv('GITLAB_APP_ID', 'CHANGE_ME')
```
diegodelemos: For all GitLab ones, `REANA_GITLAB_OAUTH` prefix. For example, `REANA_GITLAB_OAUTH_APP_ID`. And you can move them to the upper section since this has nothing to do with Invenio.

reana_server/config.py
```diff
@@ -106,12 +112,19 @@ def _(x):
...
+GITLAB_APP_SECRET = os.getenv('GITLAB_APP_SECRET', 'CHANGE_ME')
+GITLAB_URL = os.getenv('GITLAB_URL', 'CHANGE_ME')
```
diegodelemos: The `REANA_GITLAB_OAUTH` prefix works for all GitLab variables but this one, this could simply be `REANA_GITLAB_URL`.

diegodelemos: Since we are using this URL to do a [direct request to GitLab including the application secret](https://github.com/reanahub/reana-server/pull/165/files#diff-eb771db72945ea9e2b49545b0938ce74R93), we should somehow force the HTTPS with something like:
```python
GITLAB_URL = 'https://{}'.format(os.getenv('REANA_GITLAB_HOST', 'CHANGE_ME'))
```

reana_server/utils.py
```diff
@@ -149,3 +154,42 @@ def _get_user_from_invenio_user(id):
     if not user:
         raise ValueError('No users registered with this id')
     return user
+
+
+def _get_yaml_file_from_gitlab(webhook_data, user_id):
```
diegodelemos: What do you think about `_get_reana_yaml_from_gitlab`?

reana_server/utils.py
```diff
@@ -149,3 +154,42 @@ def _get_user_from_invenio_user(id):
...
+    gitlab_token = secrets_store.get_secret_value('gitlab_access_token')
+    project_id = webhook_data['project']['id']
+    yaml_file = requests.get(gitlab_api.format(project_id, reana_yaml,
+                                               branch, gitlab_token))
+    return yaml.load(yaml_file.content), \
+        webhook_data['project']['path_with_namespace'], branch, \
+        commit_sha
+
+
+def _parse_gitlab_secret(gitlab_response):
```
diegodelemos: This is more formatting than parsing right? I would rename to `_format_gitlab_secrets` or similar.

PR #172, opened 2019-08-21
api: fix GitLab webhook redirect link

diegodelemos:


reana_server/rest/gitlab.py
```diff
@@ -194,7 +198,7 @@ def gitlab_webhook():  # noqa
         gitlab_url = REANA_GITLAB_URL + "/api/v4/projects/" + \
             "{0}/hooks?access_token={1}"
         webhook_payload = {
-            "url": REANA_URL + "/api/workflows",
+            "url": "https://{}/api/workflows".format(REANA_URL),
```
roksys: Will it work with - https://github.com/reanahub/reana-cluster/blob/d61451cc898ff9a0a233c9490e498e6e83831317/reana_cluster/cli/cluster.py#L73 ?

PR #173, opened 2019-08-23
setup: reana-commons upgrade

roksys:


diegodelemos:
👍Can you bump REANA-DB as well (https://github.com/reanahub/reana-db/pull/50)

diegodelemos:
Travis is failing, looks like we have to pin marshmallow, currently [it is openended dependency](https://github.com/diegodelemos/reana-server/blob/1438bf995fa70ff7b6e7fbf9a661bb5d2d95ac64/setup.py#L48) (also in https://github.com/reanahub/reana-workflow-controller/pull/245)

PR #181, opened 2019-10-11
api: stream file uploads using request.stream

diegodelemos:
* After the findings in
  https://github.com/diegodelemos/client-server-flask-big-file-uploads
  it is clear we have to move from multipart uploads, this commit changes
  the API description.

* Describing file content in request body, more here
  https://swagger.io/docs/specification/describing-request-body/file-upload/.

Addresses reanahub/reana-client#302.

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -12,23 +12,31 @@ import logging
 import subprocess
 import traceback
 
-import fs
+import requests
 from bravado.exception import HTTPError
+from reana_commons.config import INTERACTIVE_SESSION_TYPES
+from reana_commons.utils import get_workspace_disk_usage
+
+import fs
 from flask import Blueprint
 from flask import current_app as app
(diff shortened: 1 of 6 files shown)


reana_server/rest/workflows.py
```diff
@@ -869,12 +877,12 @@ def upload_file(workflow_id_or_name):  # noqa
           description: Required. Analysis UUID or name.
           required: true
           type: string
-        - name: file_content
-          in: formData
-          description: >-
-            Required. File to be transferred to the workflow workspace.
+        - name: file
+          in: body
+          description: Required. File to add to the workspace.
```
tiborsimko: Shall we update `reana-commons` with the new r-server and r-w-controller OpenAPI specs as well?

diegodelemos: Created https://github.com/reanahub/reana-commons/pull/141

Issue #176, opened 2019-08-30
config: if APP_ALLOWED_HOSTS is set REANA getting started fails


diegodelemos:
For the GitLab integration, we need the [`APP_ALLOWED_HOSTS`](https://github.com/reanahub/reana-server/blob/fb1dc2a74276999dd44359dc4f9f7b2d73a19f94/reana_server/config.py#L99) (currently commented out) set so the Flask application can generate correct URLs when doing `url_for`.

The problem with this is that we break the behaviour described in [our getting started](http://reana.io/#getstarted), where the person who deploys REANA locally would run `reana-cluster env --include-admin-token` and that would [extract **the IP of the VM** in which REANA runs](https://github.com/diegodelemos/reana-cluster/blob/6ccf0eef2dcba2f8a89b46817c698a00a8170750/reana_cluster/cli/cluster.py#L96). Because the IP is not a trusted host all `reana-client` calls would fail.

I see three solutions now:
- Only if `REANA_URL` is set we set the `APP_ALLOWED_HOSTS`
- We enable `APP_ALLOWED_HOSTS` always and then in REANA-Cluster we retrieve the minikube IP and pass it as REANA_URL.

diegodelemos:
You can rebase this on top of https://github.com/reanahub/reana-cluster/pull/240 which will be merged first as it is a small change.

PR #182, opened 2019-10-30
config: restrict allowed hosts to REANA_URL

mvidalgarcia:
Addresses #176

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -30,7 +30,7 @@ ADMIN_USER_ID = "00000000-0000-0000-0000-000000000000"
 
 SHARED_VOLUME_PATH = os.getenv('SHARED_VOLUME_PATH', '/var/reana')
 
-REANA_URL = os.getenv('REANA_URL', 'reana.io')
+REANA_URL = os.getenv('REANA_URL')
 
 REANA_SSO_CERN_CONSUMER_KEY = os.getenv('CERN_CONSUMER_KEY', 'CHANGE_ME')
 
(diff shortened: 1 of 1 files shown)


reana_server/config.py
```diff
@@ -94,7 +94,7 @@ def _(x):
 
 #: In production use the following configuration plus adding  the hostname/ip
 #: of the reverse proxy in front of REANA-Server.
-# APP_ALLOWED_HOSTS = [REANA_URL]
+APP_ALLOWED_HOSTS = [REANA_URL]
```
diegodelemos: If we proceed this way, and the `reana-cluster.yaml` doesn't have a `reana_url` the configuration variable (always the case when [running dev mode](https://github.com/reanahub/reana-cluster/blob/master/reana_cluster/configurations/reana-cluster-minikube-dev.yaml)), `REANA_URL` will fall back to `reana.io` (see [here](https://github.com/reanahub/reana-server/blob/84d06c61c17cbb1a402a09a644220aa3ad88a2b9/reana_server/config.py#L33)) which won't allow requests of any kind to come to the local REANA cluster unless they have the `reana.io` in the host header.

What do you think of doing: (1) do not fall back to `reana.io` but to `None` (2) set `APP_ALLOWED_HOSTS` only if `REANA_URL` is set (for example, we set it when we deploy in production or when we want to [set the GitLab integration](https://github.com/reanahub/reana-ui/pull/40/files#diff-2b2f3fdab1b53ff84ce39f48ddf75281R126)).

mvidalgarcia: Oh indeed, I didn't notice that. Changed it, is this what you mean?

PR #183, opened 2019-11-07
api: add /me blueprint route

mvidalgarcia:
Addresses reanahub/reana-ui#36

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -156,6 +156,7 @@
         "summary": "TEMPORARY displays a message about how to access a resource."
       }
     },
+    "/api/logout": {},
     "/api/me": {
       "get": {
         "description": "This resource provides basic information about an authenticated user based on the session cookie presence.",
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -156,6 +156,54 @@
         "summary": "TEMPORARY displays a message about how to access a resource."
       }
     },
+    "/api/me": {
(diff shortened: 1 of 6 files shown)


docs/openapi.json
```diff
@@ -156,6 +156,55 @@
         "summary": "TEMPORARY displays a message about how to access a resource."
       }
     },
+    "/api/logout": {},
+    "/api/me": {
+      "get": {
+        "description": "This resource provides basic information from an authenticated user based on the session cookie presence.",
```
roksys:
```suggestion
        "description": "This resource provides basic information about an authenticated user based on the session cookie presence.",
```

docs/openapi.json
```diff
@@ -156,6 +156,55 @@
...
+                "email": "user@reana.info",
+                "reana_token": "Drmhze6EPcv0fN_81Bj-nA"
+              }
+            },
+            "schema": {
+              "properties": {
+                "email": {
+                  "type": "string"
+                },
+                "reana_token": {
```
diegodelemos: What do you think about only serving the email? In the UI, the SPA will have a cookie already (if logged in), and if someone uses the REST API, they should be using the access token already so no need to return it.

mvidalgarcia: Discussed offline:
The REANA token can be useful there for users that log into the UI and grab the token in order to use `reana-client` for instance.

PR #190, opened 2019-12-09
users: add full_name and username to /me endpoint

mvidalgarcia:
addresses reanahub/reana-ui#50

### DIFF
--- a/reana_server/rest/users.py
+++ b/reana_server/rest/users.py
@@ -287,7 +287,8 @@ def get_me():
     if current_user.is_authenticated:
         me = _get_user_from_invenio_user(current_user.email)
         return (
-          jsonify({'email': me.email, 'reana_token': me.access_token}), 200
+          jsonify({'email': me.email, 'reana_token': me.access_token,
+                   'full_name': me.full_name, 'username': me.username}), 200
         )
     else:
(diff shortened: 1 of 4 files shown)


reana_server/rest/users.py
```diff
@@ -287,7 +287,8 @@ def get_me():
     if current_user.is_authenticated:
         me = _get_user_from_invenio_user(current_user.email)
         return (
-          jsonify({'email': me.email, 'reana_token': me.access_token}), 200
+          jsonify({'email': me.email, 'reana_token': me.access_token,
```
mvidalgarcia: Maybe we can use marshmallow for this? Just curious.

diegodelemos: There is an [old RFC](https://github.com/reanahub/reana-job-controller/issues/106), still open, concerning this discussion.

At that moment we chose to not propagate it's usage since we were only using it for internal APIs (REANA-Job-Controller one). But we also agreed that it would be nice to have for external the API (aka REANA-Server), this issue is currently in Cluster future column in [Triage](https://github.com/orgs/reanahub/projects/7).

If we want to move forward with it, porting RS to use it or even all APIs (after February workshop), we could move it to Cluster next and we would take it for next sprint. In any case, this is not lost, it is well noted down :)

PR #201, opened 2020-01-10
docs: remove reana-cluster references

diegodelemos:
* REANA-Cluster is being deprecated so we are removing all references
  to it (closes reanahub/reana#238).

### DIFF
--- a/reana_server/cli.py
+++ b/reana_server/cli.py
@@ -19,12 +19,11 @@ By default the server docker image is set to create the administrator account
 on startup, which has the email ``info@reana.io`` and a token which is
 generated using the ``secrets`` python library.
 
-To retrieve the administrator token you can use the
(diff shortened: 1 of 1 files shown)


reana_server/cli.py
```diff
@@ -20,11 +20,11 @@
 generated using the ``secrets`` python library.
 
 To retrieve the administrator token you can use the
-`reana-cluster <https://github.com/reanahub/reana-cluster>`_ CLI:
+`reana-dev <https://github.com/reanahub/reana>`_ CLI:
```
tiborsimko: Since `reana-dev` is supposed to be known to all developers, and since the target link may seem cryptic as to where to find it, perhaps simply remove the link... as people studying `reana-server` should have come across `reana-dev` already.

Issue #204, opened 2020-01-17
scheduler: container gets restarted when failure


diegodelemos:
During the investigation of https://github.com/reanahub/reana/issues/253 I've tried to make the scheduler fail to see if there was a problem with the logs (which there was https://github.com/reanahub/reana-server/issues/203). However, when I made it fail like this:
```console
$ kubectl exec -ti server-xxxxxxxxxx-xxxxx bash
> python
from reana_commons.publisher import WorkflowSubmissionPublisher
wsp._publish({'test':'msg'})
```
There were logs shown but the container got restarted so logs were gone:
```console
$ kubectl  logs server-xxxxxxxxxx-xxxxx scheduler  -f
Traceback (most recent call last):
  File "/usr/local/bin/flask", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/site-packages/flask/cli.py", line 966, in main
    cli.main(prog_name="python -m flask" if as_module else None)
  File "/usr/local/lib/python3.6/site-packages/flask/cli.py", line 586, in main
....
  File "/usr/local/lib/python3.6/site-packages/bravado/client.py", line 247, in __call__
    self.operation, request_options, **op_kwargs)
  File "/usr/local/lib/python3.6/site-packages/bravado/client.py", line 285, in construct_request
    construct_params(operation, request, op_kwargs)
  File "/usr/local/lib/python3.6/site-packages/bravado/client.py", line 307, in construct_params
    .format(operation.operation_id, param_name))
bravado_core.exception.SwaggerMappingError: set_workflow_status does not have parameter test
$ # I got kicked from following the logs (-f flag)
$ kubectl  logs server-76d6b4d68c-s7h7c scheduler
$ # and when I tried to retrieve the logs again, they were empty
```

We shouldn't exit the execution but just log the issue and continue listening.

PR #205, opened 2020-01-21
scheduler: requeue workflows when starting failure

diegodelemos:
* In the event of workflow start up errors (due to DB connections,
  failure connecting to Kubernetes etc...) REANA-Server's scheduler
  should requeue the workflow, otherwise it will be lost
  (closes reanahub/reana#253 #204).

### DIFF
--- a/reana_server/scheduler.py
+++ b/reana_server/scheduler.py
@@ -11,13 +11,14 @@
 import json
 import logging
 
-from reana_commons.config import MQ_DEFAULT_QUEUES
+from bravado.exception import HTTPBadGateway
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN apt-get update && \
       gcc \
       vim-tiny \
       libffi-dev && \
-    pip install --upgrade pip
+    pip install --upgrade "pip<20.0"
```
roksys: I think this is not needed anymore because 20.0.1 was released with a fix, right?

PR #208, opened 2020-01-27
rest: changes get_workspace method with workspace_path field

roksys:
* Addresses reanahub/reana-workflow-controller/issues/289

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -28,7 +28,8 @@ from reana_db.utils import _get_workflow_with_uuid_or_name
 from reana_server.api_client import (current_rwc_api_client,
                                      current_workflow_submission_publisher)
 from reana_server.config import SHARED_VOLUME_PATH
-from reana_server.utils import (RequestStreamWithLen,
+from reana_server.utils import (clone_workflow,
+                                RequestStreamWithLen,
                                 _get_reana_yaml_from_gitlab,
                                 _get_user_from_invenio_user,
                                 get_user_from_token, is_uuid_v4)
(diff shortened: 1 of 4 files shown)


reana_server/rest/workflows.py
```diff
@@ -698,19 +699,21 @@ def start_workflow(workflow_id_or_name):  # noqa
...
         current_workflow_submission_publisher.publish_workflow_submission(
             user_id=str(user.id_),
-            workflow_id_or_name=workflow_id_or_name,
+            workflow_id_or_name=workflow.get_full_workflow_name(),
             parameters=parameters
         )
         response = {'message': 'Workflow submitted.',
-                    'workflow_id': workflow_id_or_name,
-                    'workflow_name': workflow_id_or_name,
+                    'workflow_id': workflow.id_,
```
diegodelemos: Good catch 👍

reana_server/rest/workflows.py
```diff
@@ -698,19 +699,21 @@ def start_workflow(workflow_id_or_name):  # noqa
                     [WorkflowStatus.finished, WorkflowStatus.failed]:
                 raise ValueError(
                     "Only finished or failed workflows can be restarted.")
+            workflow = restart_workflow(workflow)
```
diegodelemos: As I read it, I already think that this will "restart" the workflow immediately. What do you think of changing the method name? currently, the docstring explains what it does, and the name explains why it does it (for restarting). I would use something more along the lines of (it is just a suggestion):

```console
def clone_workflow(workflow):
    """Create a copy of workflow in DB for the restarting process."""
...
```

reana_server/utils.py
```diff
@@ -260,3 +260,27 @@ def __len__(self):
         if not hasattr(self.limitedstream, 'limit'):
             return 0
         return self.limitedstream.limit
+
+
+def restart_workflow(workflow):
+    """Create a copy of workflow in DB for restarting."""
+    try:
+        restart_workflow = Workflow(
```
diegodelemos: WDYT of `cloned_workflow` or `copied_workflow` as var name?

reana_server/rest/workflows.py
```diff
@@ -1962,13 +1965,13 @@ def get_workflow_disk_usage(workflow_id_or_name):  # noqa
         summarize = bool(parameters.get('summarize', False))
         block_size = parameters.get('block_size')
         reana_fs = fs.open_fs(SHARED_VOLUME_PATH)
-        if reana_fs.exists(workflow.get_workspace()):
+        if reana_fs.exists(workflow.workspace_path):
             absolute_workspace_path = reana_fs.getospath(
-                workflow.get_workspace())
+                workflow.workspace_path)
```
tiborsimko: We can now upgrade the minimal version of `reana-db` to `0.7.0.dev20200205` in `setup.py`.

PR #210, opened 2020-02-03
installation: shared packages version bump

diegodelemos:
* Addresses reanahub/reana-workflow-controller/issues/289


diegodelemos:
I used the `reana-dev git-shared-modules-upgrade --commit-and-publish` with Rokas PRs checked out.. that's why the little mess... I am amending it now.

PR #213, opened 2020-02-06
rest: load json boolean

diegodelemos:
* `true` and `false` when passed to `bool()` are always true
  (closes reanahub/reana-client#367).

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -8,6 +8,7 @@
 
 """Reana-Server workflow-functionality Flask-Blueprint."""
 import io
+import json
 import logging
 import subprocess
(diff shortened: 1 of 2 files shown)


reana_server/rest/workflows.py
```diff
@@ -169,7 +170,7 @@ def get_workflows():  # noqa
         else:
             user = get_user_from_token(request.args.get('access_token'))
         type = request.args.get('type', 'batch')
-        verbose = request.args.get('verbose', False)
```
tiborsimko: Do we have the same issue with the following?

```
reana_server/rest/workflows.py:1492:        brief = request.args.get('brief', False)
```

PR #215, opened 2020-02-10
rest: return proper message if workflow can't be started

roksys:
* Closes reanahub/reana-client/issues/357


reana_server/rest/workflows.py
```diff
@@ -708,7 +708,12 @@ def start_workflow(workflow_id_or_name):  # noqa
                 workflow,
                 parameters.get('reana_specification', None))
         elif workflow.status != WorkflowStatus.created:
-            raise ValueError("Workflow cannot be started again.")
+            raise ValueError(
+                'Wokflow - {} is already {} and cannot be started '
```
tiborsimko: `Workflow {} is already`

PR #217, opened 2020-02-12
installation: pin Flask-Login

diegodelemos:
* New version of Flask-Login (0.5.0) exposes a bug in Flask-Security
  (https://github.com/mattupstate/flask-security/issues/681) which
  causes errors on endpoints which receive a JSON list in the
  body of the request.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ install_requires = [
     'webcolors==1.7',
     # Invenio dependencies
     'Flask>=1.0.2',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -59,6 +59,7 @@
     'webcolors==1.7',
     # Invenio dependencies
     'Flask>=1.0.2',
+    'Flask-Login<0.5.0',
```
tiborsimko: I'm just testing locally the same fix :wink: 

Please leave a ` # FIXME waiting for Flask-Security release` after the pin, so that we don't forget

PR #220, opened 2020-02-14
rest: add endpoint to return workflow spec

### LINKED ISSUE #219: rest: add endpoint to return the reana_specification used for an specific workflow run
In order to improve the feedback a user gets from a run, we shall create a new endpoint to return the used specification:
```console
/workflows/<workflow_id_or_name>/specification
```

diegodelemos:
* Returns a reconstruction of the `reana.yaml`, including only
  workflow specification, workflow type and input parameters
  (closes #219).

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -30,6 +30,7 @@ from reana_server.api_client import (current_rwc_api_client,
                                      current_workflow_submission_publisher)
 from reana_server.config import SHARED_VOLUME_PATH
 from reana_server.utils import (clone_workflow,
+                                reconstruct_reana_yaml,
                                 RequestStreamWithLen,
                                 _get_reana_yaml_from_gitlab,
                                 _get_user_from_invenio_user,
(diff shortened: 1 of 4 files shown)


reana_server/rest/workflows.py
```diff
@@ -355,6 +356,108 @@ def create_workflow():  # noqa
...
+        500:
+          description: >-
+            Request failed. Internal controller error.
+    """
+    try:
+        if current_user.is_authenticated:
+            user = _get_user_from_invenio_user(current_user.email)
+        else:
+            user = get_user_from_token(request.args.get('access_token'))
+        steps = request.json or None
```
mvidalgarcia: `steps` is not needed. Probably copy paste leftover.

reana_server/utils.py
```diff
@@ -286,3 +286,15 @@ def clone_workflow(workflow, reana_spec):
         logging.error(
             f'Error while creating {cloned_workflow.id_}: {message}\n{e}',
             exc_info=True)
+
+
+def reconstruct_reana_yaml(workflow):
+    """Reconstruct the REANA.yaml from a workflow object."""
+    reana_yaml_spec = {}
```
mvidalgarcia: This could be a oneliner

```python
reana_yaml_spec = {'workflow': {'specification': workflow.get_specification(), 'type': workflow.type_}, 'inputs': {'parameters': workflow.get_input_parameters()}}
```

PR #224, opened 2020-02-26
ext: add teardown REANA DB session in ext

### LINKED ISSUE #159: factory: integrate session teardown with Invenio application
According to [this commit](https://github.com/reanahub/reana-server/commit/25a0caaad108a2c77119d406c73f3abc3d7ab704), since we are using SQLAlchemy we need to look after the session lifecycle. As we begin wrapping the Invenio app around REANA, `factory` is no longer used so there should be another way to manage the session's lifecycle.

diegodelemos:
* Closes #159, closes reanahub/reana-client#358.

### DIFF
--- a/reana_server/ext.py
+++ b/reana_server/ext.py
@@ -27,6 +27,15 @@ class REANA(object):
         self.init_config(app)
         Menu(app=app)
 
+        @app.teardown_appcontext
+        def shutdown_reana_db_session(response_or_exc):
+            """Close session on app teardown."""
+            from reana_db.database import Session as reana_db_session
+            from invenio_db import db as invenio_db
+            reana_db_session.remove()
+            invenio_db.session.remove()
+            return response_or_exc
+
         @app.before_first_request
         def connect_signals():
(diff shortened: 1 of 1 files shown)


diegodelemos:
Closing REANA DB session (which wasn't being closed) is not making the memory consumption go down, after 10 minutes of doing `reana-client list`, there is a big difference in memory consumption between RS and RWC, when the number of operations is almost the same:
image
There is the default `shutdown_session` from SQLAlchemy which should be tearing down Invenio (SQLAlchemy) session plus the one for REANA-DB:
```console
>>> current_app.teardown_appcontext_funcs
[
  <function SQLAlchemy.init_app.<locals>.shutdown_session at 0x7f7f57d99c80>, 
  <bound method Alembic._clear_cache of <flask_alembic.extension.Alembic object at 0x7f7f559fb780>>, 
  <function REANA.init_app.<locals>.shutdown_reana_db_session at 0x7f7f559a7b70>
]
```
To be continued.

PR #229, opened 2020-03-19
rest: return input files and parameters in spec endpoint

### LINKED ISSUE #226: api: serve updated parameters in spec endpoint
The [spec](https://github.com/reanahub/reana-server/blob/master/reana_server/rest/workflows.py#L362) endpoint returns the original REANA specification but doesn't update its parameters when those are set via reana-client.

Steps to reproduce:

1. Run (reana-demo-root6-roofit)
```console
 reana-client run -w roofit -p events=1000
```

2. Go to UI, workflow details page, Specification tab and check `parameters`.

Can be addressed at the same time as https://github.com/reanahub/reana-server/issues/227 if possible.

### LINKED ISSUE #227: api: input files missing in spec endpoint
The [spec](https://github.com/reanahub/reana-server/blob/master/reana_server/rest/workflows.py#L362) endpoint doesn't include the input files so this information is missing in the UI.

Can be addressed at the same time as https://github.com/reanahub/reana-server/issues/226 if possible.

mvidalgarcia:
closes #226
closes #227

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ install_requires = [
     'fs>=2.0',
     'marshmallow>2.13.0,<=2.20.1',
     'pyOpenSSL==17.5.0',
-    'reana-commons[kubernetes]>=0.7.0.dev20200217,<0.8.0',
+    'reana-commons[kubernetes]>=0.7.0.dev20200319,<0.8.0',
     'reana-db>=0.7.0.dev20200205,<0.8.0',
     'requests==2.20.0',
     'rfc3987==1.3.7',
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ readme = open('README.rst').read()
 history = open('CHANGES.rst').read()
 
 tests_require = [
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -71,14 +70,14 @@
...
     # From auth bundle
-    'invenio-accounts==1.1.1',  # FIXME
     'invenio-oauth2server>=1.0.3,<1.1.0',
     'invenio-oauthclient>=1.1.2,<1.2.0',
     'invenio-userprofiles>=1.0.1,<1.1.0',
     # Not included in previous dependencies
     'invenio-db[postgresql]>=1.0.3,<1.1.0',
     # Invenio dependencies missing from Invenio packages setup.py
     'jsonpatch>=1.23,<2',
+    'six>=1.12.0',
```
diegodelemos: Why do we need this now?

mvidalgarcia: https://travis-ci.org/github/reanahub/reana-server/builds/664785079?utm_medium=notification&utm_source=github_status

PR #231, opened 2020-04-02
rest: validate operational options server-side

mvidalgarcia:
Validates operational options both on creation (reana.yaml options)
and on start (CLI passed options).

closes reanahub/reana-client#396


reana_server/rest/workflows.py
```diff
@@ -815,6 +817,8 @@ def start_workflow(workflow_id_or_name):  # noqa
         parameters = request.json
         workflow = _get_workflow_with_uuid_or_name(
             workflow_id_or_name, str(user.id_))
+        parameters['operational_options'] = validate_operational_options(
```
diegodelemos, line 820: We should add an `except` for `REANAValidationError` and return `400` otherwise, we will return `500` with a validation error, right?

PR #237, opened 2020-04-24
cli: allow admins to manage user tokens

### LINKED ISSUE #235: server: user token creation through request/approval process
- [x] Allow generating user tokens using [server Python CLI](https://github.com/reanahub/reana-server/blob/master/reana_server/cli.py#L113).
- [x] Send an automatic email informing the user that token has been created and it's accessible on https://reana.cern.ch/profile

Tightly related to https://github.com/reanahub/reana-ui/issues/93

diegodelemos:
* In order to open REANA during pilot face we need to control the
  number of users allowed in the system. For this, we create a new
  approval process which is managed with these new CLI commands
  only executable by admins (closes #235).


reana_server/cli.py
```diff
@@ -271,6 +272,50 @@ def import_users(ctx, admin_access_token, file_):
...
+    '--admin-access-token',
+    default=os.environ.get('REANA_ADMIN_ACCESS_TOKEN', None),
+    help='The access token of an administrator.')
+def token_grant(id, admin_access_token):
+    """."""
+    user_id = 'dc4aef21-c324-4fd5-8480-c724de1f0c06'
+    user_email = 'john.doe@google.com'
+    user_granted_token = 'c0fa47fa00ae4013a13fd7n'
+    # create a token in UserToken table and activate it
+    # send notification to user by email
```
mvidalgarcia: Just to remember: also create an audit log entry.

reana_server/cli.py
```diff
@@ -271,6 +272,50 @@ def import_users(ctx, admin_access_token, file_):
...
+
+@reana_users_token.command('grant')
+@click.option(
+    '--id',
+    help='The id of the user.')
+@click.option(
+    '--admin-access-token',
+    default=os.environ.get('REANA_ADMIN_ACCESS_TOKEN', None),
+    help='The access token of an administrator.')
+def token_grant(id, admin_access_token):
```
mvidalgarcia: What about allowing to grant by email address? The email is supposed to be unique in the database so it shouldn't be a problem and it would save us the step of searching the user id. If it's too much work it can be ticketized and done later IMO.

Same for revoking.

reana_server/cli.py
```diff
@@ -271,6 +272,50 @@ def import_users(ctx, admin_access_token, file_):
...
+
+
+@reana_users_token.command('revoke')
+@click.option(
+    '--id',
+    help='The id of the user.')
+@click.option(
+    '--admin-access-token',
+    default=os.environ.get('REANA_ADMIN_ACCESS_TOKEN', None),
+    help='The access token of an administrator.')
```
mvidalgarcia: Many commands have this option, what about creating a decorator?

PR #239, opened 2020-04-28
cli: add grant & revoke access token commands

### LINKED ISSUE #235: server: user token creation through request/approval process

mvidalgarcia:
closes #235

##### TODO:
- [x] Send emails to users
- [x] Add  unit tests

To test the email functionality you need to have set the following env vars:
- `REANA_EMAIL_SMTP_SERVER` (smtp.cern.ch)
- `REANA_EMAIL_SMTP_PORT` (587)
- `REANA_EMAIL_LOGIN` (youremaill@cern.ch)
- `REANA_EMAIL_SENDER` (reana-support@cern.ch)
- `REANA_EMAIL_PASSWORD` (yourpassword)

### DIFF
--- a/reana_server/reana_admin.py
+++ b/reana_server/reana_admin.py
@@ -20,12 +20,17 @@ import tablib
 from flask.cli import FlaskGroup, with_appcontext
 from reana_commons.utils import click_table_printer
 from reana_db.database import Session, init_db
-from reana_db.models import User
+from reana_db.models import AuditLogAction, User, UserTokenStatus
 
 from reana_server.config import ADMIN_USER_ID
 from reana_server.factory import create_app
+<<<<<<< HEAD
 from reana_server.status import STATUS_OBJECT_TYPES
 from reana_server.utils import (_create_user, _export_users, _get_users,
+=======
+from reana_server.utils import (_create_user, _export_users,
+                                _get_user_by_criteria, _get_users,
+>>>>>>> cli: add grant & revoke access token commands
                                 _import_users, create_user_workspace)
 
 
@@ -33,7 +38,8 @@ def admin_access_token_option(func):
     """Click option to load admin access token."""
     @click.option(
         '--admin-access-token',
-        default=os.environ.get('REANA_ADMIN_ACCESS_TOKEN', None),
+        required=True,
+        default=os.environ.get('REANA_ADMIN_ACCESS_TOKEN'),
         help='The access token of an administrator.')
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
@@ -151,6 +157,7 @@ def list_users(ctx, id, email, user_access_token, admin_access_token,
 @click.option(
     '-e',
     '--email',
+    required=True,
     help='The email of the user.')
 @click.option(
     '--user-access-token',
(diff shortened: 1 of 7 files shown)


reana_server/reana_admin.py
```diff
@@ -214,39 +218,102 @@ def import_users(ctx, admin_access_token, file_):
...
+    is_flag=True,
+    default=False,
+    help='Grant token even if user has a revoked access token.')
+def token_grant(admin_access_token, id_, email, force):
+    """Grant a token to the selected user."""
+    try:
+        admin = User.query.filter_by(id_=ADMIN_USER_ID).one_or_none()
+        if admin_access_token != admin.access_token:
+            raise ValueError('Admin access token invalid.')
+        user = _get_user_by_criteria(id_, email)
```
diegodelemos, line 239: What do you think of adding `_get_user_by_criteria` to be the callback that loads the user so `--user` could be `id` or `email`? Also, this callback could be giving an error if the user doesn't exist, so we don't have to duplicate this error handling.

diegodelemos: Seeing the `elif`'s complexity, previous comment isn't so important as more conditions are chained and cannot be reused from the two commands...

reana_server/reana_admin.py
```diff
@@ -214,39 +218,102 @@ def import_users(ctx, admin_access_token, file_):
...
+        elif (not force and user.access_token_status in
+              [UserTokenStatus.revoked.name, None]):
+            error_msg = (f'User {user.id_} ({user.email}) access token status'
+                         f' is {user.access_token_status}, if you want to'
+                         ' proceed in any case use --force option.')
+        if error_msg:
+            click.secho(f'ERROR: {error_msg}', fg='red')
+            sys.exit(1)
+
+        user_granted_token = secrets.token_urlsafe(16)
```
diegodelemos, line 255: We have the logic to create tokens a bit spread, it would be nice to centralise.
```console
$ rg "token_urlsafe\(16\)"
reana-server/tests/test_cli.py
44:    user_access_token = secrets.token_urlsafe(16)

reana-server/reana_server/reana_admin.py
255:        user_granted_token = secrets.token_urlsafe(16)

reana-server/reana_server/utils.py
85:            user_access_token = secrets.token_urlsafe(16)
```
(Not for now, but we can open an issue after merging this)

diegodelemos:
Still not working... but duplicated session seems to be fixed with the commit and https://github.com/reanahub/pytest-reana/pull/60.

Now the problems are:
- `from flask import current_app` is not the correct flask app inside the commands
- The session fixture is behaving weirdly as the should "pass" but there is this error:
  ```console
  E           sqlalchemy.orm.exc.DetachedInstanceError: Instance <User at 0x10f06e310> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/bhk3)
  ```

To be continued...

reana_server/reana_admin.py
mvidalgarcia: Definitely.

reana_server/reana_admin.py
```diff
@@ -214,39 +209,102 @@ def import_users(ctx, admin_access_token, file_):
...
+            error_msg = (f'User {user.id_} ({user.email}) does not have an'
+                         ' active access token.')
+        if error_msg:
+            click.secho(f'ERROR: {error_msg}', fg='red')
+            sys.exit(1)
+
+        user.active_token.status = UserTokenStatus.revoked
+        Session.commit()
+        log_msg = f'User\'s {user.id_} ({user.email}) token successfully' \
+            ' revoked.'
```
diegodelemos, line 298: Messages to be amended according to https://github.com/reanahub/docs.reana.io/pull/44#discussion_r417879081.

reana_server/reana_admin.py
```diff
@@ -214,39 +209,102 @@ def import_users(ctx, admin_access_token, file_):
...
+        if not user:
+            error_msg = f'User {id_ or email} does not exist.'
+        elif user.access_token:
+            error_msg = (f'User {user.id_} ({user.email}) has already an'
+                         ' active access token.')
+        elif (not force and user.access_token_status in
+              [UserTokenStatus.revoked.name, None]):
+            error_msg = (f'User {user.id_} ({user.email}) access token status'
+                         f' is {user.access_token_status}, if you want to'
+                         ' proceed in any case use --force option.')
```
diegodelemos, line 241: This means we will always have to use `--force` right? If that is the case, I think we should not use force but rather a confirmation such as https://click.palletsprojects.com/en/7.x/prompts/#confirmation-prompts, what do you think?

mvidalgarcia: Not always, only when the user has not requested the token or the token has been revoked, because they are not usual actions.
For the typical workflow, when the user has requested the token, you don't need to use `--force`.
Said that, the confirmation prompt seems like a good idea for the "force" cases, I didn't know this feature.

mvidalgarcia: Fixed, now it displays something like:
```console
$ flask reana-admin token-grant -e johndoe@cern.ch
User xxx-aaa-bbb (johndoe@cern.ch) access token status is None, do you want to proceed? [y/N]: N
Grant token aborted.
```

PR #245, opened 2020-05-18
installation: upgrade to Invenio 3.2

diegodelemos:


diegodelemos:
CI works correctly but I cannot get the auth workflow working locally (problems building REANA-UI related to craco), I would say we merge to unblock master, and if we find some trouble with auth when integrating the token request/approval process we fix it there.

PR #248, opened 2020-05-25
rest: group token data on a dict

mvidalgarcia:
closes reanahub/reana-ui#95

### DIFF
--- a/reana_server/rest/users.py
+++ b/reana_server/rest/users.py
@@ -272,15 +272,23 @@ def get_me():
               email:
                 type: string
               reana_token:
-                type: string
-              reana_token_status:
-                type: string
+                type: object
+                properties:
+                  value:
+                    type: string
+                  status:
+                    type: string
+                  requested_at:
+                    type: string
           examples:
             application/json:
               {
                 "email": "user@reana.info",
-                "reana_token": "Drmhze6EPcv0fN_81Bj-nA",
-                "reana_token_status": "active",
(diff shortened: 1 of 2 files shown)


reana_server/rest/users.py
```diff
@@ -321,11 +332,20 @@ def get_me():
...
-                             'reana_token': me.access_token,
-                             'reana_token_status': me.access_token_status,
-                             'full_name': me.full_name,
-                             'username': me.username}), 200)
+            return (jsonify({
+                'email': me.email,
+                'reana_token': {
+                    'value': me.access_token,
+                    'status': me.access_token_status,
+                    'created':
```
diegodelemos, line 340: I guess that from here comes my confusion in https://github.com/reanahub/reana-ui/pull/98/files#r430279346, what do you think of simplifying this to:
(1) Only one field `updated`: Then, because of the `status` field you know that the status is requested and last update was `yyyymmdd`, this is when the token was requested.
or
(2) We simplify even more and instead of `status` and `updated` we have `active`|`requested`|`revoked`: Each of these fields is optional and of type date.

mvidalgarcia: Indeed when the token is requested, `created` == `updated`. I think we can simply pass `updated` and renaming it to `request_time` here. 

```python
...
'request_time': me.latest_access_token.created if me.latest_access_token else None
...
```

Created time will always be request time as it doesn't change over time.

PR #253, opened 2020-06-01
Vc3v2p2

CodyKank:
### VC3 Integration
-----
Integrating changes for VC3, mainly consisting of changing file ownership to the VC3 user so we can manipulate the `user_workspace_path` while passing between some HPC and the vc3 headnode.


reana_server/utils.py
```diff
@@ -44,6 +45,10 @@ def create_user_workspace(user_workspace_path):
     reana_fs = fs.open_fs(app.config['SHARED_VOLUME_PATH'])
     if not reana_fs.exists(user_workspace_path):
         reana_fs.makedirs(user_workspace_path)
+    if os.environ.get("VC3USERID", None):
```
mvidalgarcia, line 48: You can use [`os.getenv()`](https://docs.python.org/3/library/os.html#os.getenv) directly and there's no need to specify `None` as it's the default :)

PR #254, opened 2020-06-03
rest: return announcement in /me endpoint

mvidalgarcia:
closes reanahub/reana-ui#83

### DIFF
--- a/reana_server/rest/users.py
+++ b/reana_server/rest/users.py
@@ -18,7 +18,7 @@ from invenio_oauthclient.utils import get_safe_redirect_target
 from reana_db.models import AuditLogAction
 from reana_commons.email import send_email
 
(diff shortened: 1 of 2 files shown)


reana_server/rest/users.py
```diff
@@ -339,7 +339,8 @@ def get_me():
                     if me.latest_access_token else None,
                 },
                 'full_name': me.full_name,
-                'username': me.username}), 200)
+                'username': me.username,
+                'announcement': REANA_UI_ANNOUNCEMENT}), 200)
```
mvidalgarcia, line 343: I'm not sure if this is the best place to put the announcement. Perhaps we should think of another way to pass configuration to the UI in the future.

diegodelemos: Maybe we can open an issue describing it and once we have another use case we make a new endpoint for config?

PR #258, opened 2020-06-12
accounts: support local users

### LINKED ISSUE #256: accounts: allow local users
Currently, REANA uses CERN SSO for user login.

It would be desirable to allow deployment scenarios with local user database. 

Basically AuthN should be pluggable, like r-j-controller, with offering two login techniques: local accounts, CERN SSO.

Then, any interested partners for deployment of REANA on premises would be able to follow the examples and introduce new login classes for University Foo login system.

For local accouns, we can use invenio-accounts package, and create simple forms for user login here and in r-ui.  (Note that it could be interesting to also have `reana-client login` facility.)

The local user feature could be used for integration testing, we could have several personas simulating users, group conveners, admins, etc.

mvidalgarcia:
closes #256 
closes https://github.com/reanahub/reana-ui/issues/107

ATM, signing in with a too-short password, displays error message "Password length should be between X and Y". This message will disappear once https://github.com/inveniosoftware/invenio-accounts/pull/336 is merged and we update to the latest version.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,9 @@ Version master (UNRELEASED)
 - Extends import/export commands with new user model fields.
 - Installs submodels in editable mode for live code updates.
 - Adds Black formatter support.
+- Bumps to Invenio-Accounts 1.3.0 to support REST API.
+- Adds support to local users.
+- Renames ``/api/me`` to ``/api/you`` due to conflict with Invenio-Accounts.
 
 Version 0.6.1 (2020-05-25)
 --------------------------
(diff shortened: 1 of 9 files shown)


mvidalgarcia:
> LGTM, we can move the todo to an issue maybe?

I left it as a comment because I wasn't sure if this is something we're interested about. IIRC social logins also have this problem, we only have the email. Should I create an issue?
> 
> P.S. Not tested yet as I am working on kind locally and UI is not building currently.

UI should work now, let me know if it doesn't

> P.S2. I need rebase.

Done 👍

diegodelemos:
> I left it as a comment because I wasn't sure if this is something we're interested in. IIRC social logins also have this problem, we only have the email. Should I create an issue?

@tiborsimko

PR #259, opened 2020-06-23
rest: add ui config endpoint

mvidalgarcia:
closes reanahub/reana-ui#103

### DIFF
--- a/reana_server/rest/config.py
+++ b/reana_server/rest/config.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2020 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Reana-Server config-functionality Flask-Blueprint."""
+
+import logging
+import traceback
+
+from flask import Blueprint, jsonify
+
+from reana_server.config import (
+    REANA_UI_ANNOUNCEMENT,
+    REANA_UI_POOLING_SECS,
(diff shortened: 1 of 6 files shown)


reana_server/rest/config.py
```diff
@@ -0,0 +1,72 @@
...
+
+from flask import Blueprint, jsonify
+from reana_commons.config import REANAConfig
+
+
+blueprint = Blueprint("config", __name__)
+
+
+@blueprint.route("/config", methods=["GET"])
+def get_config():
```
diegodelemos, line 22: Right now we have what seems a generic endpoint which returns REANA-UI only config. Therefore, it can be a bit misleading, I see two solutions:
1. We return a more complex object, where each key is the type of config, and inside there is the specific config payload. Example (I am just making up some config to illustrate):
```json
{
  "ui": {
    "local_users": true,
    ...
  },
  "cli": {
    "colored_output": "true"
    ...
  }
}
```
2. Take a parameter in the endpoint which determines which kind of config should be retrieved.

reana_server/rest/config.py
```diff
@@ -0,0 +1,72 @@
...
+          description: >-
+            Request failed. Internal server error.
+          examples:
+            application/json:
+              {
+                "message": "Internal server error."
+              }
+    """
+    try:
+        return (
```
diegodelemos, line 66: If we would also use this endpoint to retrieve config through the REST API (e.g. `reana-client`), I think it would make sense to add token auth. cc @tiborsimko

mvidalgarcia: The problem is that some of this configuration is needed before been signed in, such us `local_users` and `cern_sso`.

PR #260, opened 2020-07-06
setup: pip-tools based installation

tiborsimko:
Pins all dependencies via pip-compile and amends installation procedures
to used pinned packages. The pagkage varsions will be updated from time
to time via dedicated upgrade campaigns. Addresses reanahub/reana#322.

Signed-off-by: Tibor Šimko <tibor.simko@cern.ch>


Dockerfile
```diff
@@ -14,29 +14,30 @@ RUN apt-get update && \
...
 WORKDIR /code
-RUN pip install --no-cache-dir requirements-builder && \
-    requirements-builder -l pypi setup.py | pip install --no-cache-dir -r /dev/stdin && \
-    pip uninstall -y requirements-builder
-
 COPY . /code
 
-# Debug off by default
+# Install dependencies
+RUN pip install -r requirements.txt
```
tiborsimko, line 22: Note that this was not optimal WRT docker layer caching: changing anything in the source code would trigger reinstalling of dependencies. I'll force-push an update here and elsewhere.

PR #261, opened 2020-07-07
tests: move to containerised postgres DB

mvidalgarcia:
partially addresses reanahub/reana#348
depends on https://github.com/reanahub/pytest-reana/pull/63 https://github.com/reanahub/reana-db/pull/73

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -1,17 +1,73 @@
-#!/bin/sh
+#!/bin/bash
 #
 # This file is part of REANA.
-# Copyright (C) 2017, 2018 CERN.
+# Copyright (C) 2017, 2018, 2019, 2020 CERN.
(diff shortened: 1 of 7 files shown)


run-tests.sh
```diff
@@ -1,17 +1,73 @@
...
+check_black() {
+    echo '==> [INFO] Checking Black compliance...'
+    black --check .
+}
+
+pydocstyle reana_server
+check_black
+FLASK_APP=reana_server/app.py python ./scripts/generate_openapi_spec.py
+diff -q -w temp_openapi.json docs/openapi.json
+
```
tiborsimko, line 64: In place of empty newline, add `rm temp_openapi.json` statement as in `reana-workflow-controller`.

PR #263, opened 2020-07-13
rest: pagination

### LINKED ISSUE #180: implementation: pagination
Given that the list of workflows can be really long (100+), pagination over the workflows list is handy. This pagination could be implemented either in the server side, or in the client one.

This issue is to discuss the final implementation.

mvidalgarcia:
closes #180


docs/openapi.json
```diff
@@ -797,6 +797,20 @@
...
+            "name": "page",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "description": "Number of results per page (pagination).",
+            "in": "query",
+            "name": "size",
+            "required": false,
+            "type": "string"
```
diegodelemos, line 813: Totally minor comment, not need to change. Shouldn't this and the previous parameter be integers? UI doesn't care about this, and REANA-Client doesn't use this (the difference would be to have a 1 as string or as int in the REANA-Client Python code).

mvidalgarcia: IIRC I left it as string because reana-client was sending it as string.. I can have another look

mvidalgarcia: Fixed and r-commons updated.

PR #267, opened 2020-07-24
docs: single-page RTFD outline

mvidalgarcia:
addresses reanahub/reana#335


CHANGES.rst
```diff
@@ -2,7 +2,7 @@ Changes
 =======
 
 Version master (UNRELEASED)
----------------------------
+***************************
```
tiborsimko, line 5: I'm surprised this would be necessary... According to [sphinx docs](https://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html#headings), we are using headings OK, i.e. CHANGES is underlined via `=` (as in Sample H3), so Versions should be underlined via `-` as in Sample H4.

Is this something that the theme changed? If so, perhaps we should customise it there...

docs/index.rst
```diff
@@ -1,17 +1,67 @@
...
 
-   introduction
-   cli
-   restapi
-   contributing
-   changes
-   license
-   authors
+.. include:: ../README.rst
+   :start-line: 32
```
tiborsimko, line 11: We could put a starting one-line sentence:

```
Features:

- ...
```

so that the itemised list would be introduced by something.

docs/index.rst
```diff
@@ -1,17 +1,67 @@
...
+   :end-before: Useful links
+
+**Contents**
```
tiborsimko, line 14: Consider killing Contents, to keep up with 'single-autogen-page-and-everything-important-in-docs-reana-io?

docs/index.rst
```diff
@@ -1,17 +1,67 @@
...
+
+.. contents::
+   :local:
+   :depth: 1
+   :backlinks: none
+
+
```
tiborsimko, line 21: Before CLI API, you can introduce Installation section, something like:

```
Installation
---------------

Please follow detailed _REANA cluster installation guide_.
```

pointing to docs.reana.io.  The text could be the same for all cluster components.

docs/index.rst
```diff
@@ -1,17 +1,67 @@
...
+   :members:
+
+.. automodule:: reana_server.rest.users
+   :members:
+
+.. automodule:: reana_server.rest.workflows
+   :members:
+
+
+.. include:: ../CONTRIBUTING.rst
```
tiborsimko, line 47: We can simplify Contributing by killing difference between Issues and Pull requests in order to have one simple wall of text.  We can do it by editing CONTRIBUTING.rst and killing sections there. We could reword beginning of paragraphs to be:

- Bug reports, issues, feature requests, and other contributions are...

- If you create a pull request fixing a bug or implementing a feature, you can...

docs/index.rst
```diff
@@ -1,17 +1,67 @@
...
+
+.. include:: ../CHANGES.rst
```
tiborsimko, line 49: Please put CHANGES before CONTRIBUTING.  The thinking goes to group all admin-usage sections (such as CLI API, REST API, change list) before developer-usage sections (issues, pull requests, license, authors).

docs/index.rst
```diff
@@ -1,17 +1,67 @@
...
+granted to it by virtue of its status as an Intergovernmental Organization or
+submit itself to any jurisdiction.
+
+.. include:: ../AUTHORS.rst
+
+
+Useful links
+============
+
+.. include:: ../README.rst
```
tiborsimko, line 66: Hmm, the `documentation` link from README.rst would be self-referencing.  We should probably amend the link to point to docs.reana.io... Or simply kill the thing here in rtfd and end with author list. That's probably better. This is because people who are here already know where the source code is, and we link to issues above, and list of releases is not terribly interesting since end users would use Helm releases anyway... So let's end with author list I guess.

docs/index.rst
mvidalgarcia: Do you think it's better to kill it? I find it useful to navigate among sections, even if it's single page.

docs/index.rst
mvidalgarcia: Do you mean removing the whole "Useful links" section or only the bullet-point links?

CHANGES.rst
mvidalgarcia: I've changed all the titles to match sphinx docs and it works.

docs/index.rst
tiborsimko: Yeah, in the dictdiffer style.  If the top of the page is short regarding Installation and Usage, it would sort of "drive people away" towards docs.reana.io, which is actually a good thing. The users will know more easily that this page is just an auto-generated reference, that probably only a minority would be interested in. That kind of impression.  We can see how it all looks like once

CHANGES.rst
tiborsimko: Interesting :wink: Looks nicer. BTW some of the rst page titles are still using different styles, e.g. README uses hashes, CONTRIBUTING uses asterisks, AUTHORS is only equal-sign-underlined. Would it be possible to harmonise that?

```console
$ head -4 *.rst
==> AUTHORS.rst <==
Authors
=======

The list of contributors in alphabetical order:

==> CHANGES.rst <==
*******
Changes
*******


==> CONTRIBUTING.rst <==
************
Contributing
************


==> README.rst <==
############
REANA-Server
############
```

mvidalgarcia: Yes, all of them are harmonized now except the README's that should be h1.

Issue #118, opened 2019-01-17
scheduler: assess requeueing procedure


dinosk:
The `scheduler.on_message` method is executed with every new incoming message. The conditions in set in the config.py are checked and if all are met, RWC is called to start the workflow run. If not the message is requeued so that it retries later. The kombu docs mention this as not the proper way to select what messages to process http://docs.celeryproject.org/projects/kombu/en/latest/reference/kombu.message.html#kombu.message.Message.requeue . The current procedure works, but more testing could be done with a larger number of workflow submissions, to ensure that problems like starvation or head of line blocking (shouldn't be the case for this, but relevant in other approaches) are avoided.

Stemmed from https://github.com/reanahub/reana-server/pull/116#discussion_r248757108

diegodelemos:
Moreover, this requeuing mechanism doesn't respect order, [workflows are arbitrarily sent back to the queue](https://github.com/reanahub/reana-server/blob/e15e306806118f0232ef03740ef8ecf9a4550673/reana_server/scheduler.py#L112). This causes for example that in a situation in which W1 is submitted before W2, there is a chance that W2 would get executed first.

This is because one has to "acknowledge" a workflow to be able to inspect it and decide whether to run it or not [source here](https://github.com/reanahub/reana-server/blob/e15e306806118f0232ef03740ef8ecf9a4550673/reana_server/scheduler.py#L133). This requires RabbitMQ/Kombu investigation.

audrium:
Priority scheduling was introduced in https://github.com/reanahub/reana-server/issues/358, so to improve the requeueing procedure it was decided to use a message delay mechanism (with the help of [RabbitMQ Delayed Message Plugin](https://github.com/rabbitmq/rabbitmq-delayed-message-exchange)) and to count how many times the message was requeued. This could later be improved to transfer messages which are looping for a long time to some [dead letter exchange](https://www.rabbitmq.com/dlx.html) and to inform a user that a workflow has expired.

PR #277, opened 2020-09-14
rest: catch email exception on token request endpoint

audrium:
addresses: #118

### DIFF
--- a/reana_server/rest/users.py
+++ b/reana_server/rest/users.py
@@ -16,6 +16,7 @@ from flask import Blueprint, jsonify, request
 from flask_login import current_user
 from reana_db.models import AuditLogAction
 from reana_commons.email import send_email
+from reana_commons.errors import REANAEmailNotificationError
 
 from reana_server.config import ADMIN_EMAIL, REANA_URL
 from reana_server.utils import (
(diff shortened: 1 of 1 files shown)


reana_server/rest/users.py
```diff
@@ -427,7 +428,11 @@ def request_token():
         email_body = "New user access token request:\n\n" + "\n".join(
             [f"{f}: {getattr(user, f, None)}" for f in fields]
         )
-        send_email(ADMIN_EMAIL, email_subject, email_body)
+        try:
+            send_email(ADMIN_EMAIL, email_subject, email_body)
+        except REANAEmailNotificationError:
+            pass
```
tiborsimko, line 434: Instead of pass, could we log the error? Perhaps just printing it out to stdout/stderr, so that admins would see it when doing:

```
$ kubectl logs reana-server-5c5cdf658-ghv2z rest-api
```

tiborsimko: Also, currently, the error says:

```python
        raise REANAEmailNotificationError(
41-            "Cannot send email, missing server and port configuration. "
42-            "Please provide the following environment variables:\n"
43:            "REANA_EMAIL_SMTP_SERVER\nREANA_EMAIL_SMTP_PORT"
44-        )
```

If it could appear in User Web UI -- e.g. if we would merge only `r-ui` PR and not this one -- then we could reformulate it to something more user-friendly, like:

```
Error sending notification email. Missing REANA_EMAIL_SMTP_SERVER
and REANA_EMAIL_SMPT_PORT configuration. 
```

because users cannot "provide" it.

... that said, it would be quite a big change in the code base to both show the error and proceed with the user payload... we would probably need to change API response types... so if we only log this event on the server side, that might do, even without changing the error message.

tiborsimko (approved):

PR #288, opened 2020-09-30
rest: return CPU usage

diegodelemos:
Closes reanahub/reana-workflow-controller#332.


reana_server/rest/users.py
```diff
@@ -333,6 +343,7 @@ def get_you():
                         "full_name": me.full_name,
                         "username": me.username,
                         "disk_usage": me.get_user_disk_usage(),
+                        "cpu_usage": me.get_user_cpu_usage(),
```
audrium, line 346: A bit off topic question - I wonder when we should combine quotas into some kind of dict (maybe something similar like [here](https://github.com/reanahub/reana-client/pull/425/files#diff-4f27f08c3f46f0b755c1d5ac1a1e2dcaR72)) and group the usage and limits of different quota resources. Maybe we should create an issue for that?

diegodelemos: I think this is a great idea! Maybe we can do a little refactor once we merge these PRs so we do not block the rest? Can you create a new issue for this from your comment?

reana_server/utils.py
```diff
@@ -99,6 +99,7 @@ def _create_user(email, user_access_token, admin_access_token):
         user_parameters = dict(access_token=user_access_token)
         user_parameters["email"] = email
         user = User(**user_parameters)
+        user.initialize_user_quota_limits()
```
mvidalgarcia, line 102: Is this operation going to be needed **every time** we create a user? If that is the case, would it be possible to move it inside the `User` model constructor?

What about the users created via UI? Are we supposed to do this initialization manually? (Sorry if this was already discussed, I probably missed it)

docs/openapi.json
```diff
@@ -665,6 +667,9 @@
                   "access_token": {
                     "type": "string"
                   },
+                  "cpu_usage": {
+                    "type": "string"
```
mvidalgarcia, line 671: Maybe it doesn't belong to this PR as I see it's the same for `disk_usage` but,  shouldn't this be an integer?

reana_server/utils.py
diegodelemos: Thanks for spotting these things!

I agree 💯 %  with you. This is one of the few things I am not super happy about... The problem is that this `initialize_user_quota_limits` needs the User to be committed first... that's why initially I just created the method and went this way, so I would wrap up. I will have a look and see if I can figure it out.

Regarding the UI I overlooked it thinking that `_create_user` would be used there but actually [it isn't](https://github.com/reanahub/reana-server/blob/c4eb0f9a317a9701538e28b4abc25e0d7f6abb9b/reana_server/utils.py#L167), I will fix it.

docs/openapi.json
diegodelemos: Yes! I see it as something to deal with while doing https://github.com/reanahub/reana-server/issues/289, what do you think?

reana_server/utils.py
mvidalgarcia: Yeah.. too bad that we're not using the same function for CLI/API and UI creation of users... Maybe we can unify in the future?

mvidalgarcia (approved):

PR #290, opened 2020-10-06
rest: offer overall user quota consumption

### LINKED ISSUE #289: rest: combine usage and limits of different quota resources into dict
We should combine quotas into some kind of dict (maybe something similar like [here](https://github.com/reanahub/reana-client/pull/425/files#diff-4f27f08c3f46f0b755c1d5ac1a1e2dcaR72)) and group the usage and limits of different quota resources before returning the response. 

_Originally posted by @audrium in https://github.com/reanahub/reana-server/pull/288#discussion_r498788299_

diegodelemos:
* Closes #289.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -603,187 +603,6 @@
         "summary": "Requests a new access token for the authenticated user."
       }
     },
-    "/api/users": {
-      "get": {
-        "description": "Get user information.",
-        "operationId": "get_user",
-        "parameters": [
-          {
-            "description": "Not required. The email of the user.",
-            "in": "query",
-            "name": "email",
-            "required": false,
-            "type": "string"
-          },
-          {
-            "description": "Not required. UUID of the user.",
-            "in": "query",
-            "name": "id_",
-            "required": false,
-            "type": "string"
-          },
-          {
-            "description": "Not required. API key of the admin.",
-            "in": "query",
-            "name": "user_token",
-            "required": false,
-            "type": "string"
-          },
-          {
-            "description": "Required. API key of the admin.",
-            "in": "query",
-            "name": "access_token",
-            "required": true,
-            "type": "string"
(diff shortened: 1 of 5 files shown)


docs/openapi.json
```diff
@@ -647,17 +647,33 @@
...
                   "access_token": "Drmhze6EPcv0fN_81Bj-nA",
-                  "cpu_usage": "70536",
-                  "disk_usage": "408",
                   "email": "user@reana.info",
-                  "id": "00000000-0000-0000-0000-000000000000"
+                  "id": "00000000-0000-0000-0000-000000000000",
+                  "quota": {
+                    "cpu": {
+                      "limit": "200000",
+                      "usage": "70536"
```
mvidalgarcia, line 655: Shouldn't it be an int if it's type "number"?

reana_server/rest/users.py
```diff
@@ -148,98 +176,6 @@ def get_user():  # noqa
         return jsonify({"message": str(e)}), 500
 
 
-@blueprint.route("/users", methods=["POST"])
-def create_user():  # noqa
```
mvidalgarcia, line 152: Any specific reason why we get rid of this? Maybe we can specify it in the commit message.

docs/openapi.json
diegodelemos: That's what I thought, but then I realised we won't be able to fit big integers into integers... so we are left with two options: `string` or `number` without format (representing "any number" see [here](https://swagger.io/docs/specification/data-models/data-types/#numbers)).

reana_server/rest/users.py
diegodelemos: Mainly because we do not use it. I will specify in the commit message, thanks for pointing out :)

mvidalgarcia (approved):

Issue #296, opened 2020-10-13
quotas: quota unit conversion server side


diegodelemos:
Currently, we only [return quota usage numbers](https://reana-server.readthedocs.io/en/latest/_static/api.html#operation/get_you) in [canonical units](https://github.com/reanahub/reana-db/blob/08724470206675a54dc65b71b469d8ba0c18e40d/reana_db/models.py#L635-L639).

We should provide a way of getting the quota usage in human-readable ways (inside the `/api/you` and from inside Python calls) providing the number and the unit.

This feature will be used by:
- [Users: using REANA-UI's your profile page (through `/api/you`)](https://github.com/reanahub/reana-ui/issues/138)
- [Users: using REANA-Client (through `/api/you`)](https://github.com/reanahub/reana-client/issues/424)
- [Admins: consulting quota (through Python API)](https://github.com/reanahub/reana-server/issues/284)
- [Admins: reading REANA system status report (through Python API)](https://github.com/reanahub/reana-server/issues/286)

To enable this change, we need to:
- Disk: Rethink the implementation of [REANA-Commons' `get_disk_usage`](https://github.com/reanahub/reana-commons/blob/e6cdbedb62b21cc20407512e7d0a09eeb28f9481/reana_commons/utils.py#L192-L223) function to:
  - Provide a default human-readable value/unit (the one that makes more sense e.g. if the usage is `54000` bytes, return `54` KB and not `0.054` MB).
  - Offer a way of getting the usage in the requested unit (e.g. MB, GB etc..)
  The function's current problems are:
    - Based on `du` which causes incompatibilities between Mac and Linux (GNU `du` and BSD `du`)
      - This can be solved by adding a platform check and depending on the platform type use the correct command options
    - The return value is inconsistent: when a specific unit is required (through `block_size` param), the return value is a number. However, when no unit is required the return value is a string composed of the number and a suffix with the unit e.g. `334K`
       - This could be tackled by returning a tuple/dictionary, splitting the value and its unit (regardless if it was requested or it was auto-discovered).

    Whether we continue using `du` as the underlying implementation to calculate disk usage or not is indifferent, we should meet the requirement.
    
- CPU: Implement a new function to give human-readable usage times in REANA-Commons. It should 
    - Provide the most human-readable format for a given usage (e.g. `2 days and 3 hours`)
    - Offer a way of getting the usage in the requested unit (e.g. days, hours)

It would be great if both implementations implement a similar or even better, the same interface.

[...]

PR #301, opened 2020-10-19
rest: simplify disk usage format

diegodelemos:
* Shows disk usage format either in raw (bytes) or humand-friendly
  format (closes reanahub/reana-server#296).

### DIFF
--- a/reana_server/reana_admin.py
+++ b/reana_server/reana_admin.py
@@ -27,6 +27,7 @@ from reana_db.config import DEFAULT_QUOTA_LIMITS
 from reana_db.database import Session
 from reana_db.models import (
     AuditLogAction,
+    QuotaHealth,
     Resource,
     User,
     UserResource,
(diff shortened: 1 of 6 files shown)


mvidalgarcia (approved):

mvidalgarcia:
```
$ flask reana-admin quota-usage
User could not be retrieved: 
None
```

Fails when we have users with no limits, [the key doesn't exist](https://github.com/reanahub/reana-server/blob/b7c0b0fd6a44fbddc3cac9104da053264b33ceca/reana_server/reana_admin.py#L440-L441), because we don't [calculate the health](https://github.com/reanahub/reana-db/pull/98/files#diff-9c9ecb5b02eee5df82a9c1138c56e62181e9d8d56152d45c09f7fe0a693ce980R103).

diegodelemos:
To get this going I've changed the implementation so when there is no `health` returned, the user quota is displayed as healthy (in green). This is not correct and will have to be addressed in a separate issue.

reana_server/reana_admin.py
```diff
@@ -435,9 +435,8 @@ def list_quota_usage(
                     disk.get("limit", {}).get(human_readable) or "-",
                 )
             )
-
             health_ordered = max(
-                [disk.get("health"), cpu.get("health")],
+                [disk.get("health", "healthy"), cpu.get("health", "healthy")],
```
mvidalgarcia, line 439: Would it be ok to import it from the `QuotaHealth` enum?

PR #303, opened 2020-10-21
cli: add command to set limit per user & resource

### LINKED ISSUE #302: cli: reana-admin quota-set command
> (iii) offer reana-admin quota-set command that admins can use to bump limits or reduce limits for various users; (iv) write alembic recipe that updates all past users based on default values or that simply prints that admins should run this-and-this reana-admin command

_Originally posted by @tiborsimko in https://github.com/reanahub/reana/issues/425#issuecomment-712268049_

mvidalgarcia:
closes #302

```console
$ flask reana-admin quota-resources
processing time
shared storage
$ flask reana-admin quota-set  --email johndoe@example.org  --resource-name 'fake resource' --limit 15000
ERROR: Provided resource name does not exist. Available resources are ['processing time', 'shared storage'].
$ flask reana-admin quota-set  --email johndoe@example.org  --resource-name 'processing time' --limit 15000
Quota limit successfully set. 
```

##### TODO

- [x] New command (eval-style) to migrate all users that don't have quota limits set (`UserResource`) to have the default limits.

### DIFF
--- a/reana_server/reana_admin.py
+++ b/reana_server/reana_admin.py
@@ -23,6 +23,7 @@ from reana_commons.config import REANAConfig, REANA_RESOURCE_HEALTH_COLORS
 from reana_commons.email import send_email
 from reana_commons.errors import REANAEmailNotificationError
 from reana_commons.utils import click_table_printer
+from reana_db.config import DEFAULT_QUOTA_LIMITS
 from reana_db.database import Session
 from reana_db.models import (
     AuditLogAction,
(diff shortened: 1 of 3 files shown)


reana_server/reana_admin.py
```diff
@@ -464,3 +471,96 @@ def list_quota_usage(
...
+@admin_access_token_option
+@click.pass_context
+def set_default_quota_limit(ctx, admin_access_token):
+    """Set default quota limits to users that do not have any."""
+    output_commands = []
+    users_without_quota_limits = User.query.filter(~User.resources.any()).all()
+    for user in users_without_quota_limits:
+        for resource in Resource.query:
+            output_commands.append(
+                f"flask reana-admin quota-set -e {user.email} -r '{resource.name}' "
```
diegodelemos, line 556: This turns out to be very slow as Python+Click has to be initialised for every user. Also, this would be very hard on the DB as each command would be instantiating a new DB session.

A solution for this could be to make `flask reana-admin quota-set` accept a list of ids/emails and perform the quota set by invoking the command from Python passing `users_without_quota_limits ` to it. Another option is to share part of the implementation with `quota-set` (extract to a function) and do everything in pure Python/SQLAlchemy and forget about invoking with Click.

diegodelemos (commented):
Works like a charm, it is ready to be merged; we just need to:
- Release REANA-DB
  - Observation: There are [3 PRs currently in review which contain changes for REANA-DB](https://github.com/orgs/reanahub/projects/34))
- Make Travis pass

For all of this we need green light to be able to merge to master from @tiborsimko.

PR #306, opened 2020-10-26
cli: add quota usage to system status [skip ci]

### LINKED ISSUE #286: system-status: enrich status information with quota usage
With the outcome of https://github.com/reanahub/reana-server/issues/284 and https://github.com/reanahub/reana-db/issues/83, enrich the [daily status report command output](https://github.com/reanahub/reana-server/blob/2f389203c1d3ec5ace1fbf63eed3367a653eac20/reana_server/reana_admin.py#L345) (created in https://github.com/reanahub/reana-server/issues/234)

alintulu:
Add quota usage to system status.

closes #286

### DIFF
--- a/reana_server/status.py
+++ b/reana_server/status.py
@@ -14,7 +14,18 @@ from datetime import datetime, timedelta
 from invenio_accounts.models import SessionActivity
 from reana_commons.config import SHARED_VOLUME_PATH
 from reana_db.database import Session
-from reana_db.models import Workflow, RunStatus, InteractiveSession
+from reana_db.models import (
+    Workflow,
+    RunStatus,
+    InteractiveSession,
+    User,
+    Resource,
+    UserResource,
+    ResourceType,
+    ResourceUnit,
+)
+from sqlalchemy import func
+from sqlalchemy.orm import aliased
 
 
 class REANAStatus:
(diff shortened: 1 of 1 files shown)


reana_server/status.py
```diff
@@ -251,11 +262,96 @@ def get_status(self):
...
+        :param user: A REANA-DB user model.
+        :type from_: datetime
+        :type until: datetime
+        :type user: reana_db.models.User
+        """
+        super().__init__(from_=from_, until=until, user=user)
+
+    def get_top_five_all(self, resource_type):
+        """Get the top five users according to quota usage."""
+        subq = (
```
diegodelemos, line 284: What do you think of renaming these variables to more explicit names? Like `user_resources_subq ` and `top_five_resource_usage_users`?

reana_server/status.py
```diff
@@ -251,11 +262,96 @@ def get_status(self):
...
+            user.email: {
+                "id": str(user.id),
+                "used": ResourceUnit.human_readable_unit(
+                    resource.resource.unit, resource.quota_used
+                ),
+                "limit": ResourceUnit.human_readable_unit(
+                    resource.resource.unit, resource.quota_limit
+                ),
+            }
+            for user, resource in users
```
diegodelemos, line 307:
```suggestion
            for user, user_resource in users
```

reana_server/status.py
```diff
@@ -251,11 +262,96 @@ def get_status(self):
...
+        }
+
+    def get_top_five_limited(self, resource_type):
+        """Get the top five users according to quota usage per quota limit."""
```
diegodelemos, line 311:
```suggestion
        """Get the top five users with quota limit according to quota usage."""
```

reana_server/status.py
```diff
@@ -251,11 +262,96 @@ def get_status(self):
...
+            .filter(Resource.type_ == resource_type)
+            .subquery()
+        )
+        subq_alias = aliased(UserResource, subq)
+        users = (
+            Session.query(User, subq_alias)
+            .join(subq_alias, UserResource)
+            .order_by((UserResource.quota_used / UserResource.quota_limit).desc())
+            .limit(5)
+        )
```
diegodelemos, line 325: Minor comment: What do you think of extracting this logic to a private method?

reana_server/status.py
```diff
@@ -251,11 +262,96 @@ def get_status(self):
...
+            user.email: {
+                "id": str(user.id),
+                "used": ResourceUnit.human_readable_unit(
+                    resource.resource.unit, resource.quota_used
+                ),
+                "limit": ResourceUnit.human_readable_unit(
+                    resource.resource.unit, resource.quota_limit
+                ),
+            }
+            for user, resource in users
```
diegodelemos, line 336:
```suggestion
            for user, user_resource in users
```

diegodelemos (commented):
Only code reviewed, not tested locally.

reana_server/status.py
alintulu: @diegodelemos I agree! I moved the query to a private method, what do you think?

reana_server/status.py
```diff
@@ -251,11 +262,95 @@ def get_status(self):
...
+        :type from_: datetime
+        :type until: datetime
+        :type user: reana_db.models.User
+        """
+        super().__init__(from_=from_, until=until, user=user)
+
+    def get_top_five_resource_usage_users(
+        self, resource_filter, order, user_resource_filter=True
+    ):
+        user_resource_subq = (
```
diegodelemos, line 285: After testing locally doing the following change in `helm/configurations/values-dev.yaml` in reanahub/reana
```diff
+notifications:
+  enabled: true
+  system_status: "*/3 * * * *"
+  email_config:
+    receiver: admins@reana.cern.ch
+    sender: notifications@reana.cern.ch
```
and creating a few users and running workflows.

I see:
image

I've spotted a couple of things:
- Only 3 users are shown instead of five
- The aggregation is not completely correct, because UserResource might have several resources for a user of a different ResourceType, so we need to group by the resource type and do a `sum` (which corresponds currently to [this bit](https://github.com/reanahub/reana-db/blob/c03aa75fed2b0c5d3aa4cc3555bf72ac3a3cf6b0/reana_db/models.py#L156-L166)).

Maybe the best to find a solution for these to details is to write some unit tests, what do you think? It would be definitely faster than trying with the cluster :).

alintulu: Good idea! I can definitely write some unit tests. I had completely missed that one user can have several resources of the same type. By the way, why is this?

diegodelemos: This is a great question. We've chosen to have a lot of flexibility so we can have [Resources](https://github.com/reanahub/reana-db/blob/a7b7753066e57b93720cd9cbb4db0137a68c4f88/reana_db/models.py#L756) like "meyrin cpu", "general cpu", "xxx cpu" and all of them of [ResourceType](https://github.com/reanahub/reana-db/blob/a7b7753066e57b93720cd9cbb4db0137a68c4f88/reana_db/models.py#L697) `cpu`. So when we want to know how much CPU usage a given user has we need to go through all the kinds.

We have discussed with @mvidalgarcia about this flexibility making things quite complicated in general. Maybe we can simplify the problem by asking ourselves the following two questions:
- Do we really need this flexibility CC @tiborsimko 
- Do we have the simplest design to achieve it (if we really need it)

alintulu: I see, thank you for explaining!

alintulu: @diegodelemos with your suggested changes to `helm/configurations/values-dev.yaml` do I need to rebuild the cluster or can I apply the values to an existing cluster while keeping the existing users?

alintulu: Addressing your comments and my fixes:

> Only 3 users are shown instead of five

I increased the limit from 5 to 10 and now it works as expected. I think it is due to `self.get_top_five_resource_usage_users()` returning an array of tuples, each tuple containing two objects. It seems like `limit()` is not computed over each item in the array but includes the items in the tuples.

> The aggregation is not completely correct, because UserResource might have several resources for a user of a different ResourceType, so we need to group by the resource type and do a sum (which corresponds currently to this bit).

By adding `group_by()` I was able to utilise the `func.sum()` method.

diegodelemos: > @diegodelemos with your suggested changes to helm/configurations/values-dev.yaml do I need to rebuild the cluster or can I apply the values to an existing cluster while keeping the existing users?

You just need to redeploy and everything should work as expected:
```console
$ reana-dev cluster-undeploy
$ kubectl get pods -w  # wait for all pods to disappear
$ reana-dev cluster-deploy --exclude-components=r-a-vomsproxy,r-a-krb5 --admin-email admin@cern.ch --admin-password 123456  --mode debug
```

Regarding the amendments, I will shortly retest your changes, thanks for updating!

diegodelemos (approved):

PR #308, opened 2020-10-28
rest: prevent workflows from running when quota is exceeded

### LINKED ISSUE #282: api: prevent workflows from start if the user exceeds quota
Prevent users from submitting new workflows when they have exceeded their quota.

The maximum quota can be:
- Centrally configured - MVP
  - For "institutional" users - @cern.ch users in our case (or configured list of institutional users e.g. [@cern.ch, @nyu.edu])
  - For "outside" users - in our concrete case non @cern.ch accounts, in general accounts that are not in the list of institutional users
  
    For example (note that naming is for illustration and it is not the best as we may need to use guests for real guests accounts, not logged in):
   ```console
    USERS_EMAILS = ["@cern.ch", "@nyu.edu"]
    USER_MAX_CPU_QUOTA = xyz
    GUEST_MAX_CPU_QUOTA = abc
    ```
- Set by user/group through admin interface (future)

Also, we have two moments in which a workflow can be prevented from running:
- When it is submitted ([REST API level](https://github.com/reanahub/reana-server/blob/2f389203c1d3ec5ace1fbf63eed3367a653eac20/reana_server/rest/workflows.py#L757)) (MVP)
- When [it is in the queue and it is going to be sent for execution to RWC](https://github.com/reanahub/reana-server/blob/2f389203c1d3ec5ace1fbf63eed3367a653eac20/reana_server/scheduler.py#L78)

### LINKED ISSUE #283: api: prevent sessions from start if the user exceeds quota
With the outcome of https://github.com/reanahub/reana-server/issues/282 we should be able to add this prevention mechanism to the [session opening endpoint](https://github.com/reanahub/reana-server/blob/2f389203c1d3ec5ace1fbf63eed3367a653eac20/reana_server/rest/workflows.py#L1743).

mvidalgarcia:
closes #282
closes #283

### DIFF
--- a/reana_server/decorators.py
+++ b/reana_server/decorators.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2020 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA Server function decorators."""
+
+import functools
+import logging
+import os
+import traceback
+
+import click
+from flask import jsonify, request
+from flask_login import current_user
+from reana_commons.errors import REANAQuotaExceededError
+
+from reana_server.utils import _get_user_from_invenio_user, get_user_from_token
+
+
+def admin_access_token_option(func):
+    """Click option to load admin access token."""
+
+    @click.option(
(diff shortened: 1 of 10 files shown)


mvidalgarcia:
Tests failing due to missing `r-db` and `r-commons` releases.

diegodelemos (approved):

PR #312, opened 2020-11-03
rest: return reana_server url

ParthS007:
closes https://github.com/reanahub/reana-client/issues/441

### DIFF
--- a/reana_server/rest/users.py
+++ b/reana_server/rest/users.py
@@ -18,6 +18,7 @@ from reana_db.models import AuditLogAction
 from reana_commons.email import send_email
 from reana_commons.errors import REANAEmailNotificationError
 
+from reana_server import __version__
 from reana_server.config import ADMIN_EMAIL, REANA_HOSTNAME
 from reana_server.utils import (
     _create_user,
@@ -249,6 +250,8 @@ def get_you():
             properties:
               email:
                 type: string
+              reana_server_version:
+                type: string
               reana_token:
                 type: object
                 properties:
@@ -262,6 +265,7 @@ def get_you():
             application/json:
               {
                 "email": "user@reana.info",
+                "reana_server_version": "0.7.0",
                 "reana_token": {
                     "value": "Drmhze6EPcv0fN_81Bj-nA",
                     "status": "active",
@@ -311,6 +315,7 @@ def get_you():
                 jsonify(
                     {
                         "email": me.email,
+                        "reana_server_version": __version__,
                         "reana_token": {
                             "value": me.access_token,
                             "status": me.access_token_status,
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -2466,6 +2466,7 @@
               "application/json": {
                 "email": "user@reana.info",
                 "full_name": "John Doe",
+                "reana_server_version": "0.7.0",
                 "reana_token": {
                   "requested_at": "Mon, 25 May 2020 10:39:57 GMT",
                   "status": "active",
@@ -2477,6 +2478,9 @@
             "schema": {
               "properties": {
                 "email": {
+                  "type": "string"
+                },
+                "reana_server_version": {
                   "type": "string"
                 },
                 "reana_token": {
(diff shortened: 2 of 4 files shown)


reana_server/rest/users.py
```diff
@@ -177,6 +181,7 @@ def get_you(user):
                 jsonify(
                     {
                         "email": user.email,
+                        "reana_server_version": __version__,
```
diegodelemos, line 184: This might be more complicated than returning `reana_server.__version__`, the correct version of REANA-Server is necessary but not enough to guarantee compatibility with a given `reana-client` version.

The REANA cluster version with which a `reana-client` version works is not just a specific REANA-Server version, it is rather the combination of all [REANA cluster components/packages](https://github.com/reanahub/reana/blob/b962864f6fe23a2864053b7d047aa1ad33e6572a/reana/config.py#L63-L80) versions and this is represented in the [REANA Helm chart](https://github.com/reanahub/reana/blob/b962864f6fe23a2864053b7d047aa1ad33e6572a/helm/reana/Chart.yaml#L29).

For example, now we released `0.7.0`, so one would install REANA `0.7.0` on the server side (see the Helm repo https://reanahub.github.io/reana/) and all clients should be using `reana-client==0.7.x`.

Therefore, I would suggest to get this info from the Helm chart and rename the field `reana_server_version` to `cluster_version` or similar.

ParthS007: @diegodelemos  Should I get both `reana_server_version` and `reana_client_version` from the helm chart?

diegodelemos: `reana_client_version` you will get from your REANA-Client's `__version__.py`, that should be fine.

tiborsimko: Isn't it enough to just import `reana_server.__version__`?  For REANA 0.7.x helm charts, the server will be `0.7.y`, whatever the values of `x` and `y` may be. And this will always be guaranteed to work fine with `0.7.z` clients...So such a simple check might do?  (Thinking about a simple MVP solution not to complicate things. Here, even 0.8.0a123 and 0.8.0-alpha.456 would also be  OK, because nothing is really guaranteed in the alpha territory.)

tiborsimko: FWIW I'm seeing troubles locally using the following branches:

```console
$ reana-dev git-status -c ALL | grep -v '@ maint' | grep -v '@ master'
reana @ git-log-with-arguments @ c7cdbf4 cli: richer git-log command
reana-commons @ pr-231 (8 AHEAD 1 BEHIND upstream/pr/231)(STEMS FROM 26 BEHIND upstream/master) @ ed340a8 openapi: return reana_server_version
reana-server @ pr-312 (STEMS FROM 24 BEHIND upstream/master) @ 09a17b0 rest: return reana_server url
reana-client @ pr-442 (2 AHEAD 1 BEHIND upstream/pr/442)(STEMS FROM 11 BEHIND upstream/master) @ d80b1da cli: enrich ping command with version information
```

The `run-ci` works fine, but the `ping` command fails:
 
```
$ eval $(reana-dev client-setup-environment)
$ reana-client ping
Could not connect to the selected REANA cluster server at https://localhost:30443:
500 Internal Server Error
```

The error:

```
$ kubectl logs reana-server-c67755b9c-sqtv6 rest-api
...
ERROR:root:Traceback (most recent call last):
  File "./reana_server/rest/users.py", line 317, in get_you
    "email": user.email,
NameError: name 'user' is not defined

[pid: 8|app: 0|req: 4/45] 10.244.0.8 () {38 vars in 625 bytes} [Fri Nov  6 10:33:31 2020] GET /api/you?access_token=lh_EtqzC0D3bH-WFXbh6EA => generated 46 bytes in 6 msecs (HTTP/1.1 500) 13 headers in 547 bytes (1 switches on core 3)
ERROR:root:Traceback (most recent call last):
  File "./reana_server/rest/users.py", line 317, in get_you
    "email": user.email,
NameError: name 'user' is not defined
```

Probably related to rebasing to `maint-0.7`?

ParthS007: It was a rebasing issue. resolved and now working with `maint-0.7`.  👍

diegodelemos (approved):

PR #314, opened 2020-11-04
cli: skip quota usage check if only admin user

### LINKED ISSUE #313: cli: `flask reana-admin quota-usage` broken
```console
$ kubectl exec -i -t deployment/reana-server -c rest-api -- flask reana-admin quota-usage --admin-access-token $REANA_ACCESS_TOKEN
User could not be retrieved:
not enough values to unpack (expected 3, got 0)
```

alintulu:
Add check to see if the only user is the admin user. If so skip computing the quota usage.

closes #313

### DIFF
--- a/reana_server/reana_admin.py
+++ b/reana_server/reana_admin.py
@@ -412,9 +412,8 @@ def list_quota_usage(
         colours = []
         health = []
         for user in response:
-            if str(user.id_) == ADMIN_USER_ID:
-                continue
-            (disk, cpu,) = user.get_quota_usage().values()
+            quota_usage = user.get_quota_usage()
+            disk, cpu = quota_usage.get("disk"), quota_usage.get("cpu")
             data.append(
                 (
                     str(user.id_),
@@ -435,16 +434,17 @@ def list_quota_usage(
             colours.append(REANA_RESOURCE_HEALTH_COLORS[health_ordered])
             health.append(health_ordered)
 
-        data, colours, _ = (
-            list(t)
-            for t in zip(
-                *sorted(
-                    zip(data, colours, health),
-                    key=lambda t: health_order[t[2]],
-                    reverse=True,
+        if data and colours and health:
+            data, colours, _ = (
+                list(t)
+                for t in zip(
+                    *sorted(
+                        zip(data, colours, health),
+                        key=lambda t: health_order[t[2]],
+                        reverse=True,
+                    )
                 )
             )
-        )
 
         if output_format:
             tablib_data = tablib.Dataset()


alintulu:
@mvidalgarcia looks like a lot has been changed however only addition is `if len(response) > 1 or str(response.id_) != ADMIN_USER_ID:` to exclude cases with only the admin user.

However contrary to what I thought before I now see it might be useful to see the quota usage of the admin sometimes. What do you think? The bug was triggered when the `data`, `colours` and `health` lists were empty, which was the case when only the admin was a user. Hence the added check. However the bug could also be fixed by also checking for the admins quota usage. That way the three lists will never be empty since the admin is always present.

mvidalgarcia:
IMO it makes sense to have admin quotas displayed, in the end, admins are essentially users with superpowers and it's nice to see their usage.

alintulu:
OK, the latest commit should be the final. I had to add a check for the lists `data`, `colours` and `health`. Without it, it might break in the example that there is no user with email `test@reana.io` and you do

```console
$ flask reana-admin quota-usage -e test@reana.io
User could not be retrieved:
not enough values to unpack (expected 3, got 0)
```

Now it works as

```console
$ flask reana-admin quota-usage -e test@reana.io
ID   EMAIL   CPU-USED   CPU-LIMIT   DISK-USED   DISK-LIMIT
```

I tested it with the following use cases and it works as expected.

* Only admin user: `flask reana-admin quota-usage `
* Only admin user: `flask reana-admin quota-usage -e not_admin_email `
* Multiple users: `flask reana-admin quota-usage `
* Multiple users: `flask reana-admin quota-usage -e not_admin_email `

reana_server/reana_admin.py
```diff
@@ -435,16 +433,17 @@ def list_quota_usage(
...
             health.append(health_ordered)
 
-        data, colours, _ = (
-            list(t)
-            for t in zip(
-                *sorted(
-                    zip(data, colours, health),
-                    key=lambda t: health_order[t[2]],
-                    reverse=True,
+        if data and colours and health:
```
mvidalgarcia, line 436: Is there any case where any of these is `[]`?

mvidalgarcia (approved):
Works nicely 👍 

```console
$ flask reana-admin quota-usage 
ID                                     EMAIL              CPU-USED   CPU-LIMIT   DISK-USED   DISK-LIMIT
d38562b4-ac43-40f4-9049-d3796e5c6066   adelina@reana.io   9598       900000000   3647        4000      
00000000-0000-0000-0000-000000000000   marco@reana.io     62880      900000000   227891      -         
$ flask reana-admin quota-usage --human-readable
ID                                     EMAIL              CPU-USED     CPU-LIMIT   DISK-USED   DISK-LIMIT
d38562b4-ac43-40f4-9049-d3796e5c6066   adelina@reana.io   3.56 KiB     3.91 KiB    9s          250h      
00000000-0000-0000-0000-000000000000   marco@reana.io     222.55 KiB   -           1m 2s       250h   
```

mvidalgarcia:
⚠️  Oh, wait! disk and CPU values are inverted! @alintulu

PR #315, opened 2020-11-06
gitlab: fix fetching projects with other webhooks

mvidalgarcia:
closes reanahub/reana#323

##### TODO:
- [x] Add changelog

### DIFF
--- a/reana_server/rest/gitlab.py
+++ b/reana_server/rest/gitlab.py
@@ -16,7 +16,6 @@ from flask import (
     Blueprint,
     current_app,
     jsonify,
-    make_response,
     redirect,
     request,
     url_for,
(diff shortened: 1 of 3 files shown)


reana_server/rest/gitlab.py
```diff
@@ -324,7 +318,7 @@ def gitlab_webhook():  # noqa
                 REANA_GITLAB_URL + "/api/v4/projects/" + "{0}/hooks?access_token={1}"
             )
             webhook_payload = {
-                "url": "https://{}/api/workflows".format(REANA_HOSTNAME),
+                "url": "https://{}/api/workflows".format(request.host),
```
diegodelemos, line 321: This will work but it will expose a vulnerability too. `REANA_HOSTNAME` is totally under our control, however, if we use `request.host`, we are opening a door to host header attacks:
```python
$ ipython
In [1]: from flask import Flask
   ...: app = Flask(__name__)
   ...: 
   ...: @app.route('/')
   ...: def hello_world():
   ...:     return f'Hello from {request.host}'
In [2]: with app.test_client() as c:
   ...:     res = c.get()
   ...:     print(res.data)
b'Hello from localhost'
In [3]: with app.test_client() as c:
   ...:     res = c.get(headers={'Host': 'token-crusher.com'})
   ...:     print(res.data)
b'Hello from token-crusher.com'
```

Perhaps we can think of using this only if we are in debug mode, so developers can still use custom ports for local testing, what do you think @mvidalgarcia ?

mvidalgarcia: Reverted.

diegodelemos (approved):

PR #323, opened 2020-11-23
config: increment default rate limiting

diegodelemos:
* Addresses reanahub/reana#286.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changes
 =======
 
+Version 0.7.2 (UNRELEASED)
+--------------------------
+
+- Changes rate limiting defaults to allow up to 20 connections per second.
+
 Version 0.7.1 (2020-11-10)
 --------------------------
 
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -116,6 +116,11 @@ PROXYFIX_CONFIG = {"x_proto": 1}
 APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {}
 APP_HEALTH_BLUEPRINT_ENABLED = False
 
+# Rate limiting configuration
+# ===========================
+RATELIMIT_AUTHENTICATED_USER = "20 per second"
+RATELIMIT_GUEST_USER = "20 per second"
+
 # Flask-Breadcrumbs needs this variable set
 # =========================================
 BREADCRUMBS_ROOT = "breadcrumbs"


reana_server/config.py
```diff
@@ -116,6 +116,10 @@ def _(x):
 APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {}
 APP_HEALTH_BLUEPRINT_ENABLED = False
 
+# Rate limiting configuration
+# ===========================
+RATELIMIT_ENABLED = False
```
diegodelemos, line 121: Currently hardcoding this value.. but actually, it would be interesting to make it configurable in case someone wants to adjust it. Some thoughts about it:

Limiting with [Invenio-App/Flask-limiter](https://invenio-app.readthedocs.io/en/latest/configuration.html#invenio_app.config.RATELIMIT_APPLICATION):
- _Flexibility_: Configuring means touching many different variables (limiting per endpoint, limiting auth VS anonymous users) which allow fine-tuning.
- _To be adapted to Kubernetes_: [By default the limiter relies on _in memory_ entries to store for the number of requests](https://invenio-app.readthedocs.io/en/latest/configuration.html#invenio_app.config.RATELIMIT_STORAGE_URL) (this would become a problem if RS is scaled up). The solution is easy and it would be to connect the rate limiter to Redis (to be tested).

Limiting with [Traefik rate limiter](https://doc.traefik.io/traefik/middlewares/ratelimit/):
- _Less-flexibility_: We would have less granularity as doing it inside the application.
- _Closer to prod setup_: It would be more similar to what we do in prod with HAProxy.
- _More harmony_: the approach of rate-limiting inside the app wouldn't work for Interactive sessions (and future entities like them), Traefik would.

reana_server/config.py
```diff
@@ -116,6 +116,11 @@ def _(x):
 APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {}
 APP_HEALTH_BLUEPRINT_ENABLED = False
 
+# Rate limiting configuration
+# ===========================
+RATELIMIT_AUTHENTICATED_USER = "20 per second"
```
tiborsimko, line 121: Please amend `CHANGES.rst` and say something like:

- Changes rate limiting defaults to allow up to 20 connections per second.

CHANGES.rst
```diff
@@ -8,6 +8,8 @@ Version 0.7.1 (2020-11-10)
 - Fixes restarting of Yadage and CWL workflows.
 - Fixes conflicting ``kombu`` installation requirements by requiring Celery version 4.
 - Changes ``/api/you`` endpoint to include REANA server version information.
+- Changes rate limiting defaults to allow up to 20 connections per second.
```
tiborsimko, line 11: Must go into a new section:

```
Version 0.7.2 (UNRELEASED)
--------------------------

- Changes rate limiting defaults to allow up to 20 connections per second.
```
```

**Review by tiborsimko — APPROVED** [review]
Tested locally. The situations there were rate limited before:

```
File /LICENSE was successfully uploaded.
File /LICENSE was successfully uploaded.
File /LICENSE was successfully uploaded.
Something went wrong while uploading .../LICENSE:
60 per 1 minute
Something went wrong while uploading .../LICENSE:
60 per 1 minute
Something went wrong while uploading .../LICENSE:
60 per 1 minute
File /LICENSE was successfully uploaded.
File /LICENSE was successfully uploaded.
File /LICENSE was successfully uploaded.
```

are now passing fine :+1:

PR #327, opened 2020-11-24
release: 0.8.0a1

mvidalgarcia:


requirements.txt
```diff
@@ -4,39 +4,41 @@
 #
 #    pip-compile --output-file=requirements.txt setup.py
 #
-alembic==1.4.2            # via flask-alembic, reana-db
+alembic==1.4.3            # via flask-alembic, reana-db
 amqp==2.6.1               # via kombu
 appdirs==1.4.4            # via fs
-attrs==19.3.0             # via jsonschema
-babel==2.8.0              # via flask-babelex
+appnope==0.1.0            # via ipython
```
tiborsimko, line 10: Hmm, on my machine I do not get `appnope` dependency when I retry locally?

mvidalgarcia: Wow it looks that it's something related to macOS.. it ends up there when I run `pip-compile --output-file=requirements.txt setup.py`. I removed it but we should be careful. cc/ macOS users @diegodelemos @ParthS007, can you reproduce it?

tiborsimko: Indeed, `appnope` is "Disable App Nap on OS X 10.9". I wonder why we even need it inside `r-server` :wink: Seems to be coming via `invenio-app`. (Well, even `ipython` should not be really necessary to have in production images... so that they are slimmer.)

tiborsimko (approved):

PR #329, opened 2020-11-26
gitlab: add missing `signin_required` decorator

mvidalgarcia:


diegodelemos (approved):

diegodelemos:
@mvidalgarcia  Have you searched to see if we are missing the same decorator somewhere else?
E.g.
```console
reana_server/rest/gitlab.py:@blueprint.route("/gitlab/connect")
reana_server/rest/gitlab.py-def gitlab_connect():
reana_server/rest/gitlab.py-    r"""Endpoint to init the REANA connection to GitLab.
```

mvidalgarcia:
> @mvidalgarcia Have you searched to see if we are missing the same decorator somewhere else?
> E.g.
> 
> ```
> reana_server/rest/gitlab.py:@blueprint.route("/gitlab/connect")
> reana_server/rest/gitlab.py-def gitlab_connect():
> reana_server/rest/gitlab.py-    r"""Endpoint to init the REANA connection to GitLab.
> ```

That endpoint never had checks to verify if the user is authenticated, that's why is lacking the decorator. But actually, I think it makes sense to protect it even though we don't use the authenticated user object. What do you think?

mvidalgarcia:
Other endpoints that are not protected:

- `/api/config` - I think we need it for the UI so we can display SSO/local users forms before authentication.
- `/api/ping` - for obvious reasons

diegodelemos:
We can protect all the mentioned ones except `/api/ping` which we use for health checks.

mvidalgarcia:
I think we can't protect `/config` because otherwise, we don't know whether we should display SSO login, local users login or both. In any case, we're not exposing anything critical there.

PR #331, opened 2020-12-17
api: control ui signup form visibility with through config endpoint

audrium:
addresses reanahub/reana#469

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -33,11 +33,13 @@
             "description": "Configuration information to consume by Reana-UI.",
             "examples": {
               "application/json": {
+                "admin_email": "admin@example.org",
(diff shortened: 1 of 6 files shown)


docs/openapi.json
```diff
@@ -33,11 +33,13 @@
             "description": "Configuration information to consume by Reana-UI.",
             "examples": {
               "application/json": {
+                "admin_email": "team@reanahub.io",
```
mvidalgarcia, line 36: This email will be fully customizable by any institute having its REANA cluster, thus, what about using a the generic domain `example.org`? Maybe something like `admin@example.org`?

diegodelemos (approved):

PR #332, opened 2020-12-18
auth: introduce email confirmation after sign up

mvidalgarcia:
closes reanahub/reana#472
 
TODO:
- [x] Send email to new user with the confirmation link (pointing to reana-ui new `/confirm` route)
- [x] Check SSO users workflow (⚠️ https://github.com/reanahub/reana-server/pull/332/commits/6911b3f09318fe0598a5a541d5ca7fd6674d4bd3)


mvidalgarcia:
Since SSO accounts also need confirmation, after signing up with one of those, we got redirected to the default Invenio-Accounts login page that doesn't exist (404), as it's a UI view (Invenio) that's not whitelisted.

In addition, as we're using a UI view to trigger the SSO auth (check [this awfullness](https://github.com/reanahub/reana-ui/pull/170/commits/5cb07be01f57795aed6cf7ce24e26300a675563d#diff-37ec3004aa1b881eb0f0a93bf4c904a5295f638c81ecc26fbc553c70c92d244cR30)), we basically make full redirections, and reana-ui loses all context on what's happening. We can't know if the SSO sing in/up went successfully to display proper notifications from reana-ui.

This workflow is far from ideal. We should upgrade Invenio-Accounts to the latest version that supports REST apps, and revamp the whole workflow. As we also need to change to the new CERN SSO soon (supported in [latest Invenio-Accounts  version](https://github.com/inveniosoftware/invenio-oauthclient/blame/master/CHANGES.rst#L14)), that can be a good opportunity to tackle this problem.

Shall we open an issue with all this info?

reana_server/templates/emails/email_confirmation.txt
```diff
@@ -0,0 +1,9 @@
+Dear {{ user_full_name }},
+
+Your {{ reana_hostname }} account has been created successfully. Please confirm your email address by accessing this link <https://{{ reana_hostname}}/confirm/{{ confirm_token }}>.
```
audrium, line 3: I think some users might copy the link with ">." symbols mixed with the token. One way to fix it would be to change it like this: "...by accessing this link: https... ". Also, maybe the link could be made clickable?

mvidalgarcia: We've got this same format in [other templates](https://github.com/reanahub/reana-server/blob/master/reana_server/templates/emails/token_granted.txt), that's why I kept it consistent. I guess eventually we can move to HTML templates and introduce proper links.

tiborsimko: The `<...>` format for links in plain text emails is a standard that most clients support out of the box. 

But I see it could create confusion for some users perhaps... An alternative would be to print URLs clearly on a new line separated from the main text (and without using brackets this time), such as:

```
Dear {{ user_full_name }}:

Your {{ reana_hostname }} account has been created successfully. 
Please confirm your email address by accessing the following link:

  https://{{ reana_hostname}}/confirm/{{ confirm_token }}

% include 'emails/assistance_links.txt' %}

Best regards,

REANA notification system
```

P.S. We can also pay attention to have the regular text wrapped into under 78 characters per line, as per RFC 2822.

P.S. ... and let's keep plain text emails over HTML emails please :wink:

audrium (approved):

mvidalgarcia:
The initial idea was enabling email confirmation **only** for local users, thus avoiding that SSO users have to go through this process. Unfortunately, this is not easy to achieve because when we receive [this signal](https://github.com/inveniosoftware/invenio-oauthclient/blob/ca0e40e6483d2f8fe8ccda670c56bc6423bb69b0/invenio_oauthclient/handlers.py#L287), the Invenio user is still not created (happens [here](https://github.com/inveniosoftware/invenio-oauthclient/blob/ca0e40e6483d2f8fe8ccda670c56bc6423bb69b0/invenio_oauthclient/handlers.py#L304)). So in [our handler](https://github.com/reanahub/reana-server/blob/7f7351cdb57963c31ac021efdb775215dd2d1521/reana_server/utils.py#L151) we're no able to tweak the Invenio user object and set `confirmed_at` to skip the confirmation process.

After the SSO authorization happens, an Invenio user is created and [our local user handler](https://github.com/reanahub/reana-server/blob/7f7351cdb57963c31ac021efdb775215dd2d1521/reana_server/utils.py#L158) is called. Here we would be able to tweak the Invenio user as is sent together with the signal, however, we cannot determine if this call comes from an SSO user or a local user.

So we've decided to make the user confirmation feature opt-in, via a `values.yaml` reana-server environment variable.

reana_server/config.py
```diff
@@ -70,8 +71,16 @@ def _(x):
 ACCOUNTS_USERINFO_HEADERS = True
 #: Disable password recovery by users.
 SECURITY_RECOVERABLE = False
-#: Disable user to confirm their email address.
-SECURITY_CONFIRMABLE = False
+REANA_USER_CONFIRMATION = strtobool(os.getenv("REANA_USER_CONFIRMATION", "true"))
```
audrium, line 74: What do you think about leaving some comment above this line? Something like: `#: User's email address confirmation` could be used? Because it could be thought that it's related with `SECURITY_RECOVERABLE`

audrium (approved):
LGTM :robot:

diegodelemos (approved):

mvidalgarcia:
> LGTM 🤖
> 
> Edit: Can you also remove the [upper boundary of `Sphinx`](https://github.com/reanahub/reana-server/blob/maint-0.7/setup.py#L28) dependency in the same go? There is already a new release (3.4.3) with a fix for bug which broke `autodoc` for Flask projects.

Done, thanks! 👍

PR #339, opened 2021-02-22
rest: improve start workflow endpoint to work with unspecified operational_options parameter

audrium:
addresses https://github.com/reanahub/docs.reana.io/issues/23

While testing [script](https://forum.reana.io/t/python-api-to-run-workflows-on-reana/19/2) to run workflows from `reana-client` Python API, I noticed that it crashed since `operational_options` was not specified. This is a quick fix to address this problem

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -872,7 +872,7 @@ def start_workflow(workflow_id_or_name):  # noqa
         parameters = request.json
         workflow = _get_workflow_with_uuid_or_name(workflow_id_or_name, str(user.id_))
         parameters["operational_options"] = validate_operational_options(
-            workflow.type_, parameters["operational_options"]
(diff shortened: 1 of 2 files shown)


tiborsimko (approved):

reana_server/rest/workflows.py
```diff
@@ -872,7 +872,7 @@ def start_workflow(workflow_id_or_name):  # noqa
         parameters = request.json
         workflow = _get_workflow_with_uuid_or_name(workflow_id_or_name, str(user.id_))
         parameters["operational_options"] = validate_operational_options(
-            workflow.type_, parameters["operational_options"]
+            workflow.type_, parameters.get("operational_options", {})
```
tiborsimko, line 875: Can you please also add an entry to `CHANGES.rst` about this fix?

tiborsimko (approved):

PR #341, opened 2021-03-02
scheduler: read running workflows count from DB

diegodelemos:
* Stops relaying on Kubernetes API calls to determine number of
  running workflows. This tecnique is not reliable anymore when
  ``REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES`` is
  set since jobs will be kept for debugging, therefore
  counting as "running" since they didn't complete their execution.

  Since querying for jobs is not possible another path was explored
  which was, listing the pods with label `reana_workflow_mode=batch`.
  However, this is not reliable either due to race condition:
    - Querying jobs worked because we were creating a job
      (atomic operation in etcd/DB) and then consulting the API
      (reading from etcd/DB).
    - Querying pods doesn't work because their creation is not
      happening at job creation time. The time between creating
      a job and pods to be spawned produces a race contition in
      which `WorkflowExecutionScheduler` will submit more workflows
      than allowed by `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS`.

  Closes reanahub/reana-commons#250.

* Moves the `reana_ready` implentation logic from REANA-Commons to
  REANA-Server to be able to access REANA-DB. Also `reana_ready` is
  not used anywhere else, nor there are plans to do so soon.

* Avoids overloading Kubernetes API and REANA-DB when REANA is not
  ready to run workflows.

## To test

- Follow the example in https://github.com/reanahub/reana-commons/issues/250. In step `5.` you should see that `workflow.3` starts.

### DIFF
--- a/reana_server/scheduler.py
+++ b/reana_server/scheduler.py
@@ -12,15 +12,68 @@ import json
 import logging
 
 from bravado.exception import HTTPBadGateway, HTTPNotFound
+from kubernetes.client.rest import ApiException
+from sqlalchemy.exc import SQLAlchemyError
+
+from reana_commons.config import REANA_MAX_CONCURRENT_BATCH_WORKFLOWS
 from reana_commons.consumer import BaseConsumer
-from reana_commons.tasks import reana_ready
-from reana_db.database import Session
-from reana_db.models import Workflow
+from reana_commons.k8s.api_client import current_k8s_corev1_api_client
+from reana_db.models import Workflow, WorkflowStatus
 
 from reana_server.api_client import (
     current_rwc_api_client,
     current_workflow_submission_publisher,
 )
+
+
+def check_predefined_conditions():
+    """Check Kubernetes predefined conditions for the nodes."""
+    try:
+        node_info = json.loads(
+            current_k8s_corev1_api_client.list_node(
+                _preload_content=False
+            ).data.decode()
+        )
+        for node in node_info["items"]:
+            # check based on the predefined conditions about the
+            # node status: MemoryPressure, OutOfDisk, KubeletReady
+            #              DiskPressure, PIDPressure,
+            for condition in node.get("status", {}).get("conditions", {}):
+                if not condition.get("status"):
+                    return False
+    except ApiException as e:
+        logging.error("Something went wrong while getting node information.")
+        logging.error(e)
+        return False
+    return True
+
+
+def check_running_reana_workflows_count():
+    """Check upper limit on running REANA batch workflows."""
+    try:
+        running_workflows = Workflow.query.filter_by(
+            status=WorkflowStatus.running
+        ).count()
+        if running_workflows >= REANA_MAX_CONCURRENT_BATCH_WORKFLOWS:
+            return False
+    except SQLAlchemyError as e:
+        logging.error(
+            "Something went wrong while querying for number of running workflows."
+        )
+        logging.error(e)
+        return False
+    return True
+
+
+def reana_ready():
+    """Check if REANA can start new workflows."""
+    for check_condition in [
+        check_predefined_conditions,
+        check_running_reana_workflows_count,
+    ]:
+        if not check_condition():
+            return False
+    return True
 
 
 class WorkflowExecutionScheduler(BaseConsumer):
--- a/reana_server/scheduler.py
+++ b/reana_server/scheduler.py
@@ -10,6 +10,7 @@
 
 import json
 import logging
+from time import sleep
 
 from bravado.exception import HTTPBadGateway, HTTPNotFound
 from kubernetes.client.rest import ApiException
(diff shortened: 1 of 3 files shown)


reana_server/scheduler.py
```diff
@@ -10,17 +10,72 @@
...
+                if not condition.get("status"):
+                    return False
+    except ApiException as e:
+        logging.error("Something went wrong while getting node information.")
+        logging.error(e)
+        return False
+    return True
+
+
+def check_running_reana_workflows_count():
```
diegodelemos, line 53: This function changes from what was defined in REANA-Commons, see https://github.com/reanahub/reana-commons/pull/251, querying DB instead of Kubernetes API.

reana_server/scheduler.py
```diff
@@ -10,17 +10,72 @@
...
+    except SQLAlchemyError as e:
+        logging.error(
+            "Something went wrong while querying for number of running workflows."
+        )
+        logging.error(e)
+        return False
+    return True
+
+
+def reana_ready():
```
diegodelemos, line 70: `reana_ready` is simplified compared to https://github.com/reanahub/reana-commons/pull/251, referencing the `check_*` functions from inside the Python module.

reana_server/scheduler.py
```diff
@@ -12,17 +12,70 @@
...
+                if not condition.get("status"):
+                    return False
+    except ApiException as e:
+        logging.error("Something went wrong while getting node information.")
+        logging.error(e)
+        return False
+    return True
+
+
+def check_running_reana_workflows_count():
```
mvidalgarcia, line 53: Minor: By reading the name of this function is not clear what it returns for me. Would something like `exceedes_max_reana_workflow_count` be more understandable?

mvidalgarcia (commented):
Works as expected. I didn't manage to reproduce the race conditions mentioned in the description but probably I'm lacking knowledge.

diegodelemos:
> Works as expected. I didn't manage to reproduce the race conditions mentioned in the description but probably I'm lacking knowledge.

That's good, the race conditions happened on an intermediate solution, which had to be discarded due to that very problem. However, because it was the preferred in terms of design, I felt it needed justification. Perhaps for next time we can keep such solutions as a separate commit if they are important so one can test and understand better, let me know what you think.

P.S. I can rephrase the description I get some feedback on it.

reana_server/scheduler.py
```diff
@@ -12,17 +12,70 @@
...
+                _preload_content=False
+            ).data.decode()
+        )
+        for node in node_info["items"]:
+            # check based on the predefined conditions about the
+            # node status: MemoryPressure, OutOfDisk, KubeletReady
+            #              DiskPressure, PIDPressure,
+            for condition in node.get("status", {}).get("conditions", {}):
+                if not condition.get("status"):
+                    return False
```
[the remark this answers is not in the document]
diegodelemos: Thanks for the observation. If things were working as they seem the solution is almost correct, however, this is actually broken because what we get back from the response is not a boolean, it is a string, yielding always `True` when calling `condition.get("status")`, see:
```
$ kubectl exec -ti reana-server-765b6b6577-qk5sr bash
> ipython
In [1]: from reana_commons.k8s.api_client import current_k8s_corev1_api_client                                                                                                                                                        
                                                                                                                                                                                                                                      
In [2]: import json                                                                                                                                                                                                                   
                                                                                                                   
In [3]: node_info = json.loads(                                                                                                                                                                                                       
   ...:             current_k8s_corev1_api_client.list_node(                                                       
   ...:                 _preload_content=False                                                                     
   ...:             ).data.decode()                                                                                
   ...:         )                 
...
In [8]: node_info["items"][0]["status"]["conditions"][0]                                                           
Out[8]:                                                                                                            
{'type': 'MemoryPressure',                                                                                         
 'status': 'False',                                                                                                
 'lastHeartbeatTime': '2021-03-09T15:51:01Z',                                                                      
 'lastTransitionTime': '2021-03-09T13:13:54Z',                                                                     
 'reason': 'KubeletHasSufficientMemory',                                                                           
 'message': 'kubelet has sufficient memory available'}       
```
The condition check is more complex than it seems since the [conditions](https://kubernetes.io/docs/concepts/architecture/nodes/#condition) can be of positive nature (e.g. `Ready`, were we would expect `.get("status"` to be `True`) or negative nature (e.g. `DiskPressure`, where we would expect `.get("status"` to be `True`).

I think we should tackle this in a different issue since the PR at hand tackles https://github.com/reanahub/reana-commons/issues/250.

reana_server/scheduler.py
diegodelemos: Since the logic is a bit convoluted, it would have to be `doesnt_exceed_max_reana_workflow_count` (amended). Once we tackle https://github.com/reanahub/reana-server/pull/341#discussion_r590541272 we can do some clean up of these condition checks.

mvidalgarcia (approved):

PR #350, opened 2021-04-21
cli: add pending workflows to status summary

audrium:
closes https://github.com/reanahub/reana-workflow-controller/issues/363

### DIFF
--- a/reana_server/status.py
+++ b/reana_server/status.py
@@ -231,13 +231,25 @@ class WorkflowsStatus(REANAStatus):
 
         return number
 
-    def stuck_workflows(self):
-        """Get the number of stuck workflows."""
+    def stuck_in_running_workflows(self):
+        """Get the number of stuck running workflows."""
         inactivity_threshold = datetime.now() - timedelta(hours=12)
         number = (
             Session.query(Workflow)
             .filter(Workflow.status == RunStatus.running)
             .filter(Workflow.run_started_at <= inactivity_threshold)
+            .filter(Workflow.updated <= inactivity_threshold)
+            .count()
+        )
+
+        return number
+
+    def stuck_in_pending_workflows(self):
+        """Get the number of stuck pending workflows."""
+        inactivity_threshold = datetime.now() - timedelta(minutes=20)
+        number = (
+            Session.query(Workflow)
+            .filter(Workflow.status == RunStatus.pending)
             .filter(Workflow.updated <= inactivity_threshold)
             .count()
         )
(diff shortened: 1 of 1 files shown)


reana_server/status.py
```diff
@@ -236,7 +236,7 @@ def stuck_workflows(self):
         inactivity_threshold = datetime.now() - timedelta(hours=12)
         number = (
             Session.query(Workflow)
-            .filter(Workflow.status == RunStatus.running)
+            .filter(Workflow.status.in_([RunStatus.running, RunStatus.pending]))
```
diegodelemos, line 239: What do you think of moving this status check to its own new entry, e.g. `unschedulable`, for more granularity? In general, it would be more alarming to see "stuck in pending"/unschedulable workflows that "stuck in running", the latter can be a false positive (just a very long workflow), whereas the former indicates a problem with the system not being able to schedule pods. Also, a quick question, what would happen with a `pending` workflow that won't have the `Workflow.run_started_at` set?

That new entry could go instead of `pending` and we could say something like e.g. if a workflow has been in `pending` for more than 20 minutes, it is definitely `unschedulable`.

reana_server/status.py
```diff
@@ -244,6 +244,18 @@ def stuck_workflows(self):
 
         return number
 
+    def stuck_workflows(self):
```
diegodelemos, line 247:
```suggestion
    def stuck_in_pending_workflows(self):
```

reana_server/status.py
```diff
@@ -231,8 +231,8 @@ def restarted_workflows(self):
 
         return number
 
-    def stuck_workflows(self):
-        """Get the number of stuck workflows."""
+    def unschedulable_workflows(self):
```
diegodelemos, line 234:
```suggestion
    def stuck_in_running_workflows(self):
```

reana_server/status.py
```diff
@@ -255,8 +267,10 @@ def get_status(self):
         return {
             "running": self.get_workflows_by_status(RunStatus.running),
             "finished": self.get_workflows_by_status(RunStatus.finished),
-            "stuck": self.stuck_workflows(),
+            "stuck in running": self.unschedulable_workflows(),
+            "stuck in pending": self.stuck_workflows(),
```
diegodelemos, line 271:
```suggestion
            "stuck in running": self.stuck_in_running_workflows(),
            "stuck in pending": self.stuck_in_pending_workflows(),
```

reana_server/status.py
```diff
@@ -255,8 +267,10 @@ def get_status(self):
...
             "queued": self.get_workflows_by_status(RunStatus.queued),
+            "pending": self.get_workflows_by_status(RunStatus.pending),
```
diegodelemos, line 273: Perhaps we don't need this? It could give a false impression if just in the morning when the status report is calculated a few workflows are in pending, even if few seconds after are scheduled.

diegodelemos (commented):
I've just left a suggestion to perhaps simplify naming since my initial suggestion might have been confusing.

diegodelemos (approved):

PR #362, opened 2021-05-06
system-status: enrich with node, pod & job info

mvidalgarcia:
closes reanahub/reana#501
closes reanahub/reana-ui#183

#### Install `metrics-server`
If `kubectl top nodes` doesn't work for you, it's because you need  `metrics-server` installed.

```console
$ kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml
$ kubectl edit -n kube-system deployment/metrics-server
# Add `--kubelet-insecure-tls` to `spec.template.spec.containers[0].args` and save.
```

⚠️ Tests not passing due to missing [`reana-commons` release](https://github.com/reanahub/reana-commons/pull/262).

#### To test:

```console
kubectl exec -ti deployment/reana-server -- flask reana-admin status-report --admin-access-token $REANA_ACCESS_TOKEN 
```

### DIFF
--- a/reana_server/status.py
+++ b/reana_server/status.py
@@ -8,13 +8,17 @@
 
 """Status module for REANA."""
 
+import enum
 import logging
 import subprocess
 from datetime import datetime, timedelta
 
 from invenio_accounts.models import SessionActivity
+from kubernetes.client.rest import ApiException
+from marshmallow import Schema, fields
 from reana_commons.config import (
     REANA_COMPONENT_PREFIX,
+    REANA_COMPUTE_BACKENDS,
     REANA_INFRASTRUCTURE_KUBERNETES_NAMESPACE,
     REANA_RUNTIME_KUBERNETES_NAMESPACE,
     SHARED_VOLUME_PATH,
@@ -390,16 +394,16 @@ class NodesStatus(REANAStatus):
 
                 node_capacity = result[node_name]["capacity"]
                 node_usage = result[node_name]["usage"]
-                node_usage_percentage = round(
-                    kubernetes_memory_to_bytes(node_usage)
-                    / kubernetes_memory_to_bytes(node_capacity)
-                    * 100
+                node_usage_percentage = ClusterHealth.get_percentage(
+                    kubernetes_memory_to_bytes(node_usage),
+                    kubernetes_memory_to_bytes(node_capacity),
                 )
                 result[node_name]["percentage"] = f"{node_usage_percentage}%"
         except ApiException as e:
-            logging.error("Error while calling `metrics.k8s.io` API.")
+            msg = "Error while calling `metrics.k8s.io` API."
+            logging.error(msg)
             logging.error(e)
-            return {}
+            return {"error": msg}
 
         return result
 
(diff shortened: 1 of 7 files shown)


reana_server/status.py
```diff
@@ -341,12 +361,285 @@ def get_status(self):
...
+            node_metrics = current_k8s_custom_objects_api_client.list_cluster_custom_object(
+                "metrics.k8s.io", "v1beta1", "nodes"
+            )
+            for node_metric in node_metrics.get("items", []):
+                node_name = node_metric["metadata"]["name"]
+                result[node_name]["usage"] = node_metric["usage"]["memory"]
+
+                node_capacity = result[node_name]["capacity"]
+                node_usage = result[node_name]["usage"]
+                node_usage_percentage = round(
```
audrium, line 396: Minor: probably `ClusterHealth.get_percentage()` could be utilized here? Also maybe `get_percentage()` could be moved to `utils.py`? I see we have something similar there `get_usage_percentage()`, but it returns a string instead of int..

reana_server/status.py
```diff
@@ -341,12 +361,285 @@ def get_status(self):
...
+        :param from_: From which moment in time to collect information. Not
+            implemented yet.
+        :param until: Until which moment in time to collect information. Not
+            implemented yet.
+        :param user: A REANA-DB user model.
+        :type from_: datetime
+        :type until: datetime
+        :type user: reana_db.models.User
+        """
+        self.compute_backends = ["Kubernetes", "HTCondor", "Slurm"]
```
audrium, line 481: Minor: I guess there should be some config to get these values and not to duplicate it?

reana_server/status.py
```diff
@@ -341,12 +361,285 @@ def get_status(self):
...
+            "health": ClusterHealth.get_health_status(percentage),
+            "sort": 1,
+        }
+
+    def get_workflow_health(self):
+        """Get cluster workflows health information."""
+        wf_status_obj = WorkflowsStatus()
+
+        running = wf_status_obj.get_workflows_by_status(RunStatus.running)
+        queued = wf_status_obj.get_workflows_by_status(RunStatus.queued)
```
audrium, line 606: Question: why we are ignoring `pending` workflows? Since workflows can get stuck in `pending` state for some time, it could be important while checking the cluster health

reana_server/status.py
mvidalgarcia: That's a very good point. Unfortunately, we save [this value in DB as a simple string](https://github.com/reanahub/reana-db/blob/01976eb2f5a9072303715d6ac6fce4ad6de3776f/reana_db/models.py#L682)... IMO it should be an enum so we can control the values assigned and in cases like this, get them all to perform queries.

I've dug a bit into our codebase and I only see [_ these values_ in reana-job-controller](https://github.com/reanahub/reana-job-controller/blob/f89bdf78b60305f271ccfe06d35ee393884eb67a/reana_job_controller/config.py#L20-L30), but we can't even use those because we can't import from this component and the names don't match exactly what we have in the database. An ideal solution IMO would be modifying this at a DB level, but it would require a schema change, alembic recipe, etc. What do you think?

I see that `r-client` also [has something similar](https://github.com/reanahub/reana-client/blob/ac31929726c4b8da7b9d8a885ed3e640b572ba18/reana_client/cli/workflow.py#L796-L800), I'll move this to `r-commons` and use it from `r-server` as a temporary workaround.

reana_server/status.py
mvidalgarcia: Makes sense! Added.

reana_server/status.py
mvidalgarcia: PR created: https://github.com/reanahub/reana-client/pull/517

audrium (approved):
works great! :racing_car:

Issue #358, opened 2021-05-06
scheduler: start workflow that has best chances of immediately running


tiborsimko:
Consider incoming workflow queue has two workflow runs W1 and W2 issued by the same user. 

Consider W1 spawns 20 n-tupling jobs each needing 8 GiB RAM, while W2 spawns 4 n-tupling jobs each needing 6 GiB RAM.

Consider the current cluster status is such that it has 36 GiB free memory for job slots.

In these conditions, it makes sense to launch W2 instead of W1, because W2 initial jobs may be immediately "running", whilst W1 jobs would be mostly in the "pending" state.

The goal of this issue is to walk through incoming workflow queue and decide upon which workflow to run based on the initial workflow job complexity.

This task depends on DAG analysis to discover initial workflow complexities, which will be ticketised separately as dependent issues.

Corner issues to consider:
- The above example expresses free cluster memory capacity as 36 GiB by one number, but in reality the free cluster memory capacity is defined by each node, e.g. it could be one node with more merory (1, 8GiB) or more nodes with fewer memory (4, 2GiB) slots. Hence, if incoming workflow W1 is (2, 2GiB) and W2 is (1, 3GiB), in the former case W2 should have priority, in the latter case W1 should have priority.
- Waiting times for workflows in the queue. E.g. W1 is so big and the cluster is so busy that W2 and even W9 and even W39 would be scheduled beforehand.  So the ultimate formula to pick the best workflow to start should be combination of how cluster is currently busy, what is currently possible, and how long a given workflow waits in the queue (or how many times it was skipped and "overtaken" by smaller workflows already).

Stems from #357.

tiborsimko:
Depends on #359.

audrium:
Since fetching all the queue messages from the consumer is a bit troublesome, another approach we could take is to use [RabbitMQ priority queues](https://www.rabbitmq.com/priority.html) - to set workflow priority depending on the complexity while publishing the message

mvidalgarcia:
📪

PR #366, opened 2021-05-20
scheduler: prioritize workflows in the queue by complexity

audrium:
closes https://github.com/reanahub/reana-server/issues/358

This PR tweaks start_workflow() endpoint to estimate and store workflow complexity in the database. The reason behind storing this value in the database is to make scheduling process a bit lighter. If the logic would be there, it would mean additional calls to db (to fetch workflow spec, etc.) and complexity calculation each time the workflow reaches consumer (it can be rescheduled multiple times).

Complexity estimation code is taken from this [PR](https://github.com/reanahub/reana-client/pull/516) with the only changes being made [here](https://github.com/reanahub/reana-server/pull/366/files#diff-dd451815d7362d3092b6e4d2bc6880a2fe0698d82d2415e90b8adc23181eb3faR20-R35) to simply calculate memory needed for all initial jobs and to determine the workflow complexity by it (needs to be discussed and improved).

In addition it also loads Yadage specification and stores it in the database in the `start_workflow()` endpoint

### A bit of context about the PR set in general

By default `RabbitMQ just dispatches a message when the message enters the queue.`, but there is an option to tell RabbitMQ not to give more than one message to a consumer at a time. This option can be set by defining `prefetch_count=1` in the `kombu Consumer`. You can read more about it in the official docs - [Fair dispatch section](https://www.rabbitmq.com/tutorials/tutorial-two-python.html). 

With this option on all the unconsumed messages in the RabbitMQ queue are still in the `Ready` state (which means they are still in the queue ready to be consumed) instead of `Unacked` (which means they already reached the consumer, but was not yet processed.)

#### Before testing it

Make sure you redeploy the cluster after checking out the code or after making any changes to the queues. RabbitMQ does not allow to modify the queues while it is still running.

#### How to test it

- First of all here is the procedure to launch RabbitMQ management UI:
  ```
  kubectl exec reana-message-broker-5874b75cdd-xzm5c rabbitmq-plugins enable rabbitmq_management
  kubectl port-forward reana-message-broker-5874b75cdd-xzm5c 15672:15672
  ```
  Open `http://localhost:15672` in the browser

- Add `import wdb; wdb.set_trace()` breakpoint [here](https://github.com/reanahub/reana-server/blob/master/reana_server/scheduler.py#L139) before `message.ack()`.
- Restart `reana-server` pod. Note that live reloading does not work in this pod, so after any changes the pod needs to be killed.
- Start a new workflow from any demo example we have
- Open up Wdb web interface: http://localhost:31984/
- Inspect `workflow_submission`, it should also have `priority` inside
- Check RabbitMQ management console, `workflow-submission` should now have 1 `Unacked` message, it means it was received by the consumer
- Start `bsm-search` workflow
- Start `helloworld` workflow
- RabbitMQ management console should now show 3 messages in total, 2 should be in ready state
- Force close (stop tracing) the Wdb debugger session
- Inspect `workflow_submission` argument of the next debugger session, it should be the `helloworld` workflow with priority 3. Even though we submitted `bsm-search` workflow first, the next processed workflow was `helloworld`!
- Force close the debugger session, the next processed workflow should be `bsm-search`

#### TODO
- [ ] Count how many times workflow was requeued to avoid loops, e.g. if cluster has only 1 GiB free, the consumer action could loop forever (separate issue?)
- [ ] Decide on default memory limit for job while estimating the complexity since `REANA_KUBERNETES_JOBS_MEMORY_LIMIT` could be not set through helm
- [ ] Modify `create_workflow()` endpoint to call `start_endpoint()` for gitlab workflows, since we calculate the complexity and set priority only there

#### Depends on:
- https://github.com/reanahub/pytest-reana/pull/83
- https://github.com/reanahub/reana-commons/pull/264

### DIFF
--- a/reana_server/complexity.py
+++ b/reana_server/complexity.py
@@ -11,8 +11,33 @@
 import os
 from functools import reduce
 
-from reana_db.models import WorkflowComplexity
 from reana_commons.job_utils import kubernetes_memory_to_bytes
+from reana_commons.config import MQ_MAX_PRIORITY
+
+from reana_server.status import NodesStatus
+from reana_server.config import REANA_COMPLEXITY_JOBS_MEMORY_LIMIT
+
+
+def get_workflow_min_job_memory(complexity):
+    """Return minimal job memory from workflow complexity.
+
+    :param complexity: workflow complexity list which consists of number of initial jobs and the memory in bytes they require. (e.g. [(8, 1073741824), (5, 2147483648)])
+    :return: minimal job memory (e.g. 1073741824)
+    """
+    if not complexity:
+        return 0
+    return min(complexity, key=lambda x: x[1])[1]
+
+
+def calculate_workflow_priority(complexity):
+    """Calculate workflow priority based on its complexity."""
+    if not complexity:
+        return 0
+    wf_memory = int(reduce(lambda sum, item: sum + item[0] * item[1], complexity, 0))
+    total_cluster_memory = NodesStatus().get_total_memory()
+    if not total_cluster_memory or wf_memory > total_cluster_memory:
+        return 0
+    return int(round(1 - wf_memory / total_cluster_memory, 2) * MQ_MAX_PRIORITY)
 
 
 def estimate_complexity(workflow_type, reana_yaml):
@@ -38,17 +63,8 @@ def estimate_complexity(workflow_type, reana_yaml):
     try:
         complexity = estimator.estimate_complexity()
     except Exception:
-        return None
-
-    # Summing up parrarel jobs times memory to determine the workflow complexity
-    memory = reduce(lambda sum, item: sum + item[0] * item[1], complexity, 0)
-
-    # FIXME: Values should be configuranble
-    if memory <= kubernetes_memory_to_bytes("8Gi"):
-        return WorkflowComplexity.espresso
-    if memory <= kubernetes_memory_to_bytes("40Gi"):
-        return WorkflowComplexity.medium
-    return WorkflowComplexity.hard
+        return []
+    return complexity
 
 
 class ComplexityEstimatorBase:
@@ -89,8 +105,9 @@ class ComplexityEstimatorBase:
 
     def _get_memory_limit(self, step):
         """Get memory limit value."""
-        default_memory_limit = os.getenv("REANA_KUBERNETES_JOBS_MEMORY_LIMIT", "8Gi")
-        memory_limit = step.get("kubernetes_memory_limit", default_memory_limit)
+        memory_limit = step.get(
+            "kubernetes_memory_limit", REANA_COMPLEXITY_JOBS_MEMORY_LIMIT
+        )
         return kubernetes_memory_to_bytes(memory_limit)
 
 
--- a/reana_server/complexity.py
+++ b/reana_server/complexity.py
@@ -1,0 +1,343 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2021 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA workflow complexity estimation."""
+
+import os
+from functools import reduce
+
+from reana_db.models import WorkflowComplexity
+from reana_commons.job_utils import kubernetes_memory_to_bytes
+
+
+def estimate_complexity(workflow_type, reana_yaml):
+    """Estimate complexity in REANA workflow.
+
+    :param workflow_type: A supported workflow specification type.
+    :param reana_yaml: REANA YAML specification.
+    """
+
+    def build_estimator(workflow_type, reana_yaml):
+        if workflow_type == "serial":
+            return SerialComplexityEstimator(reana_yaml)
+        elif workflow_type == "yadage":
+            return YadageComplexityEstimator(reana_yaml)
+        elif workflow_type == "cwl":
+            return CWLComplexityEstimator(reana_yaml)
+        else:
+            raise Exception(
+                "Workflow type '{0}' is not supported".format(workflow_type)
+            )
+
+    estimator = build_estimator(workflow_type, reana_yaml)
+    try:
+        complexity = estimator.estimate_complexity()
+    except Exception:
+        return None
+
+    # Summing up parrarel jobs times memory to determine the workflow complexity
+    memory = reduce(lambda sum, item: sum + item[0] * item[1], complexity, 0)
+
+    # FIXME: Values should be configuranble
+    if memory <= kubernetes_memory_to_bytes("8Gi"):
+        return WorkflowComplexity.espresso
+    if memory <= kubernetes_memory_to_bytes("40Gi"):
+        return WorkflowComplexity.medium
+    return WorkflowComplexity.hard
+
+
+class ComplexityEstimatorBase:
+    """REANA workflow complexity estimator base class."""
+
+    def __init__(self, reana_yaml):
+        """Estimate complexity in REANA workflow.
+
+        :param reana_yaml: REANA YAML specification.
+        :param initial_step: initial workflow execution step.
+        """
+        self.reana_yaml = reana_yaml
+        self.specification = reana_yaml.get("workflow", {}).get("specification", {})
+        self.input_params = reana_yaml.get("inputs", {}).get("parameters", {})
+
+    def parse_specification(self, initial_step):
+        """Parse REANA workflow specification tree."""
+        raise NotImplementedError
+
+    def estimate_complexity(self, initial_step="init"):
+        """Estimate complexity in parsed REANA workflow tree."""
+        steps = self.parse_specification(initial_step)
+        return self._calculate_complexity(steps)
+
+    def _calculate_complexity(self, steps):
+        """Calculate complexity in parsed REANA workflow tree."""
+        complexity = []
+        for step in steps.values():
+            complexity += step["complexity"]
+        return complexity
+
+    def _get_number_of_jobs(self, step):
+        """Get number of jobs based on compute backend."""
+        backend = step.get("compute_backend")
+        if backend and backend != "kubernetes":
+            return 0
+        return 1
+
+    def _get_memory_limit(self, step):
+        """Get memory limit value."""
+        default_memory_limit = os.getenv("REANA_KUBERNETES_JOBS_MEMORY_LIMIT", "8Gi")
+        memory_limit = step.get("kubernetes_memory_limit", default_memory_limit)
+        return kubernetes_memory_to_bytes(memory_limit)
+
+
+class SerialComplexityEstimator(ComplexityEstimatorBase):
+    """REANA serial workflow complexity estimation."""
+
+    def _parse_steps(self, steps):
+        """Parse serial workflow specification tree."""
+        tree = []
+        for idx, step in enumerate(steps):
+            name = step.get("name", str(idx))
+            jobs = self._get_number_of_jobs(step)
+            memory_limit = self._get_memory_limit(step)
+            complexity = [(jobs, memory_limit)]
+            tree.append({name: {"complexity": complexity}})
+        return tree
+
+    def parse_specification(self, initial_step):
+        """Parse and filter out serial workflow specification tree."""
+        spec_steps = self.specification.get("steps", [])
+        steps = self._parse_steps(spec_steps)
+        if initial_step == "init":
+            return steps[0] if steps else {}
+        return next(filter(lambda step: initial_step in step.keys(), steps), {})
+
+
+class YadageComplexityEstimator(ComplexityEstimatorBase):
+    """REANA Yadage workflow complexity estimation."""
+
+    def _parse_steps(self, stages, initial_step):
+        """Parse and filter out Yadage workflow tree."""
+
+        def _is_initial_stage(stage):
+            dependencies = stage.get("dependencies", {}).get("expressions", [])
+
+            if dependencies == [initial_step]:
+                return True
+
+            if initial_step == "init":
+                # Not defined dependencies should be treated as `init`
+                return not dependencies
+            return False
+
+        def _get_stage_complexity(stage):
+            resources = (
+                stage.get("scheduler", {})
+                .get("step", {})
+                .get("environment", {})
+                .get("resources", [])
+            )
+            compute_backend = next(
+                filter(lambda r: "compute_backend" in r.keys(), resources), {},
+            )
+            k8s_memory_limit = next(
+                filter(lambda r: "kubernetes_memory_limit" in r.keys(), resources), {},
+            )
+            jobs = self._get_number_of_jobs(compute_backend)
+            memory_limit = self._get_memory_limit(k8s_memory_limit)
+            return [(jobs, memory_limit)]
+
+        def _parse_stages(stages):
+            tree = {}
+            for stage in stages:
+                if not _is_initial_stage(stage):
+                    continue
+                name = stage["name"]
+                scheduler = stage.get("scheduler", {})
+                parameters = scheduler.get("parameters", [])
+                tree[name] = {"params": parameters, "stages": {}, "scatter_params": []}
+
+                # Parse stage complexity
+                tree[name]["complexity"] = _get_stage_complexity(stage)
+
+                # Parse nested stages
+                if "workflow" in scheduler:
+                    nested_stages = scheduler["workflow"].get("stages", [])
+                    parsed_stages = _parse_stages(nested_stages)
+                    tree[name]["stages"].update(parsed_stages)
+
+                # Parse scatter parameters
+                if "scatter" in scheduler and scheduler["scatter"]["method"] == "zip":
+                    tree[name]["scatter_params"] = scheduler["scatter"]["parameters"]
+
+            return tree
+
+        return _parse_stages(stages)
+
+    def _populate_parameters(self, stages, parent_params):
+        """Populate parsed Yadage workflow tree with parameter values."""
+
+        def _parse_params(stage, parent_params):
+            parent_params = parent_params.copy()
+            for param in stage["params"]:
+                if isinstance(param["value"], list):
+                    parent_params[param["key"]] = param["value"]
+                elif isinstance(param["value"], dict):
+                    # Example: input_file: {step: init, output: files}
+                    # In this case `files` values should be taken from
+                    # `parent_params` and saved as `input_file`
+                    output = param["value"].get("output", "")
+                    parent_value = parent_params.get(output, "")
+                    parent_params[param["key"]] = parent_value
+                else:
+                    parent_params[param["key"]] = [param["value"]]
+            return parent_params
+
+        def _parse_stages(stages, parent_params):
+            stages = stages.copy()
+            for stage in stages.keys():
+                stage_value = stages[stage]
+                # Handle params
+                params = _parse_params(stage_value, parent_params)
+                stage_value["params"] = params
+                # Handle nested stages
+                stage_value["stages"] = _parse_stages(stage_value["stages"], params)
+            return stages
+
+        return _parse_stages(stages, parent_params)
+
+    def _populate_complexity(self, stages):
+        """Calculate number of jobs and memory needed for the parsed Yadage workflow tree."""
+
+        def _parse_stages(stages):
+            stages = stages.copy()
+            for stage in stages.keys():
+                stage_value = stages[stage]
+                complexity = stage_value["complexity"]
+
+                # Handle nested stages
+                parsed_stages = _parse_stages(stage_value["stages"])
+                stage_value["stages"] = parsed_stages
+                if parsed_stages:
+                    complexity = self._calculate_complexity(parsed_stages)
+
+                # Handle scatter parameters
+                if stage_value["scatter_params"]:
+                    first_param = stage_value["scatter_params"][0]
+                    param_len = len(stage_value["params"].get(first_param, []))
+                    complexity = [(item[0] * param_len, item[1]) for item in complexity]
+
+                stage_value["complexity"] = complexity
+            return stages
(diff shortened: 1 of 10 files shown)


reana_server/complexity.py
```diff
@@ -0,0 +1,343 @@
...
+    estimator = build_estimator(workflow_type, reana_yaml)
+    try:
+        complexity = estimator.estimate_complexity()
+    except Exception:
+        return None
+
+    # Summing up parrarel jobs times memory to determine the workflow complexity
+    memory = reduce(lambda sum, item: sum + item[0] * item[1], complexity, 0)
+
+    # FIXME: Values should be configuranble
```
mvidalgarcia, line 46: You can check how I'm doing it here to adapt later.

- https://github.com/reanahub/reana/pull/507/files#diff-6846786db1d8b0e9a4b1ff80931cb967630b00a58374384917c9bf356dabdfacR74
- https://github.com/reanahub/reana-db/pull/131/files#diff-9c9ecb5b02eee5df82a9c1138c56e62181e9d8d56152d45c09f7fe0a693ce980R262

mvidalgarcia:
We should change also the Dockerfile to add `reana-commons[kubernetes,yadage]` for debug mode.

https://github.com/reanahub/reana-server/blob/2cb9e61e795de1b99a3e4c8ddd6e2f7e5c93130e/Dockerfile#L31

audrium:
> We should change also the Dockerfile to add reana-commons[kubernetes,yadage] for debug mode.
This took a while for me to figure out why it did not work after fresh `run-ci` :D

reana_server/status.py
```diff
@@ -377,6 +378,21 @@ def get_unschedulable_nodes(self):
...
+    def get_nodes_memory(self):
+        """Get list of all node memory capacities."""
+        try:
+            nodes = current_k8s_corev1_api_client.list_node()
+            return [
+                kubernetes_memory_to_bytes(node.status.capacity["memory"])
+                for node in nodes.items
+            ]
+        except ApiException:
+            return [kubernetes_memory_to_bytes("30Gi")]
```
audrium, line 390: Suggestions are welcomed how to improve this part. Since we need some value for total cluster memory capacity, I simply return this dummy value on error. If we don't return any value here workflow priorities could not be set.

reana_server/status.py
```diff
@@ -407,6 +424,13 @@ def get_memory_usage(self):
 
         return result
 
+    def get_available_memory(self):
+        """Get list of available nodes memory."""
+        nodes = self.get_memory_usage()
+        if not nodes or "error" in nodes:
+            return self.get_nodes_memory()
```
audrium, line 431: In case of any errors to get available nodes memory, total memory from all nodes is returned simply not to stop any workflows from passing into the cluster, since we don't know if it's free or not..

reana_server/complexity.py
```diff
@@ -0,0 +1,354 @@
...
+    def _get_number_of_jobs(self, step):
+        """Get number of jobs based on compute backend."""
+        backend = step.get("compute_backend")
+        if backend and backend != "kubernetes":
+            return 0
+        return 1
+
+    def _get_memory_limit(self, step):
+        """Get memory limit value."""
+        default_memory_limit = os.getenv("REANA_KUBERNETES_JOBS_MEMORY_LIMIT", "8Gi")
```
mvidalgarcia, line 103: I got some problems due to this default limit. My local cluster has 6Gi, and any workflow I run without `kubernetes_memory_limit` keeps being queued forever as `check_memory_availability` is `False`.

This means that any cluster with less than `8Gi` won't be able to schedule workflows with the default configuration. I think this can be tricky for small instances, or multi-node instances with _small_ (<8Gi) nodes.

In addition, I'd move this default limit to the `config.py` and import `REANA_KUBERNETES_JOBS_MEMORY_LIMIT` from `config.py` as well, as r-w-c does. Alternatively, as this is being used in both r-server and r-w-e, we can put it in reana-commons and import it from there.

reana_server/rest/workflows.py
```diff
@@ -414,6 +417,8 @@ def create_workflow(user):  # noqa
                 user_id=str(user.id_),
                 workflow_id_or_name=response["workflow_id"],
                 parameters=request.json,
+                # TODO: send complexity
+                # TODO: send min job memory
```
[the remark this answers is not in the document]
mvidalgarcia: Could you create a separate issue and add it to the kanban board?

reana_server/complexity.py
audrium: Great that you spotted this one! As a fix I moved `REANA_KUBERNETES_JOBS_MEMORY_LIMIT` limit to `config.py` in r-server. Since it have a default value, I didn't want to put it in r-commons.

Also, reduced the default value from `8Gi` to `4Gi` to solve `check_memory_availability` issue. It's not ideal, but should work in most of the cases

reana_server/scheduler.py
```diff
@@ -27,6 +27,16 @@
...
+from reana_server.status import NodesStatus
+
+
+def check_memory_availability(workflow_min_job_memory):
+    """Check if at least one workflow job could be started in Kubernetes."""
+    nodes = NodesStatus().get_available_memory()
+    max_node_available_memory = max(nodes)
+    if max_node_available_memory >= workflow_min_job_memory:
+        return True
+    return False
```
mvidalgarcia, line 39:
```suggestion
    return max_node_available_memory >= workflow_min_job_memory
```

reana_server/scheduler.py
```diff
@@ -138,9 +153,12 @@ def on_message(self, workflow_submission, message):
         """On new workflow_submission event handler."""
         message.ack()
         workflow_submission = json.loads(workflow_submission)
-        if reana_ready():
+        workflow_min_job_memory = workflow_submission.pop("min_job_memory", 0)
+        reana_memory_available = check_memory_availability(workflow_min_job_memory)
+        if reana_ready() and reana_memory_available:
```
mvidalgarcia, line 158: Is it OK to move `reana_memory_available`/`check_memory_availability` to `reana_ready()`?

reana_server/status.py
```diff
@@ -377,14 +378,31 @@ def get_unschedulable_nodes(self):
...
+
     def get_memory_usage(self):
         """Get nodes memory usage."""
         result = dict()
-        nodes = current_k8s_corev1_api_client.list_node()
-        for node in nodes.items:
-            result[node.metadata.name] = {"capacity": node.status.capacity["memory"]}
-
         try:
+            nodes = current_k8s_corev1_api_client.list_node()
```
mvidalgarcia, line 400: If this function fails here, the message `"Error while calling metrics.k8s.io API."` is not correct as `list_node` is not part of the metrics API.

reana_server/rest/workflows.py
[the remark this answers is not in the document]
audrium: New issue can be found [here](https://github.com/reanahub/reana-server/issues/368)

reana_server/status.py
```diff
@@ -12,6 +12,7 @@
 import logging
 import subprocess
 from datetime import datetime, timedelta
+from functools import reduce
```
mvidalgarcia, line 15: Unused.

reana_server/complexity.py
```diff
@@ -0,0 +1,356 @@
...
+def calculate_workflow_priority(complexity):
+    """Calculate workflow priority based on its complexity."""
+
+    def get_bin(memory):
+        for b in range(bins):
+            if memory in range(bin_size * b, bin_size * (b + 1)):
+                return b
+        return bins - 1
+
+    # Summing up parrarel jobs times memory to determine the workflow complexity
```
mvidalgarcia, line 30:
```suggestion
    # Summing up parallel jobs times memory to determine the workflow complexity
```

reana_server/complexity.py
```diff
@@ -0,0 +1,356 @@
...
+    memory = int(reduce(lambda sum, item: sum + item[0] * item[1], complexity, 0))
+    bins = MQ_MAX_PRIORITY
+    total_cluster_memory = NodesStatus().get_total_memory()
+    bin_size = int(total_cluster_memory / bins)
+    priority = bins - get_bin(memory)
+    return priority
```
mvidalgarcia, line 36: Isn't it enough to do something like this?
```python
int((1 - wf_memory / total_cluster_memory) * MQ_MAX_PRIORITY)
```

Maybe I'm missing something.

reana_server/status.py
mvidalgarcia: Not sure what's the best solution.. It's a pity we cannot rely on k8s API... cc/ @tiborsimko

tests/test_scheduler.py
```diff
@@ -34,6 +34,7 @@ def test_scheduler_starts_workflows(
     with patch.multiple(
         "reana_server.scheduler",
         reana_ready=Mock(return_value=True),
+        check_memory_availability=Mock(return_value=True),
```
mvidalgarcia, line 37: Would it be feasible to add some tests for the new functionality? Also for the complexity estimator would be handy.

reana_server/rest/workflows.py
```diff
@@ -918,12 +924,30 @@ def start_workflow(workflow_id_or_name, user):  # noqa
...
+            # Load Yadage workflow specification
+            operational_options.update({"accept_metadir": True})
+            workflow_file = workflow.reana_specification["workflow"].get("file")
+            toplevel = operational_options.get("toplevel", "")
+            workflow_spec = yadage_load_from_workspace(
+                workflow.workspace_path, workflow_file, toplevel,
+            )
+            reana_specification = copy.deepcopy(workflow.reana_specification)
+            reana_specification["workflow"]["specification"] = workflow_spec
+            workflow.reana_specification = reana_specification
```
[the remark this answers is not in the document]
audrium: I've moved the `reana_specification` copying part into r-commons. Let me know if you have some ideas how we can simplify this even more

reana_server/scheduler.py
audrium: I haven't moved it there since it requires an argument

reana_server/status.py
audrium: You are correct, it fails with different exception than `ApiException`. However `list_node()` is the function which produces the famous `Invalid value for names, must not be None` error. I've updated the code to catch it. Also checked `kubernetes-client` issues and there is already a fix for that, but it is not released yet, so I guess we can create an issue for the future to update the k8s client with this fix

mvidalgarcia: Can you share the link?

reana_server/complexity.py
audrium: Well spotted, indeed this gives the same result, but in a much nicer way

reana_server/rest/workflows.py
```diff
@@ -918,12 +924,29 @@ def start_workflow(workflow_id_or_name, user):  # noqa
...
+        # Estimate complexity
+        complexity = estimate_complexity(workflow.type_, workflow.reana_specification)
+        workflow.complexity = json.dumps(complexity)
+        workflow.status = RunStatus.queued
+        workflow_priority = calculate_workflow_priority(complexity)
+        workflow_min_job_memory = min(complexity, key=lambda x: x[1])[1]
+        Session.commit()
```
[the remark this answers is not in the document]
audrium: If feels not right to me as well, but moving these few lines somewhere else is also weird

reana_server/scheduler.py
mvidalgarcia: We can pass the arg to `reana_ready` I guess. Then we can create a [partial](https://docs.python.org/3/library/functools.html#functools.partial) function including the arg and include it in the array of function conditions.

reana_server/status.py
audrium: Here is a link to PR which was merged recently: https://github.com/kubernetes/kubernetes/pull/102159

reana_server/config.py
```diff
@@ -38,6 +38,13 @@
 
 REANA_SSO_CERN_CONSUMER_SECRET = os.getenv("CERN_CONSUMER_SECRET", "CHANGE_ME")
 
+REANA_KUBERNETES_JOBS_MEMORY_LIMIT = os.getenv("REANA_KUBERNETES_JOBS_MEMORY_LIMIT", "4Gi")
```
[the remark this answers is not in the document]
audrium: I've renamed it to `REANA_KUBERNETES_DEFAULT_JOBS_MEMORY_LIMIT` for now, since this is used only for the complexity estimation in r-server. That's why initially I left the value hardcoded in the code, since having the config could lead to some inconsistencies. But I guess changing the name as you suggested could be enough to avoid it

reana_server/complexity.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+def calculate_workflow_priority(complexity):
+    """Calculate workflow priority based on its complexity."""
+    if not complexity:
+        return 0
+    wf_memory = int(reduce(lambda sum, item: sum + item[0] * item[1], complexity, 0))
+    total_cluster_memory = NodesStatus().get_total_memory()
+    if wf_memory > total_cluster_memory:
+        return 1
```
mvidalgarcia, line 39: Why 1 here? Is this to force a reschedule?

reana_server/config.py
```diff
@@ -38,6 +38,15 @@
 
 REANA_SSO_CERN_CONSUMER_SECRET = os.getenv("CERN_CONSUMER_SECRET", "CHANGE_ME")
 
+REANA_KUBERNETES_DEFAULT_JOBS_MEMORY_LIMIT = os.getenv(
```
mvidalgarcia, line 41: I was more thinking of creating a new variable instead, and plugging it into `REANA_KUBERNETES_JOBS_MEMORY_LIMIT` as default. Something like this:

```python
REANA_KUBERNETES_DEFAULT_JOBS_MEMORY_LIMIT = '4Gi'
REANA_KUBERNETES_JOBS_MEMORY_LIMIT = os.getenv("REANA_KUBERNETES_JOBS_MEMORY_LIMIT", REANA_KUBERNETES_DEFAULT_JOBS_MEMORY_LIMIT)
```

If it's only going to be used for complexity calculations in reana-server, maybe we can even call it differently, something like `REANA_COMPLEXITY_JOBS_MEMORY_LIMIT` and keep what you have now.

reana_server/rest/workflows.py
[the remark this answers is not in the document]
mvidalgarcia: I'll probably need to change it in my PR since `calculate_workflow_priority` will come from reana-db.

reana_server/scheduler.py
```diff
@@ -71,11 +80,12 @@ def doesnt_exceed_max_reana_workflow_count():
     return True
 
 
-def reana_ready():
+def reana_ready(workflow_min_job_memory):
     """Check if REANA can start new workflows."""
     for check_condition in [
         check_predefined_conditions,
         doesnt_exceed_max_reana_workflow_count,
+        partial(check_memory_availability, workflow_min_job_memory),
```
mvidalgarcia, line 88: Nice!

reana_server/status.py
```diff
@@ -399,14 +416,27 @@ def get_memory_usage(self):
                     kubernetes_memory_to_bytes(node_capacity),
                 )
                 result[node_name]["percentage"] = f"{node_usage_percentage}%"
+                result[node_name]["available"] = node_capacity - node_usage
         except ApiException as e:
             msg = "Error while calling `metrics.k8s.io` API."
             logging.error(msg)
             logging.error(e)
             return {"error": msg}
+        except Exception as e:
```
mvidalgarcia, line 425: What is the exception triggered? Perhaps we can be more accurate here.

tests/test_complexity.py
```diff
@@ -0,0 +1,58 @@
...
+        ([(5, 2), (5, 2)], 95),
+        ([], 0),
+        ([(1, 396)], 1),
+        ([(1, 401)], 1),
+    ],
+)
+def test_calculate_workflow_priority(complexity, priority):
+    """Test calculate_workflow_priority."""
+    with patch(
+        "reana_server.status.NodesStatus.get_total_memory", return_value=400,
```
mvidalgarcia, line 49: You can also parametrize the total memory value to have some other examples.

reana_server/complexity.py
audrium: This is a protection not to exceed over `MQ_MAX_PRIORITY` (which is 100 in our case). If a workflow requires more memory than a cluster can offer, then by our formula `int((1 - wf_memory / total_cluster_memory) * MQ_MAX_PRIORITY)` the priority could be set higher than `MQ_MAX_PRIORITY` and break the scheduler. So in this case I set it to 1, since it's a minimum priority there could be (it could be also set to 0, but that would more mean that there is no priority for this workflow)

reana_server/status.py
tiborsimko: Option A: One technique would be to return "0 GiB", which would mean "cannot detect", which would force to put all workflows into requeuing until we can detect.  This is not ideal, but could relieve troubles in case the API error occurs only infrequently.

Option B: We should probably memoise the results of `get_available_memory()` in a module-wide variable, and, if this exception occurs, we can take the last-detected value as a good approximation of what the free memory might look like.  Not ideal, but better than a fixed amount.

I guess B would be preferable over A.

reana_server/config.py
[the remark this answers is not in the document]
tiborsimko: I agree with @mvidalgarcia that it is better to set default in Helm values, should we want to use one. That would be the cleanest solution.

There may be some installations that don't want to use any RAM limit, so in theory this value could be "0 GiB"  by default for some.  We should be ready for this case when calculating priorities. (E.g. send all workflows to the highest bin for this case.)

So I'd suggest to reuse the Helm value here, and be ready for 0 GiB case (which is currently the default).

P.S. Alternatively, if this value is used only to calculate priorities, we may want to use whatever and rename the variable somewhat differently so that it wouldn't evoke the "hard limit" feeling. But I'd prefer to make it consistent with Helm value anyway...

reana_server/status.py
mvidalgarcia: Does option B have to be implemented as part of this PR?

reana_server/complexity.py
```diff
@@ -0,0 +1,360 @@
...
+
+    estimator = build_estimator(workflow_type, reana_yaml)
+    try:
+        complexity = estimator.estimate_complexity()
+    except Exception:
+        return []
+    return complexity
+
+
+class ComplexityEstimatorBase:
```
mvidalgarcia, line 70: I haven't look much into the complexity estimator classes. I assume this was already reviewed as part of the reana-client old PR. Is this right? cc/ @tiborsimko

reana_server/status.py
mvidalgarcia: We can add a FIXME to revisit it once the new Kubernetes version is released.

mvidalgarcia (approved):

mvidalgarcia:
Please add a CHANGELOG entry (or multiple) as this is a quite significant change.

reana_server/status.py
audrium: In this case it wouldn't help to wait until Kubernetes release, since `get_memory_usage()` could also fail while calling `metrics.k8s.io` API. So we would still need to handle these kind of errors.

Current logic to check if workflow could be started in Kubernetes (executed in `reana_ready()`):
```max_node_available_memory >= workflow_min_job_memory```
So returning all the nodes memory in this case would mean to start all the incoming workflows

File: reana_server/complexity.py (hunk 21 above)
Comment (tiborsimko) [reply reana_server/complexity.py:70]: Yes, I tested several scenarios, and the complexity estimator returning information `[(no_jobs_1, memory_1), (no_jobs_2, memory_2), ...] ` for the initial workflow stages (or later stages) was fully OK then :+1:

File: reana_server/status.py (hunk 2 above)
Comment (audrium) [reply reana_server/status.py]: Remade the logic to fall back to FIFO strategy in case any errors like this occurs and added couple of `#FIXME`ies to adopt the code later to a new Kubernetes release

File: reana_server/complexity.py [hunk 22]
```diff
@@ -0,0 +1,360 @@
...
+                "dependencies": list(dependencies),
+                "scatter_params": scatter_params,
+                "workflow": nested_workflow,
+            }
+        return tree
+
+    def _parse_workflow(self, workflow):
+        """Parse CWL workflow specification tree."""
+        tree = {}
+        if isinstance(workflow, dict):
```
mr-c, line 308: I highly recommend using our autogenerated classes at https://github.com/common-workflow-language/cwl-utils/blob/main/cwl_utils/parser_v1_1.py instead of parsing CWL yourself. I can see that this code doesn't support CWL sub-workflows, for example.

audrium: Thanks a lot for your suggestion @mr-c ! We will check the parser out and hopefully integrate it into our code. Funny enough, I am implementing nested workflow support at the moment :)

audrium: @mr-c, what would be the main benefit of using this parser from `cwl_utils`? Currently we are parsing the `$graph` which we get by calling `cwltool --pack --quiet workflow_file`. I just checked the output of `cwl_utils` parser and it gives pretty much the same information as the one which we use. Maybe I'm missing something?

mr-c: In general, the benefit is concrete classes, and that the schema-salad transformations are already performed (type DSLs expanded to their normalized form, ` $include` and `$import` acted upon). Likewise the document undergoes basic validation and any errors result in specific messages. However, you do get almost all of that from `cwltool --pack` besides the nice concrete classes instead of untyped dictionaries. So maybe there isn't enough to justify the change in this instance.

However, if later other types of analysis is needed, then you may benefit from the `cwl-utils` approach, so maybe leave a comment mentioning that?

audrium: Thanks for the explanation! I guess we can revive this topic once upgrading the [cwltool version to the latest](https://github.com/reanahub/reana-workflow-engine-cwl/issues/187)

PR #367, opened 2021-05-21
rest: move complexity priority calculation to r-db

### LINKED ISSUE #356: scheduler: max number of concurrent workflows per user
Currently, the max number of concurrent workflows is only limited globally by [`REANA_MAX_CONCURRENT_BATCH_WORKFLOWS`](https://github.com/reanahub/reana/blob/master/helm/reana/README.md). We shall introduce a mechanism to limit the number of concurrent workflows a user can have running at a given point in time.

Possible solutions:
- Introduce a new helm value to allow setting empty-slot-consumption time in case of no new users
- Perhaps an addition to user quotas along the lines of disk usage limit https://github.com/reanahub/reana-server/issues/302
- Use intelligently auto-discovered values to decide whether to run the workflow or not. Use case: Scenario in which a user (e.g. CERN physicist) has no maximum concurrent workflows limit, this method should be used.
    - Based on job slots

Observations:
- `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` perhaps to be removed? The application should still decide and control queued workflows (`queued` -> `pending` -> `running`).

mvidalgarcia:
Closes #356

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -39,10 +39,10 @@ from reana_server.api_client import (
 )
 from reana_server.complexity import (
     estimate_complexity,
-    calculate_workflow_priority,
     get_workflow_min_job_memory,
 )
 from reana_server.decorators import check_quota, signin_required
+from reana_server.status import NodesStatus
 from reana_server.utils import (
     RequestStreamWithLen,
     _get_reana_yaml_from_gitlab,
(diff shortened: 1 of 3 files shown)


reana_server/rest/workflows.py
```diff
@@ -898,6 +898,15 @@ def start_workflow(workflow_id_or_name, user):  # noqa
                 "message": "Status resume is not supported yet."
               }
     """
+
+    def _calculate_complexity(workflow):
+        """Place workflow in queue and calculate and set its complexity."""
```
audrium, line 903: Just a comment: it also silently stores loaded `reana_specification` for Yadage workflows in the database. It could not be obvious for someone later touching the code, so I wonder if it should be mentioned here?

audrium:
Since kubernetes monkeypatch is implemented, maybe we can remove `#FIXME`ies inside `NodesStatus()` class related to the same issue?

reana_server/rest/workflows.py
mvidalgarcia: Doesn't this happen [here](https://github.com/reanahub/reana-server/pull/367/files#diff-16a8293e6225f903f0412763851a5589f06f6f8891689da7abd0cde7a8166c52R940)? Or you mean the commit?

mvidalgarcia:
> Since kubernetes monkeypatch is implemented, maybe we can remove `#FIXME`ies inside `NodesStatus()` class related to the same issue?

Eventually we will adapt the FIXMEs as well so it's good to have pointers where things may need to be modified.

reana_server/rest/workflows.py
audrium: Yes, I mean it was manipulated in one place and committed in another. I guess it's not a big deal, but could be misleading for someone perhaps. I like your solution, it's much cleaner! The only downsize is that it introduces another DB transaction, but I guess it's fine in this case, WDYT?

mvidalgarcia: Yes, I'm aware. OTOH, if something goes wrong in the second transaction we will have the reana spec saved.

audrium (approved):
LGTM :rocket:

PR #370, opened 2021-06-04
scheduler: delay rescheduled workflows and count number of retries

audrium:
Depends on: https://github.com/reanahub/reana-commons/pull/266

Closes https://github.com/reanahub/reana-server/issues/118

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -180,5 +180,8 @@ ADMIN_EMAIL = os.getenv("REANA_EMAIL_SENDER", "CHANGE_ME")
 
 # Workflow scheduler
 # ==================
-REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY = 1
+REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY = 60
 """How many seconds to wait between retries in case of REANA not ready to run more workflows."""
+
+REANA_SCHEDULER_RETRY_DELAY = 60000
+"""How many milliseconds for a workflow to be delayed once rescheduled."""
--- a/reana_server/scheduler.py
+++ b/reana_server/scheduler.py
@@ -28,7 +28,10 @@ from reana_server.api_client import (
     current_rwc_api_client,
     current_workflow_submission_publisher,
 )
-from reana_server.config import REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY
+from reana_server.config import (
+    REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY,
+    REANA_SCHEDULER_RETRY_DELAY,
+)
 from reana_server.status import NodesStatus
 
 
@@ -141,10 +144,10 @@ class WorkflowExecutionScheduler(BaseConsumer):
                 kwargs["parameters"],
                 priority=kwargs.get("priority", 0),
                 min_job_memory=kwargs.get("min_job_memory", 0),
+                retry_count=kwargs.get("retry_count", 0) + 1,
+                delay=REANA_SCHEDULER_RETRY_DELAY,
             )
-            logging.error(
-                f"Requeueing workflow " f'{kwargs["workflow_id_or_name"]} ...'
-            )
+            logging.info(f"Requeueing workflow " f'{kwargs["workflow_id_or_name"]} ...')
         except KeyError:
             logging.error(
                 f"Wrong parameters to requeue workflow:\n"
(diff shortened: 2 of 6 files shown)


reana_server/scheduler.py
```diff
@@ -141,6 +144,8 @@ def requeue_workflow(self, **kwargs):
                 kwargs["parameters"],
                 priority=kwargs.get("priority", 0),
                 min_job_memory=kwargs.get("min_job_memory", 0),
+                retry_count=kwargs.get("retry_count", 0) + 1,
+                delay=REANA_SCHEDULER_RETRY_DELAY,
             )
             logging.error(
                 f"Requeueing workflow " f'{kwargs["workflow_id_or_name"]} ...'
```
tiborsimko, line 151: BTW as part of this PR we could perhaps clean the logs.  Currently, when REANA cluster is not ready, I'm seeing:

```
2021-06-10 14:19:15,385 | root | MainThread | INFO | REANA not ready to run workflow hello.3. Requeueing workflow and retrying in 1 second(s) ...
2021-06-10 14:19:15,399 | root | MainThread | ERROR | Requeueing workflow hello.3 ...
2021-06-10 14:19:20,559 | root | MainThread | INFO | REANA not ready to run workflow hello.2. Requeueing workflow and retrying in 1 second(s) ...
2021-06-10 14:19:20,579 | root | MainThread | ERROR | Requeueing workflow hello.2 ...
2021-06-10 14:20:15,409 | root | MainThread | INFO | REANA not ready to run workflow hello.3. Requeueing workflow and retrying in 1 second(s) ...
2021-06-10 14:20:15,424 | root | MainThread | ERROR | Requeueing workflow hello.3 ...
2021-06-10 14:20:20,590 | root | MainThread | INFO | REANA not ready to run workflow hello.2. Requeueing workflow and retrying in 1 second(s) ...
2021-06-10 14:20:20,604 | root | MainThread | ERROR | Requeueing workflow hello.2 ...
2021-06-10 14:21:15,433 | root | MainThread | INFO | REANA not ready to run workflow hello.3. Requeueing workflow and retrying in 1 second(s) ...
2021-06-10 14:21:15,447 | root | MainThread | ERROR | Requeueing workflow hello.3 ...
2021-06-10 14:21:20,615 | root | MainThread | INFO | REANA not ready to run workflow hello.2. Requeueing workflow and retrying in 1 second(s) ...
2021-06-10 14:21:20,630 | root | MainThread | ERROR | Requeueing workflow hello.2 ...
2021-06-10 14:22:15,456 | root | MainThread | INFO | REANA not ready to run workflow hello.3. Requeueing workflow and retrying in 1 second(s) ...
2021-06-10 14:22:15,470 | root | MainThread | ERROR | Requeueing workflow hello.3 ...
```

The ERROR level seems to say that there was something wrong, but actually it is only the requeueing that is happening... So something like INFO displayed above is already enough.  Two options: (1) delete the ERROR message; (2) turn it into INFO or WARNING rather than ERROR.  Let's keep ERROR only for situations where our code base may malfunction...

tiborsimko: Otherwise the retry count works well for me :+1:

reana_server/config.py
```diff
@@ -182,3 +182,6 @@ def _(x):
 # ==================
 REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY = 1
 """How many seconds to wait between retries in case of REANA not ready to run more workflows."""
```
tiborsimko, line 184: One second when REANA is not ready is perhaps too soon?  Imagine getting RAM information etc... Quite a few checks to perform.  I guess we could set it to 60 seconds too, as the delaying?

tiborsimko: BTW we may want to expose these to REANA cluster admins so that the values would be configurable via Helm values. (This could be also done after merge via separate issue.)

reana_server/scheduler.py
audrium: Changed the logging level to `INFO`

reana_server/config.py
audrium: >BTW we may want to expose these to REANA cluster admins so that the values would be configurable via Helm values. (This could be also done after merge via separate issue.)

Mentioned in [this issue](https://github.com/reanahub/reana/issues/509), could be fixed in one go

tiborsimko (approved):

reana_server/config.py
mvidalgarcia: I'm probably wrong but would this be blocking the scheduler for 60 seconds?
What is the difference between `REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY` and `REANA_SCHEDULER_RETRY_DELAY`?

tiborsimko: For production-like conditions we could use 60 seconds retry and 600 seconds queueing time; for developer-like conditions we could use 6 seconds retry and 60 seconds queueing time perhaps?

mvidalgarcia: Blocking the scheduler for 60 seconds if a workflow gets rescheduled in a dev environment seems quite a lot. Or perhaps I'm not understanding the working properly.

PR #371, opened 2021-06-08
scheduler: fix query counting running workflows

tiborsimko:
Optimises SQL query to count running or pending workflows. Avoids ORM
technique that would lead to subselects.

Closes DB session explicitly. This fixes observed "idle in transaction"
issues.

### DIFF
--- a/reana_server/scheduler.py
+++ b/reana_server/scheduler.py
@@ -15,12 +15,13 @@ from functools import partial
 
 from bravado.exception import HTTPBadGateway, HTTPNotFound
 from kubernetes.client.rest import ApiException
-from sqlalchemy import or_
+from sqlalchemy import func, or_
 from sqlalchemy.exc import SQLAlchemyError
 
 from reana_commons.config import REANA_MAX_CONCURRENT_BATCH_WORKFLOWS
 from reana_commons.consumer import BaseConsumer
 from reana_commons.k8s.api_client import current_k8s_corev1_api_client
+from reana_db.database import Session
 from reana_db.models import Workflow, RunStatus
 
 from reana_server.api_client import (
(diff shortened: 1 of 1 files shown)


reana_server/scheduler.py
```diff
@@ -64,22 +65,28 @@ def check_predefined_conditions():
...
+            doesnt_exceed = False
     except SQLAlchemyError as e:
         logging.error(
             "Something went wrong while querying for number of running workflows."
         )
         logging.error(e)
-        return False
-    return True
+        doesnt_exceed = False
+    Session.commit()
```
audrium, line 88: Why `Session.commit()` is needed here? Since we are not mutating the query, isn't Session.close() enough in this case?

mvidalgarcia: In normal conditions, nothing should be needed... I guess in this context `commit` has the same behavior as `closes`. It'd be nice to know why it's needed..

tiborsimko: I thought about commit mostly because of what might happen outside of this function by the callers. If this function gets used during long chain of calls/contexts, the session may perhaps be used later, so a commit rather than a flat-out close seemed "safer". (That said, if we revive `pool_pre_ping=True` option to the DB engine, for other reasons, then we could do either...)

reana_server/scheduler.py
```diff
@@ -64,22 +65,28 @@ def check_predefined_conditions():
...
 def doesnt_exceed_max_reana_workflow_count():
     """Check upper limit on running REANA batch workflows."""
+    doesnt_exceed = True
     try:
-        running_workflows = Workflow.query.filter(
-            or_(
-                Workflow.status == RunStatus.pending,
-                Workflow.status == RunStatus.running,
+        running_workflows = (
+            Session.query(func.count())
```
mvidalgarcia, line 71: Just curious, how does it know which table it needs to query? Should we point that we want to query against `Workflow` somehow?

mvidalgarcia (approved):
Tested [alembic recipe](https://github.com/reanahub/reana-db/blob/master/reana_db/alembic/versions/20210607_1217_f84e17bd6b18_workflow_complexity.py), downgrade and upgrade work nicely after creating some workflows.

reana_server/scheduler.py
tiborsimko: It knows by the `filter()`, but perhaps we could do write `func.count(Workflow.id_)` to make it very explicit there as well... I can retest the speed of the two.

tiborsimko: Just for the record, I tested `func.count(Workflow.id_)`, the generated query look good:

```
SELECT count(*) AS count_1 
FROM __reana.workflow 
WHERE __reana.workflow.status = %(status_1)s OR __reana.workflow.status = %(status_2)s

vs

SELECT count(__reana.workflow.id_) AS count_1 
FROM __reana.workflow 
WHERE __reana.workflow.status = %(status_1)s OR __reana.workflow.status = %(status_2)s
```

The performance (on a small table) was a bit slower though. The differences were not much, of the order of microseconds, but for extra safety for big tables, perhaps we could leave the current version...  I guess `COUNT(*)` might be slightly more optimised than `COUNT(somecolumn)` in RDBMS implementations in general.

Issue #368, opened 2021-06-01
gitlab: load yadage spec and calculate workflow complexity


audrium:
Workflows created through `Gitlab` uses single `create_workflow()` endpoint to be registered in the database and [started](https://github.com/reanahub/reana-server/blob/master/reana_server/rest/workflows.py#L409) while the other workflows uses `start_workflow()` endpoint.

[Here](https://github.com/reanahub/reana-server/issues/358) we have introduced some additional logic to the `start_workflow()` endpoint to:
- load Yadage specification
- calculate the complexity
- set workflow priority

We need to do the same for gitlab workflows by reusing the code from `start_endpoint()` 

_Originally posted by @audrium in https://github.com/reanahub/reana-server/pull/366#discussion_r641584752_

PR #373, opened 2021-06-11
gitlab: load yadage spec and calculate workflow complexity

audrium:
closes https://github.com/reanahub/reana-server/issues/368

This PR needs to be tested on dev instance in order to make sure that Gitlab integration works as expected

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -18,7 +18,6 @@ from flask import current_app as app
 from flask import jsonify, request, stream_with_context
 from reana_commons.errors import REANAQuotaExceededError, REANAValidationError
 from reana_commons.operational_options import validate_operational_options
-from reana_commons.yadage import yadage_load_from_workspace
 from reana_db.database import Session
 from reana_db.models import (
     InteractiveSessionType,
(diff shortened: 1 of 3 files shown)


mvidalgarcia:
Needs rebase.

reana_server/rest/workflows.py
```diff
@@ -413,13 +411,21 @@ def create_workflow(user):  # noqa
...
+            total_cluster_memory = NodesStatus().get_total_memory()
+            workflow_priority = workflow.get_priority(total_cluster_memory)
+            workflow_min_job_memory = get_workflow_min_job_memory(complexity)
             current_workflow_submission_publisher.publish_workflow_submission(
                 user_id=str(user.id_),
-                workflow_id_or_name=response["workflow_id"],
+                workflow_id_or_name=workflow.get_full_workflow_name(),
                 parameters=request.json,
+                priority=workflow_priority,
+                min_job_memory=workflow_min_job_memory,
```
mvidalgarcia, line 428: Since this is repeated as well in `start_workflow`, maybe we can extract it and reuse it in both places?

It's a pity that we don't have [Flask Views](https://flask.palletsprojects.com/en/2.0.x/views/).. that way we could leverage inheritance to create reusable functions and keep everything organized. This would be an important step towards a model-view-controller architecture as we could extract business logic to different places and avoid having e.g `Session.commit()`, etc. in this layer. Just dumping ideas we might need to consider for the future. cc/ @tiborsimko @ParthS007 WDYT?

mvidalgarcia (commented):
Left a couple of comments, one of them more philosophical with a view to improving our software architecture in the future.

I cannot test it properly because my instance is not reachable by gitlab-dev.cern.ch.

I added "CERN integrations locally (GitLab)" to my wiki in case you want to check it out or give it a try - https://github.com/mvidalgarcia/reana/wiki/CERN-integrations-(GitLab)

mvidalgarcia (approved):
Tested together with @audrium in the CERN network and working both serial and yadage workflows ✅

PR #374, opened 2021-06-11
config: remove default k8s memory limit

mvidalgarcia:
Rely on the default Helm k8s memory limit instead.

closes reanahub/reana#509

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -178,5 +178,5 @@ ADMIN_EMAIL = os.getenv("REANA_EMAIL_SENDER", "CHANGE_ME")
 REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY = 60
 """How many seconds to wait between retries in case of REANA not ready to run more workflows."""
 
-REANA_SCHEDULER_RETRY_DELAY = 60000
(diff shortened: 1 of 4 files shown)


reana_server/config.py
```diff
@@ -180,8 +175,12 @@ def _(x):
 
 # Workflow scheduler
 # ==================
-REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY = 60
+REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY = os.getenv(
+    "REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY", 600
+)
```
audrium, line 180: I wonder whether we should move the default production values to Helm as well? It's a bit strange to rely on non-existing env in production?

mvidalgarcia: I thought the same, but then the tests were failing as the value was `None`. I can switch it back and mock the tests.

mvidalgarcia: I've changed it and the [reana](https://github.com/reanahub/reana/compare/2876fc9dfcc88c886e45e4e568b456c5d67297a0..2235424045d01483343e14cd90ea9e6cc292c37e) one accordingly.

audrium (approved):

PR #376, opened 2021-06-14
complexity: make compatible with yadage string-like resources

audrium:
closes https://github.com/reanahub/reana-workflow-engine-yadage/issues/188


tiborsimko (approved):
LGTM, please rebase against latest master

Issue #357, opened 2021-05-06
scheduler: start most appropriate workflow based on a desired multi-user scheduling strategy


tiborsimko:
The queue for incoming workfows may contain N workfows. We should decide upon optimal strategy which workflow to start.

It may not be always FIFO. For example, if John runs 20 workflows, and has 10 waiting in the queue, and Jane has 1 waiting in the queue, then we should give priority to Jane's workflow over yet another John's workflow when a running slot liberates.

(This also depends on the complexity of the workflow itself, and whether we have enough job slots to start the first stages of the said workflow. If not, then the jobs would be in `pending` state unnecessarily.)

This is an umbrella ultimate-goal issue that needs many separate smaller tasks as dependencies before it can be implemented.

Note that workflow starting strategies may be made configurable via Helm values so that each REANA cluster admin will select what is the most appropriate for the REANA instance at hand, say "fifo" for first-in-first-out regardless of users, say "balanced" for being better multi-user-aware, etc.

tiborsimko:
Now that every feature was implemented, the only remaining thing is:

- introduce helm variable to select workflow scheduling policy; offer two options, "fifo" and "balanced";
- write documentation for the workflow scheduling policy options for `docs.reana.io`;
- amend workflow scheduler to run the DAG and user complexity only for "balanced" method, to use the same priority bin for "fifo" method, and to reply an error for non-recognised methods
- (note that this may necessitate a PR to `reana-commons` to define a list of all possible workflow scheduling policies that the user will be able to choose from)

PR #377, opened 2021-06-16
complexity: take into account workflow scheduling policy

audrium:
closes https://github.com/reanahub/reana-server/issues/357

Depends on: https://github.com/reanahub/reana-commons/pull/271

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -41,6 +41,14 @@ REANA_SSO_CERN_CONSUMER_SECRET = os.getenv("CERN_CONSUMER_SECRET", "CHANGE_ME")
 REANA_COMPLEXITY_JOBS_MEMORY_LIMIT = os.getenv("REANA_KUBERNETES_JOBS_MEMORY_LIMIT")
 """Maximum memory limit for user job containers for workflow complexity estimation."""
 
+REANA_WORKFLOW_SCHEDULING_POLICY = os.getenv("REANA_WORKFLOW_SCHEDULING_POLICY", "fifo")
+
+REANA_WORKFLOW_SCHEDULING_POLICIES = ["fifo", "balanced"]
+"""REANA workflow scheduling policies.
+- ``fifo``: first-in first-out strategy starting workflows as they come.
+- ``balanced``: a weighted strategy taking into account existing multi-user workloads and the DAG complexity of incoming workflows.
+"""
+
 
 # Invenio configuration
 # =====================
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -90,7 +90,7 @@ def get_usage_percentage(usage, limit):
 def publish_workflow_submission(workflow, user_id, total_cluster_memory, parameters):
     """Publish workflow submission."""
     complexity = _calculate_complexity(workflow)
-    workflow_priority = workflow.get_priority(total_cluster_memory)
+    workflow_priority = workflow.get_priority(total_cluster_memory) if complexity else 0
     workflow_min_job_memory = get_workflow_min_job_memory(complexity)
     current_workflow_submission_publisher.publish_workflow_submission(
         user_id=str(user_id),
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -45,6 +45,8 @@ from reana_server.config import (
     REANA_GITLAB_URL,
     REANA_HOSTNAME,
     REANA_USER_EMAIL_CONFIRMATION,
+    REANA_WORKFLOW_SCHEDULING_POLICY,
+    REANA_WORKFLOW_SCHEDULING_POLICIES,
 )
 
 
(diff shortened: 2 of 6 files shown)


reana_server/config.py
```diff
@@ -41,6 +41,8 @@
 REANA_COMPLEXITY_JOBS_MEMORY_LIMIT = os.getenv("REANA_KUBERNETES_JOBS_MEMORY_LIMIT")
 """Maximum memory limit for user job containers for workflow complexity estimation."""
 
+REANA_WORKFLOW_SCHEDULING_POLICY = os.getenv("REANA_WORKFLOW_SCHEDULING_POLICY")
```
tiborsimko, line 44: What about falling back explicitly to "fifo" in case the variable does not exist for a reason or another?

```python
REANA_WORKFLOW_SCHEDULING_POLICY = os.getenv("REANA_WORKFLOW_SCHEDULING_POLICY", "fifo")
```
        
          
            
            
            

    

  
      
        
        
        

          
          
          
          
          
          
          



        
  
  
      
        
```

Comment (mvidalgarcia) [reply reana_server/config.py]: A couple of things:
- If the [default value is `balanced`](https://github.com/reanahub/reana/pull/511/files#diff-8a781e33aeeaa4deddd5a0f1fd47e3c6e85de8f7e65e34916f12eef80fa447efR74) why should we fall back to `fifo`?
- Perhaps we can perform the validation already here and fall back to a known value to prevent [raising an exception](https://github.com/reanahub/reana-server/pull/377/files#diff-dd451815d7362d3092b6e4d2bc6880a2fe0698d82d2415e90b8adc23181eb3faR52)? OT1H it's good to be aware that a wrong policy value was passed but OTOH it'd be nice to prevent the cluster from malfunctioning if a wrong config is set.

File: reana_server/utils.py [hunk 2]
```diff
@@ -90,7 +90,7 @@ def get_usage_percentage(usage, limit):
 def publish_workflow_submission(workflow, user_id, total_cluster_memory, parameters):
     """Publish workflow submission."""
     complexity = _calculate_complexity(workflow)
```
mvidalgarcia, line 92: I guess we don't even need to run `_calculate_complexity` if we're in `fifo` mode but just set the priority to 0? It'd improve the performance. 

Just watch out because then we would need to move the `RunStatus.queued` assignment outside. I think we can just use `Workflow.update_workflow_status`

reana_server/config.py
tiborsimko: Yeah, the default should be the same, "fifo" or "balanced" everywhere.  I guess we could stay on the conservative "fifo" value until we have full CWL DAG complexity set up... and use "balanced" only at CERN.

reana_server/utils.py
audrium: > I guess we don't even need to run `_calculate_complexity` if we're in `fifo` mode but just set the priority to 0? It'd improve the performance.

Yes, that's why I've added this peace of code to `estimate_complexity()` before any calculations:
```python
# No need to estimate the complexity for "fifo" strategy
if scheduling_policy == "fifo":
    return []
```

> It's a bit strange that the logic to decide whether we should calculate complexity is inside a function called `_calculate_complexity`/`estimate_complexity`.

I was thinking where to put the policy check and for me it looked cleaner to keep this logic inside `estimate_complexity()` rather than in `utils.py`, but it could be done either way I guess

reana_server/config.py
audrium: >     * Perhaps we can perform the validation already here and fall back to a known value to prevent [raising an exception](https://github.com/reanahub/reana-server/pull/377/files#diff-dd451815d7362d3092b6e4d2bc6880a2fe0698d82d2415e90b8adc23181eb3faR52)? OT1H it's good to be aware that a wrong policy value was passed but OTOH it'd be nice to prevent the cluster from malfunctioning if a wrong config is set.

I think if we fallback silently it could be tricky to spot it in case some user makes a typo while defining the policy. Imagine someone who set "blanced" would run "fifo" without knowing it

reana_server/utils.py
mvidalgarcia: IMO, if the policy is `fifo`, there shouldn't be any complexity estimation function involved.

I understand your point, it's just a matter of keeping things separated and more legible. I'd like to have more points of view on this. cc/ @tiborsimko

mvidalgarcia: Refactor done [here](https://github.com/reanahub/reana-server/pull/377/commits/a029855d14d42ca187236bbfffcb4406ff27f003).

mvidalgarcia (approved):

reana_server/utils.py
tiborsimko: Tested the update, and I saw `r-server` scheduler pod issues:

```
$ kubectl logs reana-server-694d7955ff-7gck7 scheduler
...
  File "/usr/local/lib/python3.8/site-packages/kombu/messaging.py", line 590, in <listcomp>
    [callback(body, message) for callback in callbacks]
  File "/usr/local/lib/python3.8/site-packages/reana_server/scheduler.py", line 236, in on_message
    sleep(REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY)
TypeError: an integer is required (got type str)
```

which "self-healed" after pod restarted a few times...

At least I've got to test the how the web ui looks like :wink: 

reana-service-unavailable

mvidalgarcia: Maybe you didn't have the latest helm? That value should come from values.yaml

https://github.com/reanahub/reana/blob/2235424045d01483343e14cd90ea9e6cc292c37e/helm/reana/values.yaml#L62-L63

tiborsimko: I had, I tested with:

```console
$ reana-dev git-status -c ALL | grep -v '@ master'
docs.reana.io @ pr-84 @ f7cdef9 configuration: added configuring workflow scheduler page
reana-ui @ release-0.8.0-alpha.2 @ c9893b5 release: 0.8.0-alpha.2
reana-workflow-engine-yadage @ pr-194 @ 44ea0de release: 0.8.0a2
reana-message-broker @ pr-38 @ a869111 release: 0.8.0a1
reana @ pr-511 (1 AHEAD upstream/pr/511) @ b70c554 WIP
reana-server @ pr-377 @ a029855 global: refactor scheduling policy logic
```

and the only difference was REANA_MAX_CONCURRENT_BATCH_WORKFLOWS set to 1.

I'll rerun to confirm/infirm...

P.S. Perhaps we are not doing str->int cast along the way?

tiborsimko: I confirm I can reproduce the problem with the above setup. The first `run-ci` command pass fine, the three workflows are successful, but then when I submit several workflows with different memory limit requirements to test the queueing and scheduling policy, I am getting:

```
$ kubectl get pods | grep reana-server 
reana-server-694d7955ff-drhs6                                1/2     CrashLoopBackOff   2          12m
$ kubectl logs reana-server-694d7955ff-drhs6 scheduler | tail -5
  File "/usr/local/lib/python3.8/site-packages/kombu/messaging.py", line 590, in <listcomp>
    [callback(body, message) for callback in callbacks]
  File "/usr/local/lib/python3.8/site-packages/reana_server/scheduler.py", line 236, in on_message
    sleep(REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY)
TypeError: an integer is required (got type str)
```

tiborsimko: > Perhaps we are not doing str->int cast along the way?

Indeed, in `reana_server/config.py` we are doing:

```python
REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY = os.getenv(
    "REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY"
)
```

which will produce a string not an integer... We need a (safe) cast there.

PR #378, opened 2021-06-16
api: add search param in get_workflow_disk_usage

ParthS007:
closes https://github.com/reanahub/reana-client/issues/522

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -2237,7 +2237,10 @@ def get_workflow_disk_usage(workflow_id_or_name, user):  # noqa
             raise ValueError("workflow_id_or_name is not supplied")
         workflow = _get_workflow_with_uuid_or_name(workflow_id_or_name, str(user.id_))
         summarize = bool(parameters.get("summarize", False))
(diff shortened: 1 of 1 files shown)


reana_server/rest/workflows.py
```diff
@@ -2237,7 +2237,10 @@ def get_workflow_disk_usage(workflow_id_or_name, user):  # noqa
             raise ValueError("workflow_id_or_name is not supplied")
         workflow = _get_workflow_with_uuid_or_name(workflow_id_or_name, str(user.id_))
         summarize = bool(parameters.get("summarize", False))
-        disk_usage_info = workflow.get_workspace_disk_usage(summarize=summarize)
+        search = parameters.get("search", False)
```
audrium, line 2240: Why default value for `search` is `False` here? Since the [default value for it in all the function parameters](https://github.com/reanahub/reana-db/pull/135/files#diff-9c9ecb5b02eee5df82a9c1138c56e62181e9d8d56152d45c09f7fe0a693ce980R586) are defined as `None`

ParthS007: Thanks for the catch. I will change it to `None`.

audrium (approved):

PR #379, opened 2021-06-17
config: fix helm value casting

tiborsimko:
Fixes casting of Helm values that were read as strings but that should
be treated as integers.


mvidalgarcia (approved):
Tested with maximum concurrent workflows to 1, working nicely when they are requeued.

PR #380, opened 2021-06-18
release: 0.8.0a2

tiborsimko:
Closes https://github.com/reanahub/reana/issues/512


mvidalgarcia (approved):
The `get_available_memory` fix seems to work fine with and without the metrics API installed in kind@macOS

PR #384, opened 2021-06-24
scheduler: remove retry_count from workflow submission event handler

audrium:
Depends on: https://github.com/reanahub/reana-commons/pull/272

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -44,12 +44,3 @@ def _get_user_mock():
     mocked_get_user = Mock(return_value=mocked_user)
     with patch("flask_login.utils._get_user", mocked_get_user):
         yield flask_login.utils._get_user
-
-
-@pytest.fixture()
-def in_memory_workflow_submission_publisher(in_memory_queue_connection):
-    with patch.dict(MQ_DEFAULT_QUEUES["workflow-submission"]["exchange"], {"name": ""}):
-        in_memory_wsp = WorkflowSubmissionPublisher(
-            connection=in_memory_queue_connection
-        )
-    return in_memory_wsp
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -44,12 +44,3 @@ def _get_user_mock():
...
         yield flask_login.utils._get_user
-
-
-@pytest.fixture()
-def in_memory_workflow_submission_publisher(in_memory_queue_connection):
-    with patch.dict(MQ_DEFAULT_QUEUES["workflow-submission"]["exchange"], {"name": ""}):
-        in_memory_wsp = WorkflowSubmissionPublisher(
-            connection=in_memory_queue_connection
-        )
-    return in_memory_wsp
```
mvidalgarcia, line 55: Could you explain the reasoning behind this change?

audrium: This fixture was added in [this commit](https://github.com/reanahub/reana-server/commit/f0d55e16ba01ae4e4ca53b823eaa56d6bd2572aa#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R49) to `patch`  `workflow-submission` exchange from [x-delayed-message](https://github.com/reanahub/reana-commons/pull/272/files#diff-ffebaaffa55ff06e2876b7237a50c60f1f02fa70d3c43710d5da9aeacb1fbae0L200) (which allowed to delay messages) to the [default one](https://github.com/reanahub/reana-commons/blob/f3469ab9bd51018ae1896a6788df55dc31e1be7c/reana_commons/config.py#L179) (which is "") in order to pass tests. Since we removed `x-delayed-message` exchange (in r-commons PR) and use the default one for `workflow-submission` queue, this fixture is no longer needed

mvidalgarcia: I see, thanks for the explanation!

mvidalgarcia (approved):
Test passing by installing r-commons PR locally ✅

Issue #375, opened 2021-06-14
dag: improve workflow complexity estimation for scatter-gathered CWL workflows


audrium:
MVP complexity estimation solution for CWL workflows was implemented in https://github.com/reanahub/reana-server/issues/361.
However due to lacking complex CWL workflow demos and non working [official example](https://www.commonwl.org/user_guide/23-scatter-workflow/index.html) scattering CWL workflows was poorly tested. Task is to double test CWL complexity estimation on complex nested and/or scattering workflows and to make sure it works as expected.

PR #385, opened 2021-06-30
complexity: improve complexity estimation for scatter-gathered CWL workflows

audrium:
closes https://github.com/reanahub/reana-server/issues/375

Improve complexity estimation to take into account workflow nesting and scatter-gathering for CWL workflows.

In order to test:
- Place `wdb` breakpoint [here](https://github.com/reanahub/reana-server/blob/master/reana_server/utils.py#L111) after `complexity = _calculate_complexity(workflow)`
- Change `fifo` value to something else in [this line](https://github.com/reanahub/reana-server/blob/master/reana_server/utils.py#L105) in order to skip the `if statement`
- Launch any `cwl` workflow and inspect the estimated complexity inside the debugger
- Also I've created this [test workflow](https://github.com/audrium/reana-demo-scattered-helloworld/blob/master/workflow/cwl/helloworld.cwl) - extended version of `helloworld demo` which have `scattered` and `nested` steps to test all the functionality. In order to use it simply clone the repo. [Here](https://view.commonwl.org/workflows/github.com/audrium/reana-demo-scattered-helloworld/blob/master/workflow/cwl/helloworld.cwl) is the workflow visualization.

### DIFF
--- a/reana_server/complexity.py
+++ b/reana_server/complexity.py
@@ -261,19 +261,23 @@ class CWLComplexityEstimator(ComplexityEstimatorBase):
         """Parse CWL workflow specification tree."""
         tree = {}
         steps = workflow.get("steps", [])
+        wid = workflow.get("id")
         for step in steps:
             name = step.get("id")
+            run = step.get("run")
             hints = step.get("hints", [{}]).pop()
             # Parse scatter params
             scatter = step.get("scatter")
             scatter_params = None
             if scatter:
-                scatter_params = next(
-                    filter(lambda p: p["id"] == scatter, step.get("in", [])), {},
-                ).get("source")
+                scatter_params = (
+                    next(filter(lambda p: p["id"] == scatter, step.get("in", [])), {},)
+                    .get("source")
+                    .split("/")
+                    .pop()
+                )
             # Parse nested workflows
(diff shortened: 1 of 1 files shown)


reana_server/complexity.py
```diff
@@ -309,25 +313,41 @@ def _populate_dependencies(self, steps):
...
     def _populate_complexity(self, steps):
         """Populate complexity to parsed CWL workflow tree steps."""
-        for step, value in steps.items():
-            scatter_params = value.get("scatter_params")
-            if scatter_params:
-                param_len = len(self.input_params.get(scatter_params, []))
-                if not param_len:
-                    continue
+
+        def _parse_steps(steps):
```
ParthS007, line 325: Minor: Maybe we can make this a separate method instead of nested one.

ParthS007 (approved):
Thanks for the PR, works great!

reana_server/complexity.py
audrium: It was made as a nested function because it is being called recursively. Same technique used for Yadage, so I would propose to leave it as is

tiborsimko (approved):
Rebased against latest `master` and checked basic text suite `reana-dev run-example -w cwl`.

PR #386, opened 2021-07-02
scheduler: skip available node memory calculation on fifo strategy

audrium:
`workflow_min_job_memory` is equal to [0 on fifo mode](https://github.com/reanahub/reana-server/blob/master/reana_server/utils.py#L104-L107)

### DIFF
--- a/reana_server/scheduler.py
+++ b/reana_server/scheduler.py
@@ -88,11 +88,13 @@ def doesnt_exceed_max_reana_workflow_count():
 
 def reana_ready(workflow_min_job_memory):
     """Check if REANA can start new workflows."""
-    for check_condition in [
-        check_predefined_conditions,
-        doesnt_exceed_max_reana_workflow_count,
-        partial(check_memory_availability, workflow_min_job_memory),
-    ]:
+    conditions = [check_predefined_conditions, doesnt_exceed_max_reana_workflow_count]
+
+    # Do not calculate memory availability on fifo strategy
+    if workflow_min_job_memory:
+        conditions.append(partial(check_memory_availability, workflow_min_job_memory))
+
+    for check_condition in conditions:
         if not check_condition():
             return False
     return True


reana_server/scheduler.py
```diff
@@ -31,6 +31,8 @@
 
 def check_memory_availability(workflow_min_job_memory):
     """Check if at least one workflow job could be started in Kubernetes."""
+    if not workflow_min_job_memory:
+        return True
```
mvidalgarcia, line 35: What do you think about performing this check already in `reana_ready()` and exclude `check_memory_availability` from the array of conditions there?

audrium: I was thinking about that option, but IMHO this approach is cleaner, since in `reana_ready()` we would have to jiggle with the conditions list. I came up with a solution like this:
```python
def reana_ready(workflow_min_job_memory):
    conditions = [check_predefined_conditions, doesnt_exceed_max_reana_workflow_count]
    if workflow_min_job_memory:
        conditions.push(partial(check_memory_availability, workflow_min_job_memory))

    for check_condition in conditions:
        if not check_condition():
            return False
    return True
```
Maybe you have something else in mind?

mvidalgarcia: Yeah, that solution looks pretty good to me, it's exactly what I had in mind.

For me, it's better readable like this as if the policy is `fifo` I don't think we should be calling `check_memory_availability` at all. It's just a matter of separating concepts.

It's true that your solution is cleaner in terms of involving a smaller diff, but conceptually looks unsuitable to me. We can keep what you have if you think I'm wrong, no problem from my side.

mvidalgarcia (approved):

reana_server/scheduler.py
audrium: You made a good point that on `fifo` we shouldn't call the `check_memory_availability()` at all.. Thanks. I will update the code on Monday

PR #388, opened 2021-07-07
snakemake: support workflow type

mvidalgarcia:
closes reanahub/reana-client#531

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -25,7 +25,7 @@ from reana_db.config import SQLALCHEMY_DATABASE_URI
 SQLALCHEMY_TRACK_MODIFICATIONS = False
 """Track modifications flag."""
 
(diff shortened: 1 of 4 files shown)


audrium (approved):
LGTM!

It would be ideal to keep only one list of the engines somewhere in `reana-commons`.. But unfortunately we couldn't use it from `reana` and `reana-client`, so I guess nothing better could be done then to keep 3 copies of the same list

audrium:
Just remembered the complexity estimation. Perhaps we should create a separate issue to implement that for snakemake?

mvidalgarcia:
> It would be ideal to keep only one list of the engines somewhere in `reana-commons`.. But unfortunately we couldn't use it from `reana` and `reana-client`, so I guess nothing better could be done then to keep 3 copies of the same list

Both `reana-client` and `reana-server` install `reana-commons` so we can move the list there at least.

Issue #393, opened 2021-08-11
api: make `reana-client du` command work if the workspace_path is outside `/var/reana`. 


marcdiazsan:
Currently `get_workspace_disk_usage()` is being called in `reana-server` which does not have additional volumes mounted. When is called for workflows where the workspace path is not the default:

```python
$ reana-client du -w workflow
Disk usage could not be retrieved:
Directory does not exist.
```
One possibility to solve it would be to create `get_workflow_disk_usage()` endpoint in `reana-workflow-controller` and call it from `reana-server` just like most of the other endpoints do: https://github.com/reanahub/reana-server/blob/b5c84d9adec1d1bf5c09cea359b29ba62d3749b6/reana_server/rest/workflows.py#L2242

mvidalgarcia:
Can we close this one? I see it's in the `Done` column?

marcdiazsan:
Yes, we can close it, as it was solved with [#531](https://github.com/reanahub/reana/pull/531)

PR #394, opened 2021-08-11
workflows: change the get disk usage endpoint to call the new get_workflow_disk_usage from rwc

marcdiazsan:
Closes reanahub/reana-server#393


marcdiazsan:
As reanahub/reana#531 mounts new directories in r-server this PR is not necessary

PR #395, opened 2021-08-18
workflows: extract workspace-root-path from a new top section from .yaml file

marcdiazsan:
It passes a workspace_root_path as a parameter to rwc

Closes reanahub/reana-client#545

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -401,10 +401,15 @@ def create_workflow(user):  # noqa
         workflow_dict["operational_options"] = validate_operational_options(
             workflow_engine, reana_spec_file.get("inputs", {}).get("options", {})
         )
+        workspace_root_path = reana_spec_file.get("workspace", {}).get(
+            "root_path", None
+        )
         if git_data:
             workflow_dict["git_data"] = git_data
         response, http_response = current_rwc_api_client.api.create_workflow(
-            workflow=workflow_dict, user=str(user.id_)
+            workflow=workflow_dict,
+            user=str(user.id_),
+            workspace_root_path=workspace_root_path,
         ).result()
         if git_data:
             workflow = _get_workflow_with_uuid_or_name(


reana_server/rest/workflows.py
```diff
@@ -401,10 +401,15 @@ def create_workflow(user):  # noqa
         workflow_dict["operational_options"] = validate_operational_options(
             workflow_engine, reana_spec_file.get("inputs", {}).get("options", {})
         )
+        workspace_root_path = reana_spec_file.get("workspace", {}).get(
+            "workspace_root_path", None
```
mvidalgarcia, line 405: Do we have to call it `workspace_root_path`? Isn't it a bit redundant having `workspace` at the top level?
IMO we could simplify it and use just `path` or `root_path`. What do you think?

```yaml
...
workspace:
  path: /eos/home-s/simko/myworkflows
...
```

BTW, is it definite to have `workspace` at the top level? I thought we were discussing it [here](https://github.com/reanahub/reana-client/issues/545).

PS. For `get()`, you don't need to specify `None` as it's the default value.

mvidalgarcia: cc/ @tiborsimko @audrium

marcdiazsan: I used `workspace_root_path` as it is not the full `workspace_path`, the full path would also have the `workflow_uuid`. So I agree that `root_path` would be a good simplification

audrium: Looks good, +1 for `workspace_path`.
Also, @marcdiazsan,  for `get()`, you don't need to specify `None` as @mvidalgarcia mentioned

marcdiazsan: Yes this one can be closed as is included into the other PR

audrium:
Closing since all the commits from this PR are included in [another one](https://github.com/reanahub/reana-server/pull/397)

PR #396, opened 2021-08-25
docker: add --no-cache-dir to pip install

Sinclert:
This PR disables the pip cache when installing requirements. This is a common practice when installing packages on Docker images, as it helps reduce images size. In the case of `reana-server`, the resulting image shrink from `791` MBs to `700` MBs, ~90MBs in total.

The same optimization can be applied to:
- [reana-env-jupyter](https://github.com/reanahub/reana-env-jupyter/blob/master/Dockerfile#L16-L26).
- [reana-job-controller](https://github.com/reanahub/reana-job-controller/blob/master/Dockerfile#L68).
- [reana-workflow-controller](https://github.com/reanahub/reana-workflow-controller/blob/master/Dockerfile#L21).
- [reana-workflow-engine-cwl](https://github.com/reanahub/reana-workflow-engine-cwl/blob/master/Dockerfile#L21).
- [reana-workflow-engine-serial](https://github.com/reanahub/reana-workflow-engine-serial/blob/master/Dockerfile#L18).
- [reana-workflow-engine-snakemake](https://github.com/reanahub/reana-workflow-engine-snakemake/blob/master/Dockerfile#L18).
- [reana-workflow-engine-yadage](https://github.com/reanahub/reana-workflow-engine-yadage/blob/master/Dockerfile#L31).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN apt-get update && \
 
 # Install dependencies
 COPY requirements.txt /code/
-RUN pip install -r /code/requirements.txt
+RUN pip install --no-cache-dir -r /code/requirements.txt
 
 # Copy cluster component source code
 WORKDIR /code


tiborsimko:
Hi @Sinclert, this change would slow down development, as the dependencies would have to be constantly fetched and refetched... What about doing this only when releasing tagged images? IOW only when some Docker build parameter is set, as for `ARG DEBIG=0` business that we already have?  Can you introduce such an `if` handling?  Could be for the `pip` command, could be also alternatively removing cache directory at the end of the Dockefile together with other such possible cleaning...

Sinclert:
Hi @tiborsimko ,

As far as I understand, _most_ packages would only need to be fetched once (on the `pip install -r code/requirements.txt`). They would only need to be fetched again if the [setup.py high level requirements](https://github.com/reanahub/reana-server/blob/master/setup.py#L48-L80) are not met (on the `pip install .` / `pip install -e .[debug]`).

In principle, the _already installed_ packages should meet the `setup.py` specified ones, as both lists seems to be updating in parallel upon any dependency version upgrade. Only those that, for some reason, are specified on [setup.py](https://github.com/reanahub/reana-server/blob/master/setup.py#L48-L80) and not on [requirements.txt](https://github.com/reanahub/reana-server/blob/master/requirements.txt) would be re-fetched.

---

Regarding cleaning the _pip cache_:

Docker image sizes are determined by _operation_ layers (`RUN`, `ENV`, `COPY`...). If an operation creates a layer of size N, there is no way to reduce that size by a future operation that cleans up that layer files. In short: the _pip cache_ must be cleaned up on the same `pip install` operation to achieve the _image size reduction_ improvement.

Indeed, I have tried sticking a `RUN pip cache purge` line at the end of the `Dockerfile`, but the built image remains the same: `791` MBs.

Sinclert:
I could re-write the line as:

```Dockerfile
RUN if [ "${DEBUG}" -gt 0 ]; then pip install -r /code/requirements.txt; else pip install --no-cache-dir -r /code/requirements.tx; fi;
```

And move it after the `ARG DEBUG=0` line, if you think that would be helpful.

tiborsimko:
You are right, since we are updating both frequently together, the refetching should not be an issue... The early `pip` call would be cached by Docker layer system anyway.  Also good call on the _in situ_ cache purging rather than attempting to do it at the end of the file which would not be helpful. So I'm OK with the change here and in the other components :+1:

tiborsimko:
> I could re-write the line as...

Actually we should be good "as is", I think, so rewrite is not necessary!

Sinclert:
Great! 😄

I will create similar PRs on the repos that contain similar `Dockerfiles`, and comment on each image size reduction by the end of the day.

tiborsimko (approved):

PR #406, opened 2021-10-19
snakemake: filter out repeated job dependencies

mvidalgarcia:
A job might have multiple dependencies that can lead to a misinterpretation
of how many jobs can run in parallel. Imagine an scenario where we have jobs
A, B and C.

- A depends on no job
- B depends on A
- C depends on A and B

With the previous logic, the result would be that 2 jobs can run in parallel
as it's the maximum number of jobs that depend on another (C depends on 2).
But since B also depends on A, A and B cannot run in parallel.

So we need to filter out subdependencies to have an accurate picture of how
many jobs can run in parallel. In this particular example we need to filter out
the A dependency from C, as it's already a dependency from B.

closes reanahub/reana-commons#309

### DIFF
--- a/reana_server/complexity.py
+++ b/reana_server/complexity.py
@@ -8,7 +8,7 @@
 
 """REANA workflow complexity estimation."""
 
-from typing import List, Tuple
+from typing import Dict, List, Tuple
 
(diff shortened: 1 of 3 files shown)


audrium (approved):
Works as expected :rocket:

PR #416, opened 2021-11-11
scheduler: set correct logging level

audrium:
closes reanahub#415

This issue appeared because we were `logging` inside [configs](https://github.com/reanahub/reana-server/blob/master/reana_server/config.py#L151) before setting the log level. This resulted in `WARNING` level set to default which was not overridden by `logging.basicConfig(level=REANA_LOG_LEVEL, format=REANA_LOG_FORMAT)` later on.
This PR sets a logging level with force to override it.

### DIFF
--- a/reana_server/cli.py
+++ b/reana_server/cli.py
@@ -19,7 +19,7 @@ from reana_server.scheduler import WorkflowExecutionScheduler
 @click.command("start-scheduler")
 def start_scheduler():
     """Start a workflow execution scheduler process."""
-    logging.basicConfig(level=REANA_LOG_LEVEL, format=REANA_LOG_FORMAT)
+    logging.basicConfig(level=REANA_LOG_LEVEL, format=REANA_LOG_FORMAT, force=True)
     scheduler = WorkflowExecutionScheduler()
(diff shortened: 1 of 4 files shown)


reana_server/cli.py
```diff
@@ -11,15 +11,13 @@
...
 import click
-from reana_commons.config import REANA_LOG_FORMAT, REANA_LOG_LEVEL
 
 from reana_server.scheduler import WorkflowExecutionScheduler
 
 
 @click.command("start-scheduler")
 def start_scheduler():
     """Start a workflow execution scheduler process."""
-    logging.basicConfig(level=REANA_LOG_LEVEL, format=REANA_LOG_FORMAT)
```
VMois, line 22: I think adding `force=True` should fix it. [python 3.8 logging docs](https://docs.python.org/3.8/library/logging.html#logging.basicConfig)

Example:

`logging.basicConfig(level=REANA_LOG_LEVEL, format=REANA_LOG_FORMAT, force=True)`

audrium: Good catch! Works as expected with less changes

VMois (approved):

PR #419, opened 2021-11-19
rest: improve cluster health status calculation

audrium:
closes https://github.com/reanahub/reana-ui/issues/206

### DIFF
--- a/reana_server/status.py
+++ b/reana_server/status.py
@@ -21,6 +21,7 @@ from reana_commons.config import (
     REANA_COMPUTE_BACKENDS,
     REANA_INFRASTRUCTURE_KUBERNETES_NAMESPACE,
     REANA_RUNTIME_KUBERNETES_NAMESPACE,
+    REANA_MAX_CONCURRENT_BATCH_WORKFLOWS,
     SHARED_VOLUME_PATH,
 )
 from reana_commons.job_utils import kubernetes_memory_to_bytes
@@ -37,12 +38,13 @@ from reana_db.models import (
     ResourceType,
     ResourceUnit,
     RunStatus,
-    User,
     UserResource,
     Workflow,
 )
-from reana_server.utils import get_usage_percentage
 from sqlalchemy import desc
+
+from reana_server.config import REANA_KUBERNETES_JOBS_MEMORY_LIMIT_IN_BYTES
+from reana_server.utils import get_usage_percentage
 
 
 class REANAStatus:
@@ -579,6 +581,17 @@ class JobsStatus(REANAStatus):
         }
         return job_statuses
 
+    def get_total_slots(self):
+        """Get total amount of job slots available in REANA cluster."""
+        total_cluster_memory = NodesStatus().get_total_memory()
+        jobs_memory_limit = REANA_KUBERNETES_JOBS_MEMORY_LIMIT_IN_BYTES
+        slots = (
+            round(total_cluster_memory / jobs_memory_limit)
+            if total_cluster_memory and jobs_memory_limit
+            else None
+        )
+        return slots
+
 
 class ClusterHealthStatus(enum.Enum):
     """Enumeration of cluster health statuses."""
@@ -600,19 +613,26 @@ class ClusterHealth:
 
     @staticmethod
     def get_health_status(percentage):
-        """Calculate quota health status."""
+        """Calculate quota health status based on cluster availability."""
         health = ClusterHealthStatus.healthy
-        if percentage >= 50:
-            if percentage >= 75:
-                health = ClusterHealthStatus.critical
-            else:
-                health = ClusterHealthStatus.warning
+        if percentage <= 50:
+            health = ClusterHealthStatus.warning
+        if percentage <= 25:
+            health = ClusterHealthStatus.critical
+
         return health.name
 
     @staticmethod
-    def get_percentage(amount, total):
-        """Get percentage value based on ``amount`` and ``total``."""
-        return round((amount / total if total else 0) * 100)
+    def get_percentage(used, total):
+        """Get usage percentage based on ``used`` and ``total``."""
+        percentage = 100 - round((used / total if total else 0) * 100)
+        return percentage if percentage > 0 else 0
+
+    @staticmethod
+    def get_available(unavailable, total):
+        """Get available cluster resources based on ``unavailable`` and ``total``."""
+        available = total - unavailable
+        return available if available > 0 else 0
 
     def get_node_health(self):
         """Get cluster nodes health information."""
@@ -620,15 +640,15 @@ class ClusterHealth:
 
         total = len(nodes_status_obj.get_nodes())
         unschedulable = len(nodes_status_obj.get_unschedulable_nodes())
+        available = ClusterHealth.get_available(unschedulable, total)
         percentage = ClusterHealth.get_percentage(unschedulable, total)
 
         return {
-            "available": total - unschedulable,
+            "available": available,
             "unschedulable": unschedulable,
+            "total": total,
             "percentage": percentage,
-            "health": ClusterHealthStatus.critical.name
-            if unschedulable > 0
-            else ClusterHealthStatus.healthy.name,
+            "health": ClusterHealth.get_health_status(percentage),
             "sort": 0,
         }
 
@@ -638,12 +658,16 @@ class ClusterHealth:
 
         running = len(jobs_status_obj.get_k8s_jobs_by_status("Running"))
         pending = len(jobs_status_obj.get_k8s_jobs_by_status("Pending"))
-        total = running + pending
-        percentage = ClusterHealth.get_percentage(pending, total)
+        used = running + pending
+        total = jobs_status_obj.get_total_slots() or used
+        available = ClusterHealth.get_available(used, total)
+        percentage = ClusterHealth.get_percentage(used, total)
 
         return {
             "running": running,
             "pending": pending,
+            "available": available,
+            "total": total,
             "percentage": percentage,
             "health": ClusterHealth.get_health_status(percentage),
             "sort": 1,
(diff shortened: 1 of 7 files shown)


reana_server/status.py
```diff
@@ -599,11 +600,11 @@ def __init__(self):
         self.session = self.get_session_health()
 
     @staticmethod
-    def get_health_status(percentage):
+    def get_health_status(percentage, total):
         """Calculate quota health status."""
         health = ClusterHealthStatus.healthy
         if percentage >= 50:
-            if percentage >= 75:
+            if percentage >= 75 and total > REANA_MAX_CONCURRENT_BATCH_WORKFLOWS:
```
mvidalgarcia, line 607: `get_health_status` is a very generic method that ignores if we're calculating the Workflow or the Job health.

If I'm not wrong, [here](https://github.com/reanahub/reana-server/pull/419/files#diff-c23f1238a97f03b453a1be48238f557dcc0ce18c1ee6ba8992297d149ff14f19R649) we're passing the total number of jobs so we'd be comparing jobs with workflows, which might cause unexpected results.

audrium: You are right, it's not nice to compare jobs with workflows. Idea was to add some buffer for the `critical` status. If by default we have `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` set to 30, only if there are more jobs launched and most of them are pending (>=75%), only in this case it will be reported as `critical`. This is to prevent 2 pending jobs out of 3 to be `critical`.. I will try to come up with something next week

tiborsimko:
How does the new PR look with the folowing situation:

2021-11-29-104108_302x184_scrot

2021-11-29-112707_249x161_scrot

Seeing the job graph, I whether we shoudn't use a different colour to indicate what is "available" and what is "actually used", because the full green circle and the number of 0% may not be telling the exact story to the readers. Example: if we have 30 job nodes free and 4 slots per node (as governed by the memory limit of 3 Gi), then 4 running jobs would be occupying 4 out of 120 available slots, as it were, so one would expect the graph to indicate peraps "3%" of slots used. 

We may consider to indicate the number of "available slots" with a neutral colour as a background, e.g. for batch pods we have 40 slots available in gray, then put 1 running in green, 1 pending in orange, etc. The same for jobs, we have 120 slots available in gray, 2 running in green, the rest remain in gray.  However, such a graph could be problematic if we push more than the number of available slots onto the system, in a "going beyond 100%" kind of situations. Just jotting down some wild thoughts "live", we can come to these considerations later. 

Just wondering whether improvements for displaying "workflow slots" could also help to improve displaying "job slots".

reana_server/status.py
```diff
@@ -602,33 +604,40 @@ def __init__(self):
...
+        available = total - used
+        return available if available > 0 else 0
 
     def get_node_health(self):
         """Get cluster nodes health information."""
         nodes_status_obj = NodesStatus()
 
         total = len(nodes_status_obj.get_nodes())
         unschedulable = len(nodes_status_obj.get_unschedulable_nodes())
+        available = ClusterHealth.get_available(unschedulable, total)
```
mvidalgarcia, line 632: This might be a bit confusing as we're passing `unschedulable` as `used` arg, which I don't think is the same concept. In fact, I'm not sure if it's possible to get the `used` nodes.

reana_server/status.py
```diff
@@ -602,33 +604,40 @@ def __init__(self):
...
         return {
-            "available": total - unschedulable,
+            "available": available,
             "unschedulable": unschedulable,
+            "total": total,
             "percentage": percentage,
-            "health": ClusterHealthStatus.critical.name
-            if unschedulable > 0
-            else ClusterHealthStatus.healthy.name,
+            "health": ClusterHealth.get_health_status(percentage),
```
mvidalgarcia, line 640: I think that, if there's a single node that's `unschedulable`, it should be considered a critical situation. Under normal conditions, there shouldn't be any node that is `unschedulable`. That's probably why we kept this logic a bit different.

Imagine a situation where we have 20 nodes and 3 are malfunctioning, thus `unschedulable`. The health would still be `healthy` but actually, the cluster would need rapid attention.

reana_server/status.py
```diff
@@ -638,12 +647,22 @@ def get_job_health(self):
...
-        total = running + pending
-        percentage = ClusterHealth.get_percentage(pending, total)
+        used = running + pending
+        total_cluster_memory = NodesStatus().get_total_memory()
+        jobs_memory_limit = REANA_KUBERNETES_JOBS_MEMORY_LIMIT_IN_BYTES
+        total = (
+            round(total_cluster_memory / jobs_memory_limit)
+            if total_cluster_memory and jobs_memory_limit
+            else used
+        )
```
mvidalgarcia, line 657: I think this logic to calculate "job slots" can be moved to `JobsStatus`.

Some discussion is necessary to see how to handle situations when we run jobs that have k8s limit < `REANA_KUBERNETES_JOBS_MEMORY_LIMIT_IN_BYTES`, as in that case we might end up with more `used` than `total`. cc/ @tiborsimko

reana_server/status.py
```diff
@@ -656,13 +675,17 @@ def get_workflow_health(self):
...
-        total = running + queued + pending
-        percentage = ClusterHealth.get_percentage(queued + pending, total)
+        total = REANA_MAX_CONCURRENT_BATCH_WORKFLOWS
+        used = running + pending
+        available = ClusterHealth.get_available(used, total)
+        percentage = ClusterHealth.get_percentage(used, total)
 
         return {
             "running": running,
             "queued": queued,
```
mvidalgarcia, line 685: Just thinking out loud, since `queued` workflows are not represented in the pie chart anymore, should we display them in a more prominent way?

mvidalgarcia (commented):
We need to update the [OpenAPI spec docstring](https://github.com/reanahub/reana-server/blob/0b4229726fb0a354378970350cda7049f05c2b94/reana_server/rest/status.py#L27) (response and examples)

reana_server/status.py
audrium: Thanks for bringing this question. It's true that `unschedulable` node should be very alarming, but I thought for regular users it could send a bit too strong message about our cluster health and node availability. 
In production we have ~30 nodes. Here is how two different approaches looks like:
Screenshot from 2021-12-07 14-36-10
Screenshot from 2021-12-07 14-37-22

I agree that it's critical, just curious if we really want to show it to the end users. cc @tiborsimko

reana_server/status.py
audrium: Thanks for a nice suggestion, moved it to the `JobsStatus`.

regarding custom k8s limit set by users I guess there won't be that much of a problem, since if our calculated slot number would be exceeded, it would simply be shown that 0 slots are available as in this example (8 total slots):
Screenshot from 2021-12-07 15-48-07

reana_server/status.py
audrium: Good question. Here is how it looks like now:
Screenshot from 2021-12-07 15-58-44

Currently 10 queued workflows are not displayed in the chart and are not included in the available workflows and percentage calculation because of a bit different nature. So indeed it would be nice to display it in a more prominent way. I thought about using `italic` or `bold`, but it does not look great.. So I'm not sure, perhaps it's ok like this? Do you have any ideas? cc @tiborsimko

reana_server/status.py
audrium: Indeed this chart is a bit different, it shows live nodes instead of used ones. I guess it should be a bit problematic to show how busy are the nodes in a current moment.. That's why we use `unschedulable` as `used` arg. Are you proposing to change `used` arg name in `get_available()`?

reana_server/status.py
tiborsimko: I agree that we can display the "node slots" in red inside the circle. The plot also provides a message for us, the cluster admins, to get up and fix the nodes that have problems. 

As for the overall health badge having the number of 97% being displayed in red, I see that it may be too harsh to end users. If we have a cluster of say 500 nodes, then it is very well possible that some will be out "frequently" due to HW failures etc. The proability increases as the node count increases. Yet we would show 97% in red colour even if there are 485 nodes ready to server any heavy user workloads. That would be unnecessarily too pessimistic. We should accept that some nodes may have failures "naturally" in such a large cluster.

I guess we could use some logic such as above 90% the health badge is green, between 70% and 90% there are various shades of orange, and below 70% are various shades of red. Kind of like a battery gauge. Just an example.

reana_server/status.py
tiborsimko: We can be a bit approximative here. If our default job slot is 3 Gi, and users ask for 6Gi, this would eat 2 slots. But if they ask for 4 Gi, we cannot use integer numbers anymore. This plot is basically just an approximation of what happens with "default" job slot settings. When users ask for "more" or "less", perhaps the graph is not fully exact, but I guess we can accept some approximations here. 

One problem is how these "10 available" could be interpreted by the users. If we show absolute numbers, i.e. instead of 10 jobs available we show 30 Gi for jobs available, then things might be clearer. In this way, if a workflow only needs 1Gi job instead of the default 3Gi job, each user could calculate how much is "consumable" for their special needs. Ditto when they need 9 Gi for a job. So OT1H showing absolute numbers with units included might be actually preferable. But OTOH this could mean changing graph logic quite some... Perhaps a simple way could be to display "units = 3 Gi" somewhere near legend.

reana_server/status.py
tiborsimko: The "queued" workflow information is actually rather imporant information for end users. Imagine we have 30 slots, and somebody (User A) submits 100 workflows.  Those 70 in-excess workflows can stay in the queue for a long time. They won't be running, they won't be pending, just queued.  Now another User B could estimate wow, the cluster is busy, I won't submit my workflows now.  (Which could be paradoxically not good, since our "balanced" workflow scheduling policy would give priority to User B!  But it's hard to express this on a simple one-plot-for-all graph.)

As a simple solution, in the above graph, I would simply print "10 queued" or bold and red colour.  But perhaps we should include them in the cirle plot.  I.e. 30 slots are available, but if circle plot has 100 things to print, everything above 30 will be painted red.  This will make the health status to be direcrly visible out-of-the-box.

reana_server/status.py
mvidalgarcia: OK, let's leave it as it is then.

reana_server/status.py
mvidalgarcia: > Are you proposing to change `used` arg name in `get_available()`?

Yes, that was my idea.. Making the function a little bit more generic perhaps.

reana_server/status.py
mvidalgarcia: Leaving it as it is as it represents an approximation. We can revisit it in the future depending on the real user's experience.

reana_server/status.py
mvidalgarcia: Queued workflows are now displayed in a more prominent way (bold and [`queued` color](https://github.com/reanahub/reana-ui/blob/3b25cc4d6b971956b8f62e3d0bdb44f6b9e15f9e/reana-ui/src/util.js#L28)) when `queued` > 0. Changes made in [the UI PR](https://github.com/reanahub/reana-ui/pull/209).

reana_server/status.py
mvidalgarcia: Change to `unavailable` instead of `used` and it covers both `used` and `unschedulable`.

mvidalgarcia (approved):

PR #421, opened 2021-11-24
reana-admin: use resource type while setting quota limit

audrium:
`quota-set` command in `reana_admin` was using resource _name_ for specifying the resource. Currently we have "processing time" as a `cpu` name and "shared storage" as a `disk` name. So one had to do something like this in order to set some quota limits for a specific user:
```
kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -r "shared storage" -e audrius.mecionis@gmail.com -l 200 --admin-access-token $REANA_ACCESS_TOKEN 
```
This PR fixes `quota-set`, `quota-resources`, `quota-set-default-limits` commands to use resource types (e.g. disk, cpu) instead of a resource names ("shared storage", "processing time"). In case of same-type-multi-resource setup scenario it will ask for `--resource-name` to specify the exact resource.

To test:
```
$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-resources --admin-access-token $REANA_ACCESS_TOKEN
$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set-default-limits --admin-access-token $REANA_ACCESS_TOKEN 
$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -r cpu -e audrius.mecionis@gmail.com -l 200 --admin-access-token $REANA_ACCESS_TOKEN
$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -r disk -e audrius.mecionis@gmail.com -l 200 --admin-access-token $REANA_ACCESS_TOKEN
```

Snippet to add another `cpu` resource:
```
$ kubectl exec -i -t deployment/reana-server -c rest-api -- invenio shell

from reana_db.models import Resource, ResourceType, ResourceUnit
from reana_db.database import Session
r = Resource(name="test", type_=ResourceType.cpu, unit=ResourceUnit.milliseconds, title="test cpu")
Session.add(r)
Session.commit()
```
Then one has to do:
`$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -r cpu --resource-name "test" ...`

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -145,25 +145,8 @@ def _is_valid_rate_limit(rate_limit: str) -> bool:
 
 
 def _get_rate_limit(env_variable: str, default: str) -> str:
-    env_value = os.getenv(env_variable)
-
-    if not env_value:
-        logging.warning(
-            f"API rate limit config {env_variable} environment variable is not present. "
-            f"Using default value: {default}"
-        )
-        # assuming default has correct format
-        return default
-
-    if _is_valid_rate_limit(env_value):
-        return env_value
-    else:
-        logging.warning(
-            f"API rate limit config {env_variable} environment variable has incorrect format. "
-            f"Incorrect value: {env_value}. Using default value: {default}"
-        )
-        # assuming default has correct format
-        return default
+    env_value = os.getenv(env_variable, "")
+    return env_value if _is_valid_rate_limit(env_value) else default
 
 
 # Note: users that are connecting via reana-client will be treated as guests by the Invenio framework
--- a/reana_server/reana_admin.py
+++ b/reana_server/reana_admin.py
@@ -28,6 +28,7 @@ from reana_db.models import (
     AuditLogAction,
     QuotaHealth,
     Resource,
+    ResourceType,
     User,
     UserResource,
     UserTokenStatus,
@@ -503,12 +504,12 @@ def list_quota_usage(
 
 
 @reana_admin.command("quota-resources", help="List available quota resources.")
-@admin_access_token_option
 @click.pass_context
-def list_quota_resources(ctx, admin_access_token):
+def list_quota_resources(ctx):
     """List quota resources."""
-    resource_names = [resource.name for resource in Resource.query]
-    click.echo("\n".join(resource_names))
+    click.echo("Available resources are:")
+    for resource in Resource.query:
+        click.echo(f"{resource.type_.name} ({resource.name})")
 
 
 @reana_admin.command(
@@ -519,31 +520,56 @@ def list_quota_resources(ctx, admin_access_token):
     "--email",
     "emails",
     multiple=True,
+    required=True,
     help=(
         "The emails of the users. "
         "E.g. --email johndoe@example.org --email janedoe@example.org"
     ),
 )
-@click.option("--resource-name", "-r", help="Name of resource.", required=True)
+@click.option(
+    "--resource", "-r", "resource_type", help="Specify quota resource. e.g. cpu, disk."
+)
+@click.option("--resource-name", "-n", help="Name of resource.")
 @click.option(
     "--limit", "-l", help="New limit in canonical unit.", required=True, type=int
 )
-@admin_access_token_option
 @click.pass_context
-def set_quota_limit(ctx, admin_access_token, emails, resource_name, limit):
+def set_quota_limit(ctx, emails, resource_type, resource_name, limit):
     """Set quota limits to the given users per resource."""
     try:
         for email in emails:
             error_msg = None
+            resource = None
             user = _get_user_by_criteria(None, email)
-            resource = Resource.query.filter_by(name=resource_name).one_or_none()
+
+            if resource_name:
+                resource = Resource.query.filter_by(name=resource_name).one_or_none()
+            elif resource_type in ResourceType._member_names_:
+                resources = Resource.query.filter_by(type_=resource_type).all()
+                if resources and len(resources) > 1:
+                    click.secho(
+                        f"ERROR: There are more than one `{resource_type}` resource. "
+                        "Please provide resource name with `--resource-name` option to specify the exact resource.",
+                        fg="red",
+                        err=True,
+                    )
+                    sys.exit(1)
+                else:
+                    resource = resources[0]
+
             if not user:
                 error_msg = f"ERROR: Provided user {email} does not exist."
             elif not resource:
+                resources = [
+                    f"{resource.type_.name} ({resource.name})"
+                    for resource in Resource.query
+                ]
                 error_msg = (
-                    "ERROR: Provided resource name does not exist. Available "
-                    f"resources are {[resource.name for resource in Resource.query]}."
+                    f"ERROR: Provided resource `{resource_name or resource_type}` does not exist. "
+                    if resource_name or resource_type
+                    else "ERROR: Please provide a resource. "
                 )
+                error_msg += f"Available resources are: {', '.join(resources)}."
             if error_msg:
                 click.secho(
                     error_msg, fg="red", err=True,
@@ -568,7 +594,7 @@ def set_quota_limit(ctx, admin_access_token, emails, resource_name, limit):
                 )
         Session.commit()
         click.secho(
-            f"Quota limit {limit} for '{resource.name}' successfully set to users {emails}.",
+            f"Quota limit {limit} for '{resource.type_.name} ({resource.name})' successfully set to users {emails}.",
             fg="green",
         )
     except Exception as e:
@@ -584,16 +610,18 @@ def set_quota_limit(ctx, admin_access_token, emails, resource_name, limit):
     "quota-set-default-limits",
     help="Set default quota limits to users that do not have any.",
 )
-@admin_access_token_option
 @click.pass_context
-def set_default_quota_limit(ctx, admin_access_token):
+def set_default_quota_limit(ctx):
     """Set default quota limits to users that do not have any."""
     users_without_quota_limits = User.query.filter(~User.resources.any()).all()
+    if not users_without_quota_limits:
+        click.secho("There are no users without quota limits.", fg="green")
+        sys.exit(0)
     for resource in Resource.query:
         ctx.invoke(
             set_quota_limit,
-            admin_access_token=admin_access_token,
             emails=[user.email for user in users_without_quota_limits],
             resource_name=resource.name,
+            resource_type=resource.type_.name,
             limit=DEFAULT_QUOTA_LIMITS.get(resource.type_.name),
         )
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 Changes
 =======
 
+Version 0.8.1 (2021-11-29)
+--------------------------
+
+- Changes ``quota-set`` command used by the REANA administrators to use the resource type along with a resource name for specifying the resource.
+- Changes email validation used in ``create-admin-user`` command by the REANA administrators to be more permissive.
+
 Version 0.8.0 (2021-11-22)
 ---------------------------
 
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -3,7 +3,7 @@
   "info": {
     "description": "Submit workflows to be run on REANA Cloud",
     "title": "REANA Server",
-    "version": "0.8.0"
+    "version": "0.8.1"
   },
   "parameters": {},
   "paths": {
@@ -2599,7 +2599,7 @@
                     }
                   }
                 },
-                "reana_server_version": "0.8.0",
+                "reana_server_version": "0.8.1",
                 "reana_token": {
                   "requested_at": "Mon, 25 May 2020 10:39:57 GMT",
                   "status": "active",
--- a/reana_server/rest/users.py
+++ b/reana_server/rest/users.py
@@ -112,7 +112,7 @@ def get_you(user):
             application/json:
               {
                 "email": "user@reana.info",
-                "reana_server_version": "0.8.0",
+                "reana_server_version": "0.8.1",
                 "reana_token": {
                     "value": "Drmhze6EPcv0fN_81Bj-nA",
                     "status": "active",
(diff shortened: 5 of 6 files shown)


reana_server/reana_admin.py
```diff
@@ -589,11 +592,14 @@ def set_quota_limit(ctx, admin_access_token, emails, resource_name, limit):
 def set_default_quota_limit(ctx, admin_access_token):
     """Set default quota limits to users that do not have any."""
     users_without_quota_limits = User.query.filter(~User.resources.any()).all()
+    if not users_without_quota_limits:
+        click.secho("There are no users without quota limits.", fg="green")
+        sys.exit(1)
```
mvidalgarcia, line 597: Question: I guess it was failing because the function wasn't idempotent?

audrium: Not really, when I tested it locally there were no users without quota limits, so `users_without_quota_limits = []` and this was causing some errors while invoking `set_quota_limit`

reana_server/reana_admin.py
```diff
@@ -507,8 +508,9 @@ def list_quota_usage(
 @click.pass_context
 def list_quota_resources(ctx, admin_access_token):
     """List quota resources."""
-    resource_names = [resource.name for resource in Resource.query]
-    click.echo("\n".join(resource_names))
+    click.echo("Available resources are:")
+    for resource in Resource.query:
+        click.echo(f"{resource.type_.name} ({resource.name})")
```
mvidalgarcia, line 513: Shouldn't it be the other way around? First the resource name and then resource type in parenthesis. In the end, the unique value here is the resource name.

reana_server/reana_admin.py
```diff
@@ -524,25 +526,50 @@ def list_quota_resources(ctx, admin_access_token):
         "E.g. --email johndoe@example.org --email janedoe@example.org"
     ),
 )
-@click.option("--resource-name", "-r", help="Name of resource.", required=True)
+@click.option(
+    "--resource", "-r", help="Specify quota resource. e.g. cpu, disk.", required=True
+)
```
mvidalgarcia, line 531: Should this be required? If one wants to set a limit to a single resource name, how should they proceed?

reana_server/reana_admin.py
```diff
@@ -524,25 +526,50 @@ def list_quota_resources(ctx, admin_access_token):
...
+@click.option("--resource-name", help="Name of resource.")
```
mvidalgarcia, line 532: Perhaps we can allow `-n` here, what do you think?

reana_server/reana_admin.py
```diff
@@ -524,25 +526,50 @@ def list_quota_resources(ctx, admin_access_token):
...
     try:
         for email in emails:
             error_msg = None
             user = _get_user_by_criteria(None, email)
-            resource = Resource.query.filter_by(name=resource_name).one_or_none()
+            try:
+                resource = Resource.query.filter(
+                    Resource.type_ == resource
+                ).one_or_none()
+            except MultipleResultsFound:
```
mvidalgarcia, line 548: What about checking `count() > 1` instead of risking an exception raise? I think it'd look cleaner and we would avoid the try-except block, session rollback, etc.

mvidalgarcia (commented):
Haven't tested it yet. Some things I've found by just looking at the code.

reana_server/reana_admin.py
audrium: No, the idea is that on one-type-one-resource scenario one can only use the type for specifying the limits, etc. So it makes sense to have the type of a resource as a main value

reana_server/reana_admin.py
audrium: There are some issues while querying a `Resource` by `type_` which is enum:
```
$ kubectl exec -i -t deployment/reana-server -c rest-api -- invenio shell
from reana_db.models import Resource
Resource.query.filter(Resource.type_ == "some resource").first()
# Gives: DataError: (psycopg2.errors.InvalidTextRepresentation) invalid input value for enum resourcetype: "some resource"
```
So try-except block would still be needed.
Also `resource` value is used later in the code.. with `count() > 1` we would need two queries

reana_server/reana_admin.py
audrium: I think yes, if one wants to set a limit to a resource:

- on one-type-one-resource scenario:
```kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -r disk ...```
- on one-type-multi-resource scenario one has to additionally specify resource name:
```kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -r disk --resource-name "my disk" ...```

Comment (mvidalgarcia) [reply reana_server/reana_admin.py:531]: > - on one-type-multi-resource scenario one has to additionally specify resource name:
```kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -r disk --resource-name "my disk" ...```


The resource name is unique, so requiring to pass also the resource type is redundant.

File: reana_server/reana_admin.py (hunk 2 above)
Comment (mvidalgarcia) [reply reana_server/reana_admin.py:513]: For me this:
```
fast processing time (cpu)                                                                                                                                                                                                                                                               
shared storage (disk)                                                                                                                                                                                                                                                        
slow processing time (cpu)
```

looks better than this:

```
cpu (fast processing time)
disk (shared storage)                                                                                                                                                                                                                                               
cpu (slow processing time)
```

Furthermore, the command is called `quota-resources`, no `quota-resource-types`!

File: reana_server/reana_admin.py (hunk 5 above)
Comment (mvidalgarcia) [reply reana_server/reana_admin.py:548]: It can be done with something like this:

```python
In [0]: my_type = 'cpu'

In [1]: ResourceType._member_names_
Out[1]: ['cpu', 'disk']

In [2]: my_type in ResourceType._member_names_
Out[2]: True

In [3]: ResourceType[my_type]
Out[3]: <ResourceType.cpu: 0>

In [4]: Resource.query.filter_by(type_=ResourceType[my_type]).all()
Out[4]:
[<Resource 62bcabea-f073-456e-97a7-e274ec5c1e2b>,
 <Resource 1cb18814-a556-4c54-ab4f-5d04a4f22c8c>]
```

As far as I know count queries are very cheap. But if you prefer you can fetch all the results and use `len()` instead.

audrium: I came up with this peace of code:
```python

if resource in ResourceType._member_names_:
    resource_count = Resource.query.filter_by(type_=resource).count()
    if resource_count > 1:
        if resource_name:
            resource = Resource.query.filter_by(
                name=resource_name
            ).one_or_none()
        else:
            error_msg = (
                f"ERROR: There are more than one `{resource}` resource. "
                "Please provide resource name with `--resource-name` option to specify the exact resource."
            )
    else:
        resource = Resource.query.filter_by(type_=resource).one_or_none()
else:
    resource = None
```
which indeed avoids the try-except block and session rollback, but still looks messy IMO. Is this what you had in mind?

reana_server/reana_admin.py
audrium: Our primary use case right now is one-type-one-resource scenario, so second option for me makes more sense. We refer to the resource type as a `resource` in client and admin facing commands:
- reana-client quota-show --resource disk -h
- reana-admin quota-set --resource disk ...

reana_server/reana_admin.py
audrium: Agree that it's redundant. My idea was that on one-type-one-resource scenario user would see `resource` as a required field while doing `reana-admin quota-set --help` and it would signal that this is the option one has to use.. (since we should focus on this use case).
Also, supporting `resource-name` without `resource` uglifies the implementation a bit

reana_server/reana_admin.py
mvidalgarcia: Similar to that, but I think we don't 3 queries. I've been playing a little bit with the code. Here more or less is the diff of the idea I had in mind:

```diff
diff --git a/reana_server/reana_admin.py b/reana_server/reana_admin.py
index 6002de6..a093e65 100644
--- a/reana_server/reana_admin.py
+++ b/reana_server/reana_admin.py
@@ -28,6 +28,7 @@ from reana_db.models import (
     AuditLogAction,
     QuotaHealth,
     Resource,
+    ResourceType,
     User,
     UserResource,
     UserTokenStatus,
@@ -521,13 +522,14 @@ def list_quota_resources(ctx, admin_access_token):
     "--email",
     "emails",
     multiple=True,
+    required=True,
     help=(
         "The emails of the users. "
         "E.g. --email johndoe@example.org --email janedoe@example.org"
     ),
 )
 @click.option(
-    "--resource", "-r", help="Specify quota resource. e.g. cpu, disk.", required=True
+    "--resource", "-r", "resource_type", help="Specify quota resource. e.g. cpu, disk.",
 )
 @click.option("--resource-name", "-n", help="Name of resource.")
 @click.option(
@@ -535,30 +537,30 @@ def list_quota_resources(ctx, admin_access_token):
 )
 @admin_access_token_option
 @click.pass_context
-def set_quota_limit(ctx, admin_access_token, emails, resource, resource_name, limit):
+def set_quota_limit(
+    ctx, admin_access_token, emails, resource_type, resource_name, limit
+):
     """Set quota limits to the given users per resource."""
     try:
         for email in emails:
             error_msg = None
+            resource = None
             user = _get_user_by_criteria(None, email)
-            try:
-                resource = Resource.query.filter(
-                    Resource.type_ == resource
-                ).one_or_none()
-            except MultipleResultsFound:
-                # If case of multiple resources of the same type query by the resource_name
-                if resource_name:
-                    resource = Resource.query.filter_by(
-                        name=resource_name
-                    ).one_or_none()
-                else:
-                    error_msg = (
-                        f"ERROR: There are more than one `{resource}` resource. "
-                        "Please provide resource name with `--resource-name` option to specify the exact resource."
+
+            if resource_name:
+                resource = Resource.query.filter_by(name=resource_name).one_or_none()
+            elif resource_type in ResourceType._member_names_:
+                resources = Resource.query.filter_by(type_=resource_type).all()
+                if resources and len(resources) > 1:
+                    click.secho(
+                        f"ERROR: There are more than one `{resource_type}` resource. "
+                        "Please provide resource name with `--resource-name` option to specify the exact resource.",
+                        fg="red",
+                        err=True,
                     )
-            except Exception:
-                Session.rollback()
-                resource = None
+                    sys.exit(1)
+                else:
+                    resource = resources[0]
 
             if not user:
                 error_msg = f"ERROR: Provided user {email} does not exist."
@@ -568,7 +570,7 @@ def set_quota_limit(ctx, admin_access_token, emails, resource, resource_name, li
                     for resource in Resource.query
                 ]
                 error_msg = (
-                    "ERROR: Provided resource does not exist. Available "
+                    f"ERROR: Provided resource `{resource_name or resource_type}` does not exist. Available "
                     f"resources are: {', '.join(resources)}."
                 )
             if error_msg:

```

audrium: Looks like a nice solution, thanks!

reana_server/reana_admin.py
```diff
@@ -589,11 +619,15 @@ def set_quota_limit(ctx, admin_access_token, emails, resource_name, limit):
 def set_default_quota_limit(ctx, admin_access_token):
     """Set default quota limits to users that do not have any."""
     users_without_quota_limits = User.query.filter(~User.resources.any()).all()
+    if not users_without_quota_limits:
+        click.secho("There are no users without quota limits.", fg="green")
+        sys.exit(1)
```
mvidalgarcia, line 624: We might not need to do exit with error code 1 as this is not really an error but some information. What about `sys.exit(0)` instead?

reana_server/reana_admin.py
```diff
@@ -584,16 +610,18 @@ def set_quota_limit(ctx, admin_access_token, emails, resource_name, limit):
...
+    if not users_without_quota_limits:
+        click.secho("There are no users without quota limits.", fg="green")
+        sys.exit(0)
     for resource in Resource.query:
         ctx.invoke(
             set_quota_limit,
-            admin_access_token=admin_access_token,
             emails=[user.email for user in users_without_quota_limits],
             resource_name=resource.name,
+            resource=resource.type_.name,
```
mvidalgarcia, line 625: This should be `resource_type`, or we can just omit it and rely on `resource_name`.

reana_server/reana_admin.py
```diff
@@ -584,16 +610,18 @@ def set_quota_limit(ctx, admin_access_token, emails, resource_name, limit):
...
+    if not users_without_quota_limits:
+        click.secho("There are no users without quota limits.", fg="green")
+        sys.exit(0)
     for resource in Resource.query:
         ctx.invoke(
             set_quota_limit,
-            admin_access_token=admin_access_token,
             emails=[user.email for user in users_without_quota_limits],
             resource_name=resource.name,
+            resource_type=resource.type_.name,
```
mvidalgarcia, line 625: This can be omitted as the code, as it is now, will always check `resource_name` first and ignore `resource_type` but we can keep it just in case.

reana_server/config.py
```diff
@@ -145,25 +145,8 @@ def _is_valid_rate_limit(rate_limit: str) -> bool:
...
-        # assuming default has correct format
-        return default
-
-    if _is_valid_rate_limit(env_value):
-        return env_value
-    else:
-        logging.warning(
-            f"API rate limit config {env_variable} environment variable has incorrect format. "
-            f"Incorrect value: {env_value}. Using default value: {default}"
-        )
```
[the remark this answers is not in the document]
audrium: The problem with this warning is that it keeps popping after every `reana-admin` command use. Motivation of removing it can be found here: https://mattermost.web.cern.ch/it-dep/pl/gbcszsb7hfncpqx9kus478os8y

mvidalgarcia (approved):
All looking good 👍

reana_server/config.py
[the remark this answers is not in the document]
mvidalgarcia: Ah OK, I thought that the one that was popping all the time was `"API rate limit config {env_variable} environment variable is not present. "Using default value: {default}"`, not both.

PR #423, opened 2021-11-25
utils: email validation, handle additional cases

### LINKED ISSUE #422: Regex in _validate_email is very picky.
Email addresses with a dash fail with `ERROR: Invalid email format`, which is very confusing when running the create-admin-user script.

https://github.com/reanahub/reana-server/blob/4c986b2847f546ee3eaac3a1d752dd77a2ff4ed6/reana_server/utils.py#L435-L439

VMois:
I went with a minimal implementation just to cover the issue, but there are more cases that we do not support and we might consider it in this PR or open an issue for later. What do you think?

## Another option

I found a [website](http://emailregex.com) with a less strict email regex. In python:

```python
email_regex = r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)"
```

This regex covers cases like `john.doe@edu.uni.ch` or `john.doe.boe@example.org` (double dots) that are not covered currently. But it allows for emails like `john..doe@example.org` which are incorrect, etc.

*question(non-blocking):* @mvidalgarcia do you remember what was the goal for email validation? Like being more strict or catching common input mistakes.

closes #422

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,29 @@
+# This file is part of REANA.
+# Copyright (C) 2021 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA-Server tests for utils module."""
+
+import pytest
+from reana_server.utils import is_valid_email
+
+
+@pytest.mark.parametrize(
+    "email,is_valid",
+    [
+        ("john@example.org", True),
+        ("john.doe@example.org", True),
+        ("john-doe@example.org", True),
+        ("john.doe@edu.uni.org", True),
+        ("jean-yves.le.meur@cern.ch", True),
+        ("john.doe@exampleorg", False),
+        ("john.doeexample.org", False),
+        ("john@example.org.", False),
+        ("john@example..org", False),
+        ("john@@example.org", False),
+    ],
+)
+def test_is_email_valid(email: str, is_valid: bool):
+    assert is_valid_email(email) == is_valid
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -19,11 +19,13 @@ import os
 from uuid import UUID, uuid4
 
 import click
-import fs
 import requests
 import yaml
-from flask import current_app as app, url_for
+from flask import url_for
 from jinja2 import Environment, PackageLoader, select_autoescape
+from marshmallow.exceptions import ValidationError
+from marshmallow.validate import Email
+
 from reana_commons.config import REANAConfig, REANA_WORKFLOW_UMASK
 from reana_commons.email import send_email
 from reana_commons.k8s.secrets import REANAUserSecretsStore
@@ -36,7 +38,6 @@ from sqlalchemy.exc import (
     SQLAlchemyError,
     StatementError,
 )
-from werkzeug.wsgi import LimitedStream
 
 from reana_server.api_client import current_workflow_submission_publisher
 from reana_server.complexity import get_workflow_min_job_memory, estimate_complexity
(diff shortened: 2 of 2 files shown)


tests/test_utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+import pytest
+from reana_server.utils import is_valid_email
+
+
+@pytest.mark.parametrize(
+    "email,is_valid",
+    [
+        ("john.doe@example.org", True),
+        ("john-doe@example.org", True),
```
tiborsimko, line 17: A good test persona is `jean-yves.le.meur@cern.ch`. It should pass.

tests/test_utils.py
```diff
@@ -0,0 +1,23 @@
...
+"""REANA-Server tests for utils module."""
+
+import pytest
+from reana_server.utils import is_valid_email
+
+
+@pytest.mark.parametrize(
+    "email,is_valid",
+    [
+        ("john.doe@example.org", True),
```
tiborsimko, line 16: You may also put the simplest example `john@example.org` to make it explicit that it passes.

tests/test_utils.py
audrium: Would be nice to support also emails like `john.doe.poe@example.org` (which you found that it's not working). While testing it [here](https://regex101.com/r/S9bgjX/1) it seems it's not working currently

VMois: This will not pass rn. We only accept a single dot before `@`. More details in PR description. We have two options: (A) to modify the current regex to allow two (or more?) dots or (B) use regex I found online which is less strict (allows for some incorrect values) but covers a lot of cases.

What should be the way?

VMois: I am inclined towards a less strict regex but I don't know how important strictness is for us.

VMois: W3C email regex used in type=email HTML field ([details](https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email))) :) 

```python
email_regex = r"^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$"
```

tiborsimko: Yes, I'm also for having more permissive check, so that emails like that of Jean-Yves would pass.  (That was what I meant by the terse expression "it should pass"  there -- i.e. I have tested that one and it didn't. So it's nice to add it in a TDD style and then change the implementation to make it pass...)

mvidalgarcia:
> _question(non-blocking):_ @mvidalgarcia do you remember what was the goal for email validation? Like being more strict or catching common input mistakes.

To know that, you'd need to ask the [15-month-ago](https://github.com/reanahub/reana-server/commit/bac7c8ba16473e512bfeae4fae83f700cd6d5950) Marco 😅 

Jokes aside, I guess we were just trying to have a very simple validation. In the end, the [`email` is a String](https://github.com/reanahub/reana-db/blob/4c4cebb6afc1565fac90ef3ab7f85b8f4631cdd2/reana_db/models.py#L139) so any string could potentially go to the DB. As this command is only meant to be used by admins I guess we didn't put too much emphasis, as normal users register themselves via UI. It'd be nice to review that use case as well just in case, but I guess [there](https://github.com/reanahub/reana-ui/blob/ee704c5ad417867d05c4d188ed8878ba4966ab6e/reana-ui/src/pages/signin/components/SignForm.js#L47) we rely on the [HTML email standard validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email).

tests/test_utils.py
VMois: I changed to the regex that I posted in the PR description. Now tests are passing. The new one is more permissive but also allows for some incorrect values (e.g `john@example.org..`). I think it is good for now. In case admins will want to have a stricter check we can reiterate.

I didn't want to use the W3C regex as it is pretty huge and can be intimidating. It covers *some* edge cases (e.g `john@example.org..`) but we are fine without it.

tiborsimko: Well if it's battle-tested, I'd say let's go for the W3C one! Better to catch trailing dots...

tiborsimko: (e.g. the current one would accept happily `john@example....org` which is not good... while the W3C one is working better. Perhaps add all these double-dots into test suite so that we retain easily what we tested...)

VMois: Done. 

Interesting thing is that `john@exampleorg` is a valid email. HTML input (type=email) allows it.

tiborsimko: It is definitely nicer to use the Marshmallow one :+1:

VMois: Agree. It is already installed for `r-server`. I will change PR.

VMois: Done

PR #424, opened 2021-12-01
reana-admin: validate email on `user-create`

mvidalgarcia:


audrium (approved):
Works as expected!

mvidalgarcia:
> Can you please enrich `CHANGES.rst` accordingly?

Added, please check. I went for "Version 0.8.2 (UNRELEASED)" but I'm actually not sure.

Note that the changelog item is different from the ``create-admin-user`` one as in this case there was no validation at all, hence "Changes" -> "Adds".

tiborsimko:
> I went for "Version 0.8.2 (UNRELEASED)" but I'm actually not sure.

If we put this into 0.8.* release series, then the PR should go against `maint-0.8` branch.

If we put this into 0.9.* release series, then the PR can stay against `master` but we should change the `CHANGES.rst` entry to read 0.9.0.  

The UNRELEASED is perfect in either case.

I guess this PR can be considered a bug fix, so it is rather something for the 0.8.* release series and the `maint-0.8` branch.

(Yeah, we'll have some more of the `maint-0.8 -> master` merge fun ahead! Fortunately it should be simple until there are new big feature branches in `master`, which is not yet the case.)

PR #425, opened 2021-12-03
ext: enable logging in REANA Invenio app

### LINKED ISSUE #298: logger: basic configuration does not work because of the invenio-logging handlers
Currently logger level set in reana-commons does not work for reana-server since:

- `Invenio-logger` has it's own handlers and set logging level to `None` by default: [link](https://invenio-logging.readthedocs.io/en/latest/configuration.html#invenio_logging.config.LOGGING_CONSOLE_LEVEL)

- logger [basic config](https://github.com/reanahub/reana-server/blob/master/reana_server/factory.py#L29) where we set logger level does nothing if the logger already has handlers configured: [link to docs](https://docs.python.org/3/library/logging.html#logging.basicConfig)

### LINKED ISSUE #417: rest: logging does not work correctly in production mode
Currently on production mode installation there are some issues with `rest-api` logging:

1) It is only working for limited amount of endpoints. E.g. `/ping` is not being logged while `/api/you` logged as expected

2) There are lots of deprecation warnings in the logs like this:
```
<unknown>:1: DeprecationWarning: invalid escape sequence \[

2021-11-12 16:01:59,147 | py.warnings | uWSGIWorker2Core0 | WARNING | <unknown>:1: DeprecationWarning: invalid escape sequence \[
```
which are repeated for almost 1.5k lines after application starts. This could be due to outdated `invenio` dependencies. However it's not trivial to upgrade it because we have a conflicting `celery` dependency (from outdated `kombu` version which requires `celery<5.0.0` pinning).

3) Logging level set in reana-commons is not respected: https://github.com/reanahub/reana-server/issues/298

4) It seems that `factory.py` and `create_app()` are only being used in `reana-admin` and `scheduler` (defined [here](https://github.com/reanahub/reana-server/blob/master/setup.py#L103)). Needs a deeper investigation, but if that's the case there is no point in registering the blueprints there. Looks like [invenio-base](https://github.com/reanahub/reana-server/blob/master/setup.py#L107) is taking care for [starting up](https://github.com/reanahub/reana/blob/master/helm/reana/templates/reana-server.yaml#L42) the application. To reproduce try to:
    - comment/remove `factory.py` and `app.py`
    - `reana-dev docker-build -c . -b DEBUG=1`
    - `reana-dev kind-load-docker-image -c .`
    - kill the old pod and inspect the new one - works as expected
    - it does not work while redeploying the cluster because it can't start `flask reana-admin create-admin-user `

mvidalgarcia:
closes #417
closes #298

#### To test:

Two scenarios:
1. Debug mode (`invenio run`)
2. Production mode (`uwsgi`)

To switch between scenarios check https://github.com/reanahub/reana/pull/603

In **both** scenarios, perform these steps:
1. Leave reana-server pod logs open in a terminal
```console
$ kubectl logs deployment/reana-server rest-api -f
```
2. Navigate to the different endpoints via UI and CLI, e.g. UI: sign in, go to the status page, etc.; CLI: `reana-client ping/info/run..`
3. Verify that the logs correspond to the network calls
4. Go to a certain endpoint (e.g. [`/api/config`](https://github.com/reanahub/reana-server/blob/0b4229726fb0a354378970350cda7049f05c2b94/reana_server/rest/config.py#L22)) and add some log messages at the beginning of the function
```diff
diff --git a/reana_server/rest/config.py b/reana_server/rest/config.py
index 9353aad..014a916 100644
--- a/reana_server/rest/config.py
+++ b/reana_server/rest/config.py
@@ -65,6 +65,11 @@ def get_config():
                 "message": "Internal server error."
               }
     """
+    logging.debug("debug foo")
+    logging.error("error foo")
+    logging.info("info foo")
+    logging.critical("critical foo")
+    print("this is a print")
     try:
         return (
             jsonify(REANAConfig.load("ui")),

```
5. Call this endpoint and check logs.
6. Do the same but instead of printing a message, raise an Exception to verify that is caught properly and the traceback is printed.
```python
raise Exception("foo")
```

### DIFF
--- a/reana_server/app.py
+++ b/reana_server/app.py
@@ -6,7 +6,7 @@
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
-"""Main entrypoint for Reana-Server."""
+"""Main entrypoint for REANA-Server."""
 
 from reana_server.factory import create_app
 
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -9,7 +9,6 @@
 """Flask application configuration."""
 
 import copy
-import logging
 import os
 import re
 
--- a/reana_server/ext.py
+++ b/reana_server/ext.py
@@ -8,7 +8,10 @@
 
 """Flask extension REANA-Server."""
 
+import logging
+
 from flask_menu import Menu
+from reana_commons.config import REANA_LOG_FORMAT, REANA_LOG_LEVEL
 
 from reana_server import config
 
@@ -18,6 +21,9 @@ class REANA(object):
 
     def __init__(self, app=None):
         """Extension initialization."""
+        logging.basicConfig(level=REANA_LOG_LEVEL, format=REANA_LOG_FORMAT, force=True)
+        werkzeug_logger = logging.getLogger("werkzeug")
+        werkzeug_logger.propagate = False
         if app:
             self.app = app
             self.init_app(app)
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -14,20 +14,17 @@ import traceback
 import requests
 from bravado.exception import HTTPError
 from flask import Blueprint, Response
-from flask import current_app as app
 from flask import jsonify, request, stream_with_context
 from reana_commons.config import REANA_WORKFLOW_ENGINES
 from reana_commons.errors import REANAQuotaExceededError, REANAValidationError
 from reana_commons.operational_options import validate_operational_options
 from reana_commons.workspaces import validate_workspace
-from reana_db.database import Session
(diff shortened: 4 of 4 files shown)


audrium (approved):
Works nicely!

Have you tried investigating 1st and 4th items from the [issue](https://github.com/reanahub/reana-server/issues/417)? We should create separate issues to address it

Edit: created https://github.com/reanahub/reana-server/issues/427 and https://github.com/reanahub/reana-server/issues/426, feel free to post your findings there, if you did some research

mvidalgarcia:
> feel free to post your findings there, if you did some research

No more finding apart from what you've already described on the issues. I focused on the logs themselves as it was the critical point.

PR #428, opened 2021-12-21
rest: include `Content-Disposition` header when needed

mvidalgarcia:
* the `preview` query param must contain a `true` value to consider it True.
* when `Content-Disposition` header is set, even if its content is empty,
is not possible to preview eg. HTML directly on the server, thus we set this
header only when it's needed (downloads).

closes reanahub/reana-ui#190


audrium (approved):
Works nicely!

audrium (approved):
Works nicely!

Issue #437, opened 2022-02-08
scheduler: fix job memory requirement check when launching a workflow


tiborsimko:
**Current behaviour**

When a user asks for a lot of memory, for example for RooFit workflow:

```diff
diff --git a/workflow/yadage/workflow.yaml b/workflow/yadage/workflow.yaml
index 14bc7f7..ab64a46 100644
--- a/workflow/yadage/workflow.yaml
+++ b/workflow/yadage/workflow.yaml
@@ -34,7 +34,7 @@ stages:
           image: 'reanahub/reana-env-root6'
           imagetag: '6.18.04'
           resources:
-            - kubernetes_memory_limit: '256Mi'
+            - kubernetes_memory_limit: '256Gi'
   - name: fitdata
     dependencies: [gendata]
     scheduler:
@@ -56,4 +56,4 @@ stages:
           image: 'reanahub/reana-env-root6'
           imagetag: '6.18.04'
           resources:
-            - kubernetes_memory_limit: '256Mi'
+            - kubernetes_memory_limit: '256Gi'
```

and the cluster is set with maximum memory of 10 Gi:

```console
$ rg Gi dev/values.yaml
117:kubernetes_jobs_memory_limit: 3Gi
118:kubernetes_jobs_max_user_memory_limit: 10Gi
```

Then the workflow in question is not scheduled and stays "forever" queued:

```console
$ reana-client status -w roofit-yad-256gi                
NAME               RUN_NUMBER   CREATED               STATUS
roofit-yad-256gi   1            2022-02-08T12:27:55   queued
```

and requeued:

```console
$ kubectl logs deployment/reana-server scheduler 
2022-02-08 12:27:56,452 | root | MainThread | INFO | Received workflow: {'user': 'UUUU', 'workflow_id_or_name': 'roofit-yad-256gi.1', 'parameters': {'input_parameters': {}, 'operational_options': {'accept_metadir': True}}, 'priority': 0, 'min_job_memory': 274877906944.0}
2022-02-08 12:27:56,560 | root | MainThread | INFO | REANA not ready to run workflow roofit-yad-256gi.1.
2022-02-08 12:28:11,567 | root | MainThread | INFO | Received workflow: {'user': 'UUUU', 'workflow_id_or_name': 'roofit-yad-256gi.1', 'parameters': {'input_parameters': {}, 'operational_options': {'accept_metadir': True}}, 'priority': 0, 'min_job_memory': 274877906944.0}
[...]

PR #438, opened 2022-02-09
rest: validate Kubernetes job memory limits before publishing workflow

audrium:
closes https://github.com/reanahub/reana-server/issues/437

About the issue:
It only appears in the `balanced` scheduling mode and with working Kubernetes metrics-server, probably that's why it slipped through our testing. The root cause of the issue happens [here](https://github.com/reanahub/reana-server/blob/master/reana_server/scheduler.py#L38). In `balanced` mode we also measure if at least one workflow job could be started:
- We take `min_job_memory` [here](https://github.com/reanahub/reana-server/blob/master/reana_server/utils.py#L114) and send it as a param in message header to the scheduler.
- The scheduler [verifies if there are enough memory to execute the job](https://github.com/reanahub/reana-server/blob/master/reana_server/scheduler.py#L32). If not - it reschedules it for later. 
- And here is the problem - if `workflow_min_job_memory` exceeds `max_node_available_memory` it can never be started and is stuck if never ending requeuing loop.

This fix is to validate the maximal value of initial job memory against `REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT` to fail early. Note that it only checks the initial jobs which are returned from `estimate_complexity()`. All the rest workflow jobs are validated with [this check](https://github.com/reanahub/reana-job-controller/blob/master/reana_job_controller/kubernetes_job_manager.py#L484).

To test:
- In order to reproduce the issue follow steps described in https://github.com/reanahub/reana-server/issues/437
- In addition change `REANA_WORKFLOW_SCHEDULING_POLICY: "balanced"` in Helm values
- Make sure you have working Kubernetes metrics-server. Instructions to install can be found [here](https://github.com/reanahub/reana/wiki/Tips-for-Kind#install-metrics-server-locally)
- Checkout https://github.com/reanahub/reana/pull/615

### DIFF
--- a/reana_server/complexity.py
+++ b/reana_server/complexity.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2021 CERN.
+# Copyright (C) 2021, 2022 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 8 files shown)


reana_server/complexity.py
```diff
@@ -11,8 +11,37 @@
...
+    max_job_memory = max(complexity, key=lambda x: x[1])[1]
+
+    if (
+        REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT_IN_BYTES
+        and max_job_memory > REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT_IN_BYTES
+    ):
+        raise REANAKubernetesMemoryLimitExceeded(
+            f'The "kubernetes_memory_limit" provided in the workflow ({workflow_name}) exceeds the limit ({REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT}).'
+        )
+    return max_job_memory
```
VMois, line 44: This function just validates and doesn't need to return any value. I think `return` can be removed.

reana_server/complexity.py
```diff
@@ -11,8 +11,37 @@
...
-from reana_server.config import REANA_KUBERNETES_JOBS_MEMORY_LIMIT
+from reana_server.config import (
+    REANA_KUBERNETES_JOBS_MEMORY_LIMIT,
+    REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT,
+    REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT_IN_BYTES,
+)
+
+
+def validate_job_memory_limits(
+    workflow_name: str, complexity: List[Tuple[int, float]]
```
VMois, line 24: *minor:* I think, `workflow_name` is also unnecessary as it is obvious from the error message in `reana-client` what failed:

```
...
==> ERROR: Cannot start workflow q-1.1: 
The "kubernetes_memory_limit" provided in the workflow (q-1.1) exceeds the limit (8Gi).
```

reana_server/complexity.py
audrium: Function looks very weird without `return`, I've left `return None`

reana_server/complexity.py
audrium: Agree, my intention was to have the workflow name in the logs for better traceability, but it's easy to get it there from the requests

reana_server/complexity.py
VMois: *question(non-blocking):* What do you mean by "weird"?

VMois (approved):

PR #411, opened 2021-10-27
scheduler: catch rabbitmq message conflict exception

audrium:
closes https://github.com/reanahub/reana-message-broker/issues/41

While testing https://github.com/reanahub/reana/pull/558 I've noticed that in some cases after killing `reana-message-broker` the queue gets stuck with the following error in the `scheduler`:
```
2021-10-27 12:08:05,443 | root | MainThread | ERROR | Something went wrong while calling RWC :
409 CONFLICT
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/reana_server/scheduler.py", line 143, in on_message
    ) = current_rwc_api_client.api.set_workflow_status(
  File "/usr/local/lib/python3.8/site-packages/bravado/http_future.py", line 271, in result
    swagger_result = self._get_swagger_result(incoming_response)
  File "/usr/local/lib/python3.8/site-packages/bravado/http_future.py", line 124, in wrapper
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/bravado/http_future.py", line 299, in _get_swagger_result
    unmarshal_response(
  File "/usr/local/lib/python3.8/site-packages/bravado/http_future.py", line 352, in unmarshal_response
    raise_on_expected(incoming_response)
  File "/usr/local/lib/python3.8/site-packages/bravado/http_future.py", line 416, in raise_on_expected
    raise make_http_exception(
bravado.exception.HTTPConflict: 409 CONFLICT
```
It happens because `reana-message-broker` is being killed while still processing the message which is not yet acknowledged. After this happens new `reana-message-broker` pod starts to send the same message to the consumer which triggers this error. This PR catches these cases and drops the message.

To test:
- Should be tested with https://github.com/reanahub/reana/pull/558
- Similar approach as https://github.com/reanahub/reana/pull/558 (NFS and  sleep time (e.g. sleep 60) to the helloworld serial demo)
- Modify helm values (similar to -dev environment):
  - `REANA_SCHEDULER_REQUEUE_SLEEP: 15`
  - `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS: 5`
 - Launch 10 helloworld demos and kill `reana-message-broker-0` after some have started
 - Check that all workflows finish as expected
 - Check `scheduler` logs for `bravado.exception.HTTPConflict: 409 CONFLICT` error

PR #439, opened 2022-02-14
scheduler: add retry mechanism

### LINKED ISSUE #436: scheduler: avoid "infinite" requeuing
**Current behaviour**

Observed on REANA 0.8.0.

If some workflow cannot be scheduled for a reason or another, it is requeued for later:

```console
$ kubectl logs reana-server scheduler
2022-02-07 10:03:47,745 | root | MainThread | INFO | Received workflow: {'user': 'uuuu', 'workflow_id_or_name': 'wwww', 'parameters': {'input_parameters': {}, 'operational_options': {}}, 'priority': 98, 'min_job_memory': 12884901888.0}
2022-02-07 10:03:47,943 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:03:52,950 | root | MainThread | INFO | Received workflow: {'user': 'uuuu', 'workflow_id_or_name': 'wwww', 'parameters': {'input_parameters': {}, 'operational_options': {}}, 'priority': 98, 'min_job_memory': 12884901888.0}
2022-02-07 10:03:53,289 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:03:58,291 | root | MainThread | INFO | Received workflow: {'user': 'uuuu', 'workflow_id_or_name': 'wwww', 'parameters': {'input_parameters': {}, 'operational_options': {}}, 'priority': 98, 'min_job_memory': 12884901888.0}
2022-02-07 10:03:58,563 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
```

This can go on for a very long time:

```console
$ kubectl logs reana-server scheduler | grep 'REANA not ready to run workflow WWWW. | head
2022-02-07 10:03:47,943 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:03:53,289 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:03:58,563 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:04:03,790 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:04:08,992 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:04:14,184 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:04:19,376 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:04:24,583 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:04:29,778 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-07 10:04:34,981 | root | MainThread | INFO | REANA not ready to run workflow WWWW.

$ kubectl logs reana-server scheduler | grep 'REANA not ready to run workflow WWWW. | tail
2022-02-08 09:35:20,398 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-08 09:35:25,623 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-08 09:35:30,981 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-08 09:35:36,365 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-08 09:35:41,662 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-08 09:35:46,888 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-08 09:35:52,188 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-08 09:35:57,434 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-08 09:36:02,691 | root | MainThread | INFO | REANA not ready to run workflow WWWW.
2022-02-08 09:36:07,955 | root | MainThread | INFO | REANA not ready to run workflow WWWW.

$ kubectl logs reana-server scheduler | grep 'REANA not ready to run workflow WWWW. | wc -l
16066
```

This can cause scheduling troubles in case a workflow has very high priority.

**Expected behaviour**

If a workflow cannot be scheduled for more than N times, it is better to make
it fail and report it back to user and "liberate the queue".

We should make the behaviour configurable and exposed to cluster admins via
Helm environment variables, similarly to `REANA_SCHEDULER_REQUEUE_SLEEP`, so
that the cluster admins can decide as to how many times a workflow rescheduling
can fail.

We should configure reasonable defaults, so that the above use case when a
workflow was unsuccessfully rescheduled more than 16000 times wouldn't happen.

(A good default might be perhaps 100 or 200 times, which with the default
requeue sleep of 15 seconds would give about an hour.)

PRs:

- https://github.com/reanahub/reana-db/pull/161
- https://github.com/reanahub/reana-server/pull/439
- https://github.com/reanahub/reana-commons/pull/336
- https://github.com/reanahub/reana/pull/616

VMois:
closes #436

**Important:** there are two commits in a not-ready state because I moved some scheduler code to a separate folder. if reviewers will not object to the change, I will squash commits before merging. 

How to test:

1. Checkout all related PRs from the issue description

2. Edit `values-dev.yaml`:

```yaml
components:
    reana_server:
      image: reanahub/reana-server
      environment:
        REANA_SCHEDULER_REQUEUE_SLEEP: 5
        REANA_SCHEDULER_REQUEUE_COUNT: 10
        REANA_WORKFLOW_SCHEDULING_POLICY: "balanced"
...
kubernetes_jobs_memory_limit: 2Gi
kubernetes_jobs_max_user_memory_limit: 24Gi  # this value should be bigger then your available memory
```

3. Re-deploy REANA

4. Set a memory limit for a job in a workflow that still fits within the specified limit but exceeds your machine capacity. For example, `20Gi`.

5. Start a workflow, observe logs from `scheduler`.

6. Workflow should fail after a couple of retries, check `reana-client logs`.

### DIFF
--- a/reana_server/scheduler.py
+++ b/reana_server/scheduler.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2019, 2020, 2021 CERN.
+# Copyright (C) 2019, 2020, 2021, 2022 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -10,8 +10,9 @@
 
 import json
 import logging
-from time import sleep
 from functools import partial
+from time import sleep
+from typing import Dict
 
 from bravado.exception import HTTPBadGateway, HTTPNotFound, HTTPConflict
 from kubernetes.client.rest import ApiException
@@ -20,12 +21,19 @@ from sqlalchemy.exc import SQLAlchemyError
 
 from reana_commons.config import REANA_MAX_CONCURRENT_BATCH_WORKFLOWS
 from reana_commons.consumer import BaseConsumer
+from reana_commons.publisher import WorkflowStatusPublisher
 from reana_commons.k8s.api_client import current_k8s_corev1_api_client
 from reana_db.database import Session
 from reana_db.models import Workflow, RunStatus
 
-from reana_server.api_client import current_rwc_api_client
-from reana_server.config import REANA_SCHEDULER_REQUEUE_SLEEP
+from reana_server.api_client import (
+    current_rwc_api_client,
+    current_workflow_submission_publisher,
+)
+from reana_server.config import (
+    REANA_SCHEDULER_REQUEUE_SLEEP,
+    REANA_SCHEDULER_REQUEUE_COUNT,
+)
 from reana_server.status import NodesStatus
 
 
@@ -114,6 +122,10 @@ class WorkflowExecutionScheduler(BaseConsumer):
             queue="workflow-submission", **kwargs
         )
 
+        self.workflow_status_publisher = WorkflowStatusPublisher(
+            connection=self.connection
+        )
+
     def get_consumers(self, Consumer, channel):
         """Implement providing kombu.Consumers with queues/callbacks."""
         return [
@@ -125,18 +137,51 @@ class WorkflowExecutionScheduler(BaseConsumer):
             )
         ]
 
-    def on_message(self, workflow_submission, message):
+    def _fail_workflow(self, workflow_id: str, logs: str = "") -> None:
+        self.workflow_status_publisher.publish_workflow_status(
+            workflow_id, status=RunStatus.failed.value, logs=logs,
+        )
+
+    def _retry_submission(self, workflow_id: str, workflow_submission: Dict) -> None:
+        retry_count = workflow_submission.get("retry_count", 0)
+        if retry_count >= REANA_SCHEDULER_REQUEUE_COUNT:
+            error_message = (
+                f"Workflow {workflow_submission['workflow_id_or_name']} failed to schedule after "
+                f"{retry_count} retries. Giving up."
+            )
+            logging.error(error_message)
+            self._fail_workflow(workflow_id, logs=error_message)
+        else:
+            current_workflow_submission_publisher.publish_workflow_submission(
+                user_id=workflow_submission["user"],
+                workflow_id_or_name=workflow_submission["workflow_id_or_name"],
+                parameters=workflow_submission["parameters"],
+                priority=workflow_submission["priority"],
+                min_job_memory=workflow_submission["min_job_memory"],
+                retry_count=retry_count + 1,
+            )
+
+    def on_message(self, body, message):
         """On new workflow_submission event handler."""
-        workflow_submission = json.loads(workflow_submission)
-        logging.info("Received workflow: {}".format(workflow_submission))
+        workflow_submission = json.loads(body)
+        logging.info(f"Received workflow: {workflow_submission}")
+
+        workflow_submission_copy = workflow_submission.copy()
+
+        workflow_id = workflow_submission["workflow_id_or_name"]
         workflow_min_job_memory = workflow_submission.pop("min_job_memory", 0)
+
+        workflow_submission.pop("priority", None)
+        workflow_submission.pop("retry_count", None)
+
         if reana_ready(workflow_min_job_memory):
-            logging.info("Starting queued workflow: {}".format(workflow_submission))
-            workflow_submission.pop("priority", 0)
+            logging.info(f"Starting queued workflow: {workflow_id}")
             workflow_submission["status"] = "start"
+
+            retry = True
+            started = False
+
             try:
-                requeue = True
-                started = False
                 (
                     response,
                     http_response,
@@ -144,15 +189,10 @@ class WorkflowExecutionScheduler(BaseConsumer):
                     **workflow_submission
                 ).result()
                 http_response_json = http_response.json()
-                if http_response.status_code == 200:
-                    started = True
-                    logging.info(
-                        f'Workflow {http_response_json["workflow_id"]} successfully started.'
-                    )
-                else:
-                    logging.error(
-                        f"RWC returned an unexpected status code:\n {http_response_json}"
-                    )
+                started = True
+                logging.info(
+                    f'Workflow {http_response_json["workflow_id"]} successfully started.'
+                )
 
             except HTTPBadGateway as api_e:
                 logging.error(
@@ -161,26 +201,28 @@ class WorkflowExecutionScheduler(BaseConsumer):
                     exc_info=True,
                 )
             except HTTPNotFound as not_found_e:
+                # if workflow is not found, we cannot retry or report an error to workflow logs
+                retry = False
                 logging.error(
                     "Workflow failed to start because it does not exist or was deleted \n"
                     f"{not_found_e}",
                     exc_info=True,
                 )
-                requeue = False
             except HTTPConflict as e:
+                retry = False
                 logging.error(
-                    f"Workflow failed to start because of Rabbit MQ message conflict.\n {e}",
+                    f"Workflow failed to start because of duplicated message from RabbitMQ.\n {e}",
                     exc_info=True,
                 )
-                requeue = False
             except Exception as e:
                 logging.error(
                     f"Something went wrong while calling RWC:\n {e}", exc_info=True
                 )
             finally:
                 sleep(REANA_SCHEDULER_REQUEUE_SLEEP)
-                if not started and requeue:
-                    message.reject(requeue=True)
+                if not started and retry:
+                    message.reject()
+                    self._retry_submission(workflow_id, workflow_submission_copy)
                 else:
                     message.ack()
         else:
@@ -188,4 +230,5 @@ class WorkflowExecutionScheduler(BaseConsumer):
                 f'REANA not ready to run workflow {workflow_submission["workflow_id_or_name"]}.'
             )
             sleep(REANA_SCHEDULER_REQUEUE_SLEEP)
-            message.reject(requeue=True)
+            message.reject()
+            self._retry_submission(workflow_id, workflow_submission_copy)
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -12,7 +12,6 @@ import csv
 import io
 import json
 import logging
-import re
 import secrets
 import sys
 import os
@@ -119,7 +118,7 @@ def publish_workflow_submission(workflow, user_id, parameters):
         validate_job_memory_limits(complexity)
     current_workflow_submission_publisher.publish_workflow_submission(
         user_id=str(user_id),
-        workflow_id_or_name=workflow.get_full_workflow_name(),
+        workflow_id_or_name=str(workflow.id_),
         parameters=parameters,
         priority=workflow_priority,
         min_job_memory=workflow_min_job_memory,
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -1,18 +1,20 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2019, 2020, 2021 CERN.
+# Copyright (C) 2019, 2020, 2021, 2022 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """REANA-Server Workflow Execution Scheduler."""
+import base64
+import json
 
-
-from bravado.exception import HTTPError
+import pytest
+from bravado.exception import HTTPBadGateway, HTTPNotFound, HTTPConflict
 from mock import DEFAULT, Mock, patch
-from reana_commons.publisher import WorkflowSubmissionPublisher
 
+from reana_server.api_client import WorkflowSubmissionPublisher
 from reana_server.scheduler import WorkflowExecutionScheduler
 
 
@@ -41,7 +43,7 @@ def test_scheduler_starts_workflows(
     assert in_memory_queue_connection.channel().queues["workflow-submission"].empty()
 
 
-def test_scheduler_requeues_workflows(
+def test_scheduler_requeues_when_not_ready(
     in_memory_queue_connection, default_in_memory_producer, consume_queue,
 ):
     """Test that the scheduler requeues workflows if conditions not met."""
@@ -52,6 +54,7 @@ def test_scheduler_requeues_workflows(
     with patch.multiple(
         "reana_server.scheduler",
         reana_ready=Mock(return_value=False),
+        current_workflow_submission_publisher=in_memory_wsp,
         REANA_SCHEDULER_REQUEUE_SLEEP=0,
     ):
         consume_queue(scheduler, limit=1)
@@ -60,10 +63,29 @@ def test_scheduler_requeues_workflows(
             .queues["workflow-submission"]
             .empty()
         )
+        message = (
+            in_memory_queue_connection.channel().queues["workflow-submission"].get()
+        )
+        message_body = base64.b64decode(message["body"]).decode("ascii")
+        message_body = json.loads(json.loads(message_body))
+        assert message_body["retry_count"] == 1
 
 
+@pytest.mark.parametrize(
+    "error,should_retry",
+    [
+        (HTTPBadGateway(Mock()), True),
+        (Exception(Mock()), True),
+        (HTTPNotFound(Mock()), False),
+        (HTTPConflict(Mock()), False),
+    ],
+)
 def test_scheduler_requeues_on_rwc_failure(
-    in_memory_queue_connection, default_in_memory_producer, consume_queue,
+    in_memory_queue_connection,
+    default_in_memory_producer,
+    consume_queue,
+    error,
+    should_retry,
 ):
     """Test scheduler requeues requests if RWC fails to start workflows."""
     scheduler = WorkflowExecutionScheduler(connection=in_memory_queue_connection)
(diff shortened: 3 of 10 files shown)


reana_server/scheduler/consumer.py
```diff
@@ -173,19 +215,27 @@ def on_message(self, workflow_submission, message):
                     exc_info=True,
                 )
                 requeue = False
+                should_fail_workflow_if_not_started = False
```
VMois, line 218: This is due to this comment - https://github.com/reanahub/reana-server/pull/411#issuecomment-952969077 . If the workflow is still running and it is fine but we cancel it in `scheduler`, the user will be confused on why workflow failed.

*note(non-blocking):* It would be much better if we can easily distinguish between error types - HTTPConflict doesn't say much. Instead of `HTTPConflict` error, would be good to get `DuplicatedStartError` and similar. Any ideas on that?

audrium: I haven't tested this edge case for now, but I think it should be transparent to user and we don't need to `fail` the workflow because of it. In case RabbitMQ gets restarted and duplicate message is sent to consumer, the workflow should still continue running and a duplicate message should be dropped silently. WDYT? In this case we could get rid of `should_fail_workflow_if_not_started`

reana_server/scheduler/consumer.py
```diff
@@ -125,18 +133,50 @@ def get_consumers(self, Consumer, channel):
...
+        )
+
+    def _retry_workflow_submission(
+        self, workflow_id: str, workflow_submission: Dict
+    ) -> None:
+        retry_count = workflow_submission.get("retry_count", 0)
+        if retry_count >= REANA_SCHEDULER_REQUEUE_COUNT:
+            error_message = (
+                f"Workflow {workflow_submission['workflow_id_or_name']} failed to schedule "
+                f"{retry_count + 1} times. Giving up."
```
audrium, line 148: Minor: I guess we don't need to do `retry_count + 1` here? With `REANA_SCHEDULER_REQUEUE_COUNT=10` I got `Workflow 36ea2285-aa10-4664-a2c7-4bf3e1a0e20c failed to schedule 11 times. Giving up.`

reana_server/scheduler/consumer.py
```diff
@@ -125,18 +133,50 @@ def get_consumers(self, Consumer, channel):
             )
         ]
 
-    def on_message(self, workflow_submission, message):
+    def _fail_workflow(self, workflow_id: str, logs: str = "") -> None:
+        self.workflow_status_publisher.publish_workflow_status(
+            workflow_id, status=RunStatus.failed.value, logs=logs,
+        )
+
+    def _retry_workflow_submission(
```
audrium, line 141: Extremely minor: perhaps naming this function `requeue_workflow()` as we [used to have](https://github.com/reanahub/reana-server/blob/f0d55e16ba01ae4e4ca53b823eaa56d6bd2572aa/reana_server/scheduler.py#L132) would be a bit shorter?

reana_server/scheduler/consumer.py
```diff
@@ -125,18 +133,50 @@ def get_consumers(self, Consumer, channel):
...
+        workflow_submission = json.loads(body)
+        logging.info(f"Received workflow: {workflow_submission}")
+        workflow_submission_copy = workflow_submission.copy()
+
+        workflow_id = workflow_submission["workflow_id_or_name"]
+
         workflow_min_job_memory = workflow_submission.pop("min_job_memory", 0)
         if reana_ready(workflow_min_job_memory):
-            logging.info("Starting queued workflow: {}".format(workflow_submission))
+            logging.info(f"Starting queued workflow: {workflow_id}")
```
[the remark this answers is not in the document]
VMois: We have had this information before. Printed for every workflow submission. Line 163.
In addition, `priority` and `min_job_memory` are removed before so you will not see them anyway.

reana_server/scheduler/consumer.py
VMois: That is why I added `should_fail_workflow_if_not_started=False`, so workflow **does not** fail in case of a duplicate message.

Maybe it is confusing to understand my intentions in the code. Do you have suggestions on how to make it more clear?

reana_server/scheduler/consumer.py
```diff
@@ -125,18 +133,50 @@ def get_consumers(self, Consumer, channel):
...
+
+    def on_message(self, body, message):
         """On new workflow_submission event handler."""
-        workflow_submission = json.loads(workflow_submission)
-        logging.info("Received workflow: {}".format(workflow_submission))
+        workflow_submission = json.loads(body)
+        logging.info(f"Received workflow: {workflow_submission}")
+        workflow_submission_copy = workflow_submission.copy()
+
+        workflow_id = workflow_submission["workflow_id_or_name"]
```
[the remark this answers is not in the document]
VMois: I like it with a line, that reads cleaner, but I will remove it.

reana_server/scheduler/consumer.py
VMois: Why do you think the longer name is less clean?

`requeue` implies it will re-send only, so I would still prefer `retry`.
But, now I look at it, the `workflow` part might be unnecessary. What do you think about `retry_submission()`?

reana_server/scheduler/consumer.py
VMois: Another idea will be to completely abandon numbers there and say just "Workflow <id> failed to schedule multiple times. Giving up."

audrium: Yeah, exactly what I wanted to say, "Workflow failed after 10 retries. Giving up.". sounds much better in my eyes, since 10 is the number which was set by the admin. I also think it's better to print it, just to be sure what is the number used while debugging

VMois:
Couple of issues discovered by @audrium:

- [ ] Need to `.pop("retry_count")`, otherwise submission to r-workflow-controller will fail (**important**);

- [ ] The logic behind whatever workflow should be retried/fail is not clear, needs improvement;

- [ ] Bravado API `result()` is raising an exception if status code is not 2xx, but we still have an `if` to check status code; doesn't look consistent: https://github.com/reanahub/reana-server/blob/4440d3b2d418c7f39f091a1d7f0df38ddeca82e8/reana_server/scheduler.py#L147-L153

Thank you!

VMois:
> Bravado API result() is raising an exception if status code is not 2xx, but we still have an if to check status code

According to [Bravado code](https://github.com/Yelp/bravado/blob/v10.3.2/bravado/http_future.py#L281), it will raise an exception on anything that is not 2xx. According to our OpenAPI docs for `r-workflow-controller`, it will always send 200 if the operation on `set_workflow_status` was correct. So, the `if/else` is unnecessary.

reana_server/scheduler.py
```diff
@@ -125,67 +135,106 @@ def get_consumers(self, Consumer, channel):
...
                     f"Workflow failed to start because of Rabbit MQ message conflict.\n {e}",
                     exc_info=True,
                 )
-                requeue = False
             except Exception as e:
+                retry = True
+                should_fail_on_error = True
                 logging.error(
-                    f"Something went wrong while calling RWC:\n {e}", exc_info=True
+                    f"Something went wrong while calling RWC:\n{e}", exc_info=True
```
audrium, line 222: Minor: for consistency with other error logs we could keep the space

reana_server/scheduler.py
```diff
@@ -125,67 +135,106 @@ def get_consumers(self, Consumer, channel):
...
         if reana_ready(workflow_min_job_memory):
-            logging.info("Starting queued workflow: {}".format(workflow_submission))
-            workflow_submission.pop("priority", 0)
+            logging.info(f"Starting queued workflow: {workflow_id}")
             workflow_submission["status"] = "start"
+
+            retry = True
+            should_fail_on_error = True
+            started = False
+            error_message = ""
```
audrium, line 180: Minor: I guess there is no point in keeping this variable here if the value never changes?

reana_server/utils.py
```diff
@@ -119,7 +118,7 @@ def publish_workflow_submission(workflow, user_id, parameters):
         validate_job_memory_limits(complexity)
     current_workflow_submission_publisher.publish_workflow_submission(
         user_id=str(user_id),
-        workflow_id_or_name=workflow.get_full_workflow_name(),
+        workflow_id_or_name=str(workflow.id_),
```
audrium, line 121: Question: can you explain why this is needed?

tests/test_scheduler.py
```diff
@@ -11,8 +11,8 @@
 
 from bravado.exception import HTTPError
 from mock import DEFAULT, Mock, patch
-from reana_commons.publisher import WorkflowSubmissionPublisher
 
+from reana_server.api_client import WorkflowSubmissionPublisher
```
audrium, line 15: Leftover here

reana_server/scheduler.py
```diff
@@ -125,67 +135,106 @@ def get_consumers(self, Consumer, channel):
...
+                f"Workflow {workflow_submission['workflow_id_or_name']} failed to schedule after "
+                f"{retry_count} retries. Giving up."
+            )
+            logging.error(error_message)
+            self._fail_workflow(workflow_id, logs=error_message)
+        else:
+            workflow_submission["retry_count"] = retry_count + 1
+            workflow_submission["user_id"] = workflow_submission["user"]
+            del workflow_submission["user"]
+            current_workflow_submission_publisher.publish_workflow_submission(
```
audrium, line 156: Minor: alternatively we could specify all the arguments like it was done [here](https://github.com/reanahub/reana-server/blob/f0d55e16ba01ae4e4ca53b823eaa56d6bd2572aa/reana_server/scheduler.py#L141), it would be clear to what actually we do send and we would avoid `user_id` switch with `user`, WDYT?

reana_server/scheduler.py
```diff
@@ -125,67 +135,106 @@ def get_consumers(self, Consumer, channel):
...
-                        f"RWC returned an unexpected status code:\n {http_response_json}"
-                    )
+                started = True
+                logging.info(
+                    f'Workflow {http_response_json["workflow_id"]} successfully started.'
+                )
 
             except HTTPBadGateway as api_e:
+                retry = True
+                should_fail_on_error = True
```
audrium, line 197: Question: it's a bit confusing why do we have both flags set to `True` here? Since by reading I want to quickly understand the logic and it seems that If it retries it will never fail based on `should_fail_on_error`? In this case it will retry, so perhaps `should_fail_on_error` should be `False`?

reana_server/scheduler.py
```diff
@@ -125,67 +135,106 @@ def get_consumers(self, Consumer, channel):
...
+                should_fail_on_error = False
                 logging.error(
                     "Workflow failed to start because it does not exist or was deleted \n"
                     f"{not_found_e}",
                     exc_info=True,
                 )
-                requeue = False
             except HTTPConflict as e:
+                retry = False
+                should_fail_on_error = False
```
audrium, line 213: Minor: IMO small comment to explain that in this case the message should be `acked` silently would really help here

reana_server/scheduler.py
```diff
@@ -125,67 +135,106 @@ def get_consumers(self, Consumer, channel):
...
+                retry = True
+                should_fail_on_error = True
                 logging.error(
                     "Workflow failed to start because RWC got an error while calling"
                     f"an external service (i.e. DB):\n {api_e}",
                     exc_info=True,
                 )
             except HTTPNotFound as not_found_e:
+                retry = False
+                should_fail_on_error = False
```
audrium, line 205: I think in this case `should_fail_on_error` should be set to `True`?

reana_server/scheduler.py
```diff
@@ -125,67 +135,106 @@ def get_consumers(self, Consumer, channel):
...
                 logging.error(
                     f"Workflow failed to start because of Rabbit MQ message conflict.\n {e}",
                     exc_info=True,
                 )
-                requeue = False
             except Exception as e:
+                retry = True
+                should_fail_on_error = True
```
audrium, line 220: Similar to my observation above

reana_server/scheduler/consumer.py
```diff
@@ -150,42 +193,57 @@ def on_message(self, workflow_submission, message):
...
+                retry = True
+                should_fail_on_error = True
                 logging.error(
                     "Workflow failed to start because RWC got an error while calling"
                     f"an external service (i.e. DB):\n {api_e}",
                     exc_info=True,
                 )
             except HTTPNotFound as not_found_e:
+                retry = False
+                should_fail_on_error = False
```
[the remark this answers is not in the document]
audrium: Yes, I agree, it's a pity that I spotted your comment too late.. I was about to write a long comment about it in a review.. Please ignore some of my comments about the flags

reana_server/utils.py
VMois: Because `scheduler` sends messages to `job-status-consumer` to fail workflows, and `job-status-consumer` needs UUID, I replaced it here with UUID.

Why `str`? Because, `workflow.id_` is not a string, I needed to convert it.

audrium: I see, then probably it would be better to rename `workflow_id_or_name` to `workflow_uuid` (like in `WorkflowStatusPublisher`) to make it more clear what value we are sending

reana_server/scheduler.py
VMois: I will still keep them as default values for visiblity.

reana_server/scheduler.py
VMois: You cannot fail a workflow that doesn't exist (`deleted` -> `failed` transition is not possible). `job-status-consumer` will just ignore the message. We will need to modify `job-status-consumer` for this.

reana_server/utils.py
VMois: Yep. Will update the other PR.

reana_server/scheduler/consumer.py
```diff
@@ -173,19 +214,27 @@ def on_message(self, workflow_submission, message):
                     exc_info=True,
                 )
                 requeue = False
+                should_fail_workflow_if_not_started = False
             except Exception as e:
                 logging.error(
-                    f"Something went wrong while calling RWC:\n {e}", exc_info=True
+                    f"Something went wrong while calling RWC:\n{e}", exc_info=True
```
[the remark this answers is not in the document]
VMois: Because space at the beginning of the line doesn't make sense. But I can return it to align with other logs.

reana_server/scheduler.py
VMois: Not sure it is worth doing for now.

reana_server/scheduler.py
VMois: What about some duplication like this, makes things very clear:

```python
try:
    (
        response,
        http_response,
    ) = current_rwc_api_client.api.set_workflow_status(
        **workflow_submission
    ).result()
    http_response_json = http_response.json()
    logging.info(
        f'Workflow {http_response_json["workflow_id"]} successfully started.'
    )
    sleep(REANA_SCHEDULER_REQUEUE_SLEEP)
    message.ack()
except HTTPBadGateway as api_e:
    logging.error(
        "Workflow failed to start because RWC got an error while calling"
        f"an external service (i.e. DB):\n {api_e}",
        exc_info=True,
    )
    sleep(REANA_SCHEDULER_REQUEUE_SLEEP)
    message.reject()
    self._retry_submission(workflow_id, workflow_submission_copy)
except HTTPNotFound as not_found_e:
    logging.error(
        "Workflow failed to start because it does not exist or was deleted \n"
        f"{not_found_e}",
        exc_info=True,
    )
    sleep(REANA_SCHEDULER_REQUEUE_SLEEP)
    # if workflow is not found, we cannot retry or report an error to workflow logs
    message.reject()
except HTTPConflict as e:
    logging.error(
        f"Workflow failed to start because of duplicated message from RabbitMQ.\n {e}",
        exc_info=True,
    )
    sleep(REANA_SCHEDULER_REQUEUE_SLEEP)
    message.reject()
except Exception as e:
    logging.error(
        f"Something went wrong while calling RWC:\n {e}", exc_info=True
    )
    sleep(REANA_SCHEDULER_REQUEUE_SLEEP)
    message.reject()
    self._retry_submission(workflow_id, workflow_submission_copy)
```

My concerns with comments are that they can get outdated, and no one will change them.

audrium: IMO something simple as this would work (haven't tested):
```python
    def on_message(self, body, message):
        """On new workflow_submission event handler."""
        workflow_submission = json.loads(body)
        logging.info(f"Received workflow: {workflow_submission}")

        workflow_min_job_memory = workflow_submission.pop("min_job_memory", 0)
        retry_count = workflow_submission.pop("retry_count", 0)

        if reana_ready(workflow_min_job_memory):
            logging.info(f"Starting queued workflow: {workflow_submission}")
            workflow_submission["status"] = "start"
            priority = workflow_submission.pop("priority", 0)

            requeue = True
            started = False

            try:
                (
                    response,
                    http_response,
                ) = current_rwc_api_client.api.set_workflow_status(
                    **workflow_submission
                ).result()
                http_response_json = http_response.json()
                if http_response.status_code == 200:
                    started = True
                    logging.info(
                        f'Workflow {http_response_json["workflow_id"]} successfully started.'
                    )
                else:
                    logging.error(
                        f"RWC returned an unexpected status code:\n"
                        f"{http_response_json}"
                    )

            except HTTPBadGateway as api_e:
                logging.error(
                    "Workflow failed to start because RWC got an error while calling"
                    f"an external service (i.e. DB):\n {api_e}",
                    exc_info=True,
                )
            except HTTPNotFound as not_found_e:
                logging.error(
                    "Workflow failed to start because it does not exist or was deleted \n"
                    f"{not_found_e}",
                    exc_info=True,
                )
                requeue = False
            except HTTPConflict as e:
                logging.error(
                    f"Workflow failed to start because of Rabbit MQ message conflict.\n {e}",
                    exc_info=True,
                )
                requeue = False
            except Exception as e:
                logging.error(
                    f"Something went wrong while calling RWC:\n{e}", exc_info=True
                )
            finally:
                sleep(REANA_SCHEDULER_REQUEUE_SLEEP)
                if not started and requeue:
                    message.reject()
                    self._retry_submission(
                        **workflow_submission,
                        priority=priority,
                        min_job_memory=workflow_min_job_memory,
                    )
                else:
                    message.ack()
        else:
            logging.info(
                f'REANA not ready to run workflow {workflow_submission["workflow_id_or_name"]}.'
            )
            sleep(REANA_SCHEDULER_REQUEUE_SLEEP)
            message.reject()
            self._retry_submission(
                **workflow_submission,
                min_job_memory=workflow_min_job_memory,
                retry_count=retry_count,
            )
```

audrium (approved):
Works nicely!

PR #444, opened 2022-02-22
rest: add new `/launch` blueprint and route

mvidalgarcia:
closes reanahub/reana-ui#221

### DIFF
--- a/reana_server/rest/launch.py
+++ b/reana_server/rest/launch.py
@@ -1,0 +1,119 @@
+# This file is part of REANA.
+# Copyright (C) 2022 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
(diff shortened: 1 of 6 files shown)


reana_server/rest/launch.py
```diff
@@ -0,0 +1,108 @@
...
+      summary: Launch workflow from a remote REANA specification file.
+      description: >-
+        This resource expects a remote reference to a REANA specification
+        file needed to launch a workflow via URL.
+      operationId: launch
+      produces:
+        - application/json
+      parameters:
+        - name: url
+          in: query
```
mdonadoni, line 44: The parameters (`url`, `name` and `parameters`) are passed in the body, aren't they? If this is the case, then we also have to add that this endpoint consumes JSON data.

mdonadoni: Yes, this is exactly what I had in mind!

mdonadoni (approved):

Issue #440, opened 2022-02-21
epic: implement "run" backend prototype


tiborsimko:
epic: implement "run" backend prototype

Implement first prototype of the "run" functionality on the backend side.

When a URL like `/run?from=...&name=myworkflow&parameters=myparam;myvalue` is hit, the backend should:

- create workflow with desired name -> this will pre-reserve uuid42
    - internal function; just "pre-register" workflow, since we don't have specs yet
    - this will create also workspace where to fetch sources

- fetch workflow specs
    - if git, then do shallow clone such as `git clone --depth=1` and use `reana.yaml` discovered there
    - if URL to exact `reana.yaml` location, then `wget` it
        - but beware that we won't get input files
    - if Zenodo, then "discover" if `reana.yaml` can be hidden inside tarballs

- fetch workflow parameters
    - if user specifies other input parameters via URL
    - could be perhaps defined in `inputs.yaml` file for Yadage etc

- validate specs on sever-side
    - will require to move some of the workflow validation functions from client-side to server-side

- submit for execution
    - now that everything is specified and "cloned" to workflow, we can launch it

- redirect to /details/uuid42
    - for Web endpoints; for API endpoints return uuid42

This is an epic style issue with musings how the workflow behind the scene could look like.

This issue will generate other issues such as flexible `WorkflowSpecificationFetcher` class, etc. The subsequent issues should stem from this one for better work splitting and progress tracking.

VMois:
Some examples we might target:

- [Zenodo record](https://zenodo.org/record/5752285)
- Github, REANA demo example can be used for a testing

PR #448, opened 2022-03-03
rest: implement launch logic 

audrium:
closes https://github.com/reanahub/reana-server/issues/440


reana_server/config.py
```diff
@@ -68,6 +68,11 @@
 SUPPORTED_COMPUTE_BACKENDS = json.loads(os.getenv("REANA_COMPUTE_BACKENDS", "[]")) or []
 """List of supported compute backends."""
 
+DEFAULT_NAME_FOR_LAUNCHED_WORKFLOWS = "remote_workflow"
```
tiborsimko, line 71: You can simply use `workflow` as the default name when nothing is specified. Just as we do from CLI client. 

I'll create a separate issue for generating nice user-visible names for each external source.  (So that this PR would simply use it later.)

reana_server/fetcher.py
```diff
@@ -119,6 +119,7 @@ def fetch(self) -> None:
         if self._git_ref:
             repository.remote().fetch(self._git_ref, depth=1)
             repository.git.checkout(self._git_ref)
+        shutil.rmtree(os.path.join(self._output_dir, ".git"))
```
audrium, line 122: For git repos we also download `.git` directory with lots of files which we have to move from one directory to another, so it's better to remove it here

tiborsimko: Is shallow copy still creating a lot of these internal files?

Anyway, it should be OK to remove top-level `.git` directory, as I guess users won't assume that we have access to `git` commands at the time of their workflow execution... Albeit could give some troubles if somebody would use `git lfs` or something for data access.  (We don't have such a use case yet I think.)

audrium: For `roofit` example there are additional ~60 `git` files, additional 3 pages of files displayed in UI workspace tab. Perhaps later we could introduce some logic to keep these files in case workflow needs it

reana_server/fetcher.py
```diff
@@ -236,11 +237,11 @@ def get_fetcher(url: str, output_dir: str) -> WorkflowFetcherBase:
 
     if extension == ".git":
         return WorkflowFetcherGit(url, output_dir)
+    elif extension == ".zip":
+        return WorkflowFetcherZip(url, output_dir, archive_name=basename)
```
tiborsimko, line 241: Hmm, we may want to refactor the source guessing later down the line, since "zip" can be hosted on URL, on GitHub... So I see the if/elif/elif chain order can be fragile. Basically, we could consider URL-fetcher and GitHub-fetcher as two main "source fetcher" methods, and Zip only as a kind of post-processor of what was fetched from wherever... CC @mdonadoni for later.

tiborsimko (approved):

PR #450, opened 2022-03-08
config: add custom rate limiter to launch endpoint

### LINKED ISSUE #443: rate limit for run-on-reana launch endpoint
Currently, REANA cluster administrators can set API endpoint rate limiting via `RATELIMIT_GUEST_USER` and `REANA_RATELIMIT_GUEST_USER` environment variables. The default value is "20 per second".

This is generally OK for "fast" endpoints, but it may be too much for "slow" endpoints.

For the Run-on-REANA sprint, we shall have a `run?from=...` like endpoint which will gather the workflow specification from external sources. If the specification is living inside big tarball, it may take several seconds to get the sources and start a workflow.

This could bring our cluster to the knees.

It is therefore good to investigate solutions to prevent this:

- For example, on the app-level, [Flask-Limiter](https://github.com/alisaifee/flask-limiter) can have different rates for different endpoints. We can simply decorate different endpoints differently and expose that to cluster admins. A complex setup of many different environment variables, or a variable encoding a Python dictionary, could be envisaged. But a simple setup of distinguishing only two cases, a "fast" endpoint and a "slow" endpoint, with two different corresponding rate limiting values, may already be enough.

- For example, on the devops-level, we could investigate [traefik rate-limiting](https://doc.traefik.io/traefik/middlewares/http/ratelimit/) or [haproxy maxconn](https://doc.traefik.io/traefik/middlewares/http/ratelimit/) which would allow only N requests and queue the others until these are solved, preventing cluster overload.

We should investigate best options and implement either an in-app solution or an external-service solution to prevent cluster overload when many hundreds of users would click on "Run-on-REANA" badge at the same time.

VMois:
closes #443

Not sure if we want to have a flexible configuration for endpoint here like `slow` and `fast`?

How to test:

1. Redeploy cluster
2. Try to hit `/api/launch` endpoint more than 1 time per 30 seconds. Example curl command:

```bash
curl -X POST -H "Content-Type: application/json" -d '{"url": "http://hello.com", "name":"hello"}' -k https://localhost:30443/api/launch?access_token=<REANA_TOKEN>
```


reana_server/config.py
```diff
@@ -161,22 +162,23 @@ def _(x):
...
+        return default
 
 
 # Note: users that are connecting via reana-client will be treated as guests by the Invenio framework
 RATELIMIT_GUEST_USER = _get_rate_limit("REANA_RATELIMIT_GUEST_USER", "20 per second")
 RATELIMIT_AUTHENTICATED_USER = _get_rate_limit(
     "REANA_RATELIMIT_AUTHENTICATED_USER", "20 per second"
 )
+RATELIMIT_PER_ENDPOINT = {
+    "launch.launch": _get_rate_limit("REANA_RATELIMIT_LAUNCH_WORKFLOW", "1/30 second"),
```
VMois, line 180: *thought(non-blocking):* Because we have `launch` Flask blueprint, and the endpoint is also named `launch` the naming is confusing here.

reana_server/config.py
```diff
@@ -161,22 +162,23 @@ def _(x):
 # ===========================
 
 
-def _is_valid_rate_limit(rate_limit: str) -> bool:
-    return bool(
-        re.match(r"[0-9]+(\sper\s|\/)(second|minute|hour|day|month|year)", rate_limit)
-    )
```
VMois, line 167: Never liked it ad happy to get rid of it, and replace it with `limits.util.parse`. Fewer things to worry about from our side.

PR #454, opened 2022-03-14
fetcher: add limitations to the download of files

mdonadoni:
This PR adds some limitations and checks to the download of files:
- Connection and read timeout
- Limit to the size of the file to be downloaded
- Allow only `http` and `https`

How to test:
Launch this workflow: `https://zenodo.org/record/5752285/files/circular-health-data-processing-master.zip`

In each of these cases, the workflow is not launched and an error is reported:
- Set `FETCHER_MAXIMUM_FILE_SIZE` to a small value (e.g. 1024)
- Remove `https` from `FETCHER_ALLOWED_SCHEMES`
- Set `FETCHER_REQUEST_TIMEOUT` to a small value (e.g. 0.1)

### DIFF
--- a/reana_server/fetcher.py
+++ b/reana_server/fetcher.py
@@ -13,16 +13,28 @@ import os
 import shutil
 from typing import List, Optional
 from urllib.parse import urlparse
-from urllib.request import urlretrieve
 import zipfile
 
 from git import Repo
+import requests
+from requests.exceptions import HTTPError, Timeout, RequestException
 
 from reana_server.config import (
+    FETCHER_ALLOWED_SCHEMES,
+    FETCHER_MAXIMUM_FILE_SIZE,
+    FETCHER_REQUEST_TIMEOUT,
     REGEX_CHARS_TO_REPLACE,
     WORKFLOW_SPEC_EXTENSIONS,
     WORKFLOW_SPEC_FILENAMES,
 )
+
+
+class REANAFetcherError(Exception):
+    """Workflow specification fetcher error."""
+
+    def __init__(self, message):
+        """Initialize REANAFetcherError exception."""
+        self.message = message
 
 
 class ParsedUrl:
@@ -42,6 +54,7 @@ class ParsedUrl:
         )
         self.hostname = self._parsed_url.hostname
         self.netloc = self._parsed_url.netloc
+        self.scheme = self._parsed_url.scheme
 
 
 class WorkflowFetcherBase(ABC):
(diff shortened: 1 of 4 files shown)


reana_server/fetcher.py
```diff
@@ -82,6 +95,63 @@ def _clean_workflow_name(name: str) -> str:
...
+            :returns: Number of bytes read from the response content.
+            """
+            read_bytes = 0
+            with open(output_path, "wb") as output_file:
+                # Use the same chunk size of `urlretrieve`
+                for chunk in response.iter_content(chunk_size=1024 * 8):
+                    read_bytes += len(chunk)
+                    output_file.write(chunk)
+                    if read_bytes > FETCHER_MAXIMUM_FILE_SIZE:
+                        break
```
audrium, line 123: According to the [docs](https://docs.python-requests.org/en/latest/user/advanced/#body-content-workflow):
`If you set stream to True when making a request, Requests cannot release the connection back to the pool unless you consume all the data or call [Response.close](https://docs.python-requests.org/en/latest/api/#requests.Response.close).`

Perhaps we should close the connection in this case

audrium (commented):
Nicely done! Left one minor question, but it works great.

Also the PR needs a rebase.

Outside this PR: have you considered some protection against zip bombs? I read there are some for `ZipFile.extractall()` method already implemented, but to be extra safe we could also check for any zips inside the zip, etc. as mentioned [here](https://stackoverflow.com/questions/1459080/how-can-i-protect-myself-from-a-zip-bomb)

reana_server/fetcher.py
mdonadoni: This should not be an issue, given that the request is made using a `with` statement (see line 127 of fetcher.py) as recommended by the docs: `If you find yourself partially reading request bodies (or not reading them at all) while using stream=True, you should make the request within a with statement to ensure it’s always closed`

audrium: Ah, perfect, thanks for the explanation, somehow missed the fact that you used `with` statement

audrium:
> > Outside this PR: have you considered some protection against zip bombs? I read there are some for `ZipFile.extractall()` method already implemented, but to be extra safe we could also check for any zips inside the zip, etc. as mentioned [here](https://stackoverflow.com/questions/1459080/how-can-i-protect-myself-from-a-zip-bomb)
> 
> Nested zip archives are not a problem, given that we do not recursively decompress them, but we might want to check whether the total (uncompressed) size of the files inside the archive is bigger than a given threshold. In any case, we will have full protection against zip bombs when we will fetch and unzip resources inside a sandbox. Do you have in mind any other protection mechanisms that would be nice to have?

Yeah, it's true that we don't decompress recursively, so we should be fine in this case. Another possible threat is these kind of [zipbombs](https://stackoverflow.com/a/1459820) - in this example 10 Kb zip file will extract a 10 Mb file. Of course this is just for illustration, in practice the numbers could be much bigger. In principle [ZipFile.infolist()](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.infolist) should return information about the file size of each file and as you mentioned we could prevent decompression if the total size of the files is bigger than some threshold. But I agree that we could address these issues later once we will have a sandbox environment for this.

audrium (approved):
Works nicely!

Issue #456, opened 2022-03-15
launcher: fix GitHub CWL usage example


tiborsimko:
Consider the following launcher usage example:

- `https://0.0.0.0:30443/launch?url=https://github.com/reanahub/reana-demo-root6-roofit/blob/master/reana-cwl.yaml`

It leads to:

![c](https://user-images.githubusercontent.com/517546/158360004-cde2f292-0e86-4dab-bd2d-1dee5869827f.png)

PR #457, opened 2022-03-15
rest: add `spec` parameter to launch endpoint

### LINKED ISSUE #452: launcher: implement one-repo-multiple-analyses use case
Some GitHub repositories contain multiple analyses, note HWW, HZZ etc in:

- https://github.com/artfisica/atlas-outreach-cpp-framework-13tev/tree/1.0/Analysis

Note the common "Plotting" step in the directory above:

- https://github.com/artfisica/atlas-outreach-cpp-framework-13tev/tree/1.0/Plotting

This repository would therefore need to specify several `reana-HWW.yaml`, `reana-HZZ.yaml` at the top level.

The labels would then read "Launch HWW analysis on REANA", "Launch HZZ analysis on REANA", etc in the top-level README file.

We should check whether launcher's URL is enough to cover this use case and/or whether we need to introduce a new `spec` variable specifying the desired YAML.

Let's take our standard example repository where we also have several YAML files:

- https://github.com/reanahub/reana-demo-root6-roofit

Would `url=https://github.com/reanahub/reana-demo-root6-roofit/blob/master/reana-cwl.yaml` work well with the launcher?

Currently I'm getting an error launching this URL:

```
Command '['cwltool', '--pack', '--quiet', 'workflow/cwl/workflow.cwl']' returned non-zero exit status 1.
```

Let's investigate and:

- either recognise above URLs pointing-to-some-concrete-yaml-on-github and amend workflow spec fetcher to recognise these URLs to do both (i) git cloning and (ii) setting up non-default `reana-cwl.yaml` for workflow creation and execution logic;
- or keep URLs to point only to github-home-or-branch and amend REST API protocol to allow specifying another `spec` argument such as `url=...&spec=reana-cwl.yaml` to point to any wanted file as the desired REANA spec. 

Slight preference for the latter as it could be useful also for pointing to desired files inside zip tarballs.

mdonadoni:
Closes #452

How to test:
Launch some of the example workflows, choosing different workflow specifications, for example:
```
/launch?url=https://zenodo.org/record/5752285/files/circular-health-data-processing-master.zip
/launch?spec=reana.yaml&url=https://zenodo.org/record/5752285/files/circular-health-data-processing-master.zip
/launch?url=https://github.com/reanahub/reana-demo-root6-roofit
/launch?spec=reana.yaml&url=https://github.com/reanahub/reana-demo-root6-roofit
/launch?spec=reana-cwl.yaml&url=https://github.com/reanahub/reana-demo-root6-roofit
```
Note that the last example will fail due to #456

### DIFF
--- a/reana_server/fetcher.py
+++ b/reana_server/fetcher.py
@@ -193,7 +193,9 @@ class WorkflowFetcherBase(ABC):
             if not self._is_path_inside_output_dir(spec_path):
                 raise REANAFetcherError("Invalid path to the workflow specification")
             if not os.path.isfile(spec_path):
-                raise REANAFetcherError("Cannot find provided workflow specification")
+                raise REANAFetcherError(
+                    "Cannot find the provided workflow specification"
+                )
             return spec_path
 
         specs = [os.path.abspath(path) for path in self._discover_workflow_specs()]
@@ -289,13 +291,16 @@ class WorkflowFetcherYaml(WorkflowFetcherBase):
 class WorkflowFetcherZip(WorkflowFetcherBase):
     """Fetch the specification of a workflow from a zip archive."""
 
-    def __init__(self, parsed_url: ParsedUrl, output_dir: str):
+    def __init__(
+        self, parsed_url: ParsedUrl, output_dir: str, spec: Optional[str] = None
+    ):
         """Initialize the workflow specification fetcher.
 
         :param parsed_url: Parsed URL of the workflow specification to fetch.
         :param output_dir: Directory where all the data will be saved to.
+        :param spec: Optional path to the workflow specification.
         """
-        super().__init__(parsed_url, output_dir)
+        super().__init__(parsed_url, output_dir, spec)
         self._archive_name = self._parsed_url.basename
 
     def fetch(self) -> None:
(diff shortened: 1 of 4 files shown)


reana_server/fetcher.py
```diff
@@ -364,38 +372,52 @@ def _get_github_fetcher(parsed_url: ParsedUrl, output_dir: str) -> WorkflowFetch
...
+    if spec:
+        _, spec_ext = os.path.splitext(spec)
+        if spec_ext not in WORKFLOW_SPEC_EXTENSIONS:
+            raise ValueError(
+                "The provided specification doesn't have a valid file extension"
+            )
+
     if parsed_url.extension == ".git":
-        return WorkflowFetcherGit(parsed_url, output_dir)
+        return WorkflowFetcherGit(parsed_url, output_dir, spec=spec)
```
audrium, line 410: Extremely minor: is there a reason not to pass `spec` as a positional argument here like for the other classes?

audrium (approved):
Nicely done, works as expected!

reana_server/fetcher.py
mdonadoni: The third parameter of the initializer of `WorkflowFetcherGit` is an optional reference to a Git branch/commit (`git_ref`). Should I swap `git_ref` and `spec` so that the initialization of `WorkflowFetcherGit` is consistent with the initialization of the other types of fetchers?

audrium: I see, I've missed that there is additional `git_ref` parameter in between `spec` and others for `WorkflowFetcherGit`.. I think there is no need to change anything here, thanks for explaining

Issue #455, opened 2022-03-14
API rate limiter: prettify error messages


audrium:
Currently API rate limiter error messages are not easily understandable by the users. Improving it for both Web UI users and CLI UI users would be really nice to have.

If user hits Launch on REANA web page twice in 5 seconds, this kind of error will be displayed:
![](https://user-images.githubusercontent.com/4990025/157864863-5016eaba-bb68-45ed-a714-50319e0f94f9.png)

See also https://github.com/reanahub/reana/issues/286 how the rate limiter error message looks like in regular reana-client usage scenario if user has many input files.

_Originally posted in https://github.com/reanahub/reana/pull/619#discussion_r824660035_ and https://github.com/reanahub/reana/pull/619#discussion_r825728316

PR #458, opened 2022-03-15
api: improve api rate limiter error messages

audrium:
closes https://github.com/reanahub/reana-server/issues/455
closes https://github.com/reanahub/reana-ui/issues/241

- created new `handle_rate_limit_error()` error handler to return more verbose error messages in case rate limiter returns an error.
- moved `handle_validation_error()` from a blueprint to app level to handle these kind of errors for other endpoints as well.
- added `JSONDecodeError` exception handling for `launch` endpoint to improve the error message for the invalid format of workflow `parameters`

To test:
- modify rate limiter values for `REANA_RATELIMIT_AUTHENTICATED_USER` and/or `REANA_RATELIMIT_SLOW` to something more strict like: `1/30 second`
- observe that the error messages are now more verbose - `Too many requests in a given amount of time. Only "1 per 5 second" allowed.`
- check other endpoints (e.g. `get_workflows()`) to also use `handle_validation_error()` for handling error messages
- check https://github.com/reanahub/reana-ui/issues/241#issuecomment-1067800198 case with wrong `parameters` format to return more verbose error message


reana_server/rest/launch.py
```diff
@@ -186,6 +163,9 @@ def launch(user, url, name="", parameters="{}", spec=None):
     except HTTPError as e:
         logging.error(traceback.format_exc())
         return jsonify(e.response.json()), e.response.status_code
+    except json.JSONDecodeError:
+        logging.error(traceback.format_exc())
+        return jsonify({"message": "Parameters contains not valid JSON."}), 400
```
mdonadoni, line 168: Minor: what do you think about using `The 'parameters' field is not valid JSON`? This makes it clear that `parameters` refers to that specific field and not to the parameters of `/launch` in general

mdonadoni (approved):
LGTM! I left just one small comment

reana_server/rest/launch.py
mvidalgarcia: This naming problem kept me reflecting from time to time when developing the UI part. What I've done in the end was call them "workflow parameters" to differentiate them from the query string parameters. @mdonadoni suggestion is very valid and can be completed by prepending "workflow" before `'parameters'`: `The workflow 'parameters' field is not valid JSON`. Let me know what you think.

audrium: Thank you both for a good remark, fixed it accordingly

PR #464, opened 2022-03-18
fetcher: improve workflow naming for zip archives from GitHub

### LINKED ISSUE #453: fetcher: include repository's name into generated workflow name in github zip url case
Right now, when downloading zip files we do not parse the GitHub URL, so it's hard to get the repository's name to include it into the generated workflow name.

Example:
https://localhost:30443/launch?url=https://github.com/reanahub/reana-demo-root6-roofit/archive/refs/heads/master.zip
Fetcher generates a workflow name called `master`.

This issue could be tackled when we will refactor the fetcher (see [this comment](https://github.com/reanahub/reana-server/pull/448#discussion_r822530150)). 
_Originally posted by @mdonadoni in https://github.com/reanahub/reana-server/pull/451#discussion_r825801475_

mdonadoni:
Closes #453

How to test:
Launch the following workflows
- https://github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/archive/refs/heads/2012.zip
- https://github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/archive/refs/tags/v1.3.zip
- https://github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/archive/4eb69e073d3b2567c49b2525c2a53341a16dc1ab.zip

Expected result: the names of the workflows launched contain the name of the repo

### DIFF
--- a/reana_server/fetcher.py
+++ b/reana_server/fetcher.py
@@ -18,6 +18,8 @@ import zipfile
 from git import Repo
 import requests
 from requests.exceptions import HTTPError, Timeout, RequestException
+import werkzeug.exceptions
+from werkzeug.routing import Map, Rule
 
 from reana_server.config import (
     FETCHER_ALLOWED_SCHEMES,
@@ -295,16 +297,27 @@ class WorkflowFetcherZip(WorkflowFetcherBase):
     """Fetch the specification of a workflow from a zip archive."""
 
     def __init__(
-        self, parsed_url: ParsedUrl, output_dir: str, spec: Optional[str] = None
+        self,
+        parsed_url: ParsedUrl,
+        output_dir: str,
+        spec: Optional[str] = None,
+        workflow_name: Optional[str] = None,
     ):
         """Initialize the workflow specification fetcher.
 
         :param parsed_url: Parsed URL of the workflow specification to fetch.
         :param output_dir: Directory where all the data will be saved to.
         :param spec: Optional path to the workflow specification.
+        :param workflow_name: Workflow name that overrides the workflow name generation.
         """
         super().__init__(parsed_url, output_dir, spec)
         self._archive_name = self._parsed_url.basename
+        if workflow_name:
+            self._workflow_name = self._clean_workflow_name(workflow_name)
+        else:
+            self._workflow_name = self._clean_workflow_name(
+                self._parsed_url.basename_without_extension
+            )
 
     def fetch(self) -> None:
         """Fetch workflow specification from a zip archive."""
@@ -331,11 +344,12 @@ class WorkflowFetcherZip(WorkflowFetcherBase):
     def generate_workflow_name(self) -> str:
         """Generate a workflow name from the given URL to the zip archive.
 
-        The name of the zip archive is used as the name of the workflow.
+        The name of the zip archive is used as the name of the workflow, unless a custom
+        workflow name was specified when initializing the fetcher.
 
         :returns: Generated workflow name.
         """
-        return self._clean_workflow_name(self._parsed_url.basename_without_extension)
+        return self._workflow_name
 
 
 def _get_github_fetcher(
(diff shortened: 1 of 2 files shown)


VMois:
Works as expected.

*minor:* When a commit is referenced, the name because very long, and breaks UI layout. We can have maybe a shorter name when commit references are used. Not sure if it is better to tweak UI, or reduce the commit reference to a few first characters.

Screenshot 2022-03-21 at 13 38 40

VMois:
Works as expected.

*minor:* When a commit is referenced, the name because very long, and breaks UI layout. We can have maybe a shorter name when commit references are used. Not sure if it is better to tweak UI, or reduce the commit reference to a few first characters.

Screenshot 2022-03-21 at 13 38 40

reana_server/fetcher.py
```diff
@@ -331,10 +341,13 @@ def fetch(self) -> None:
     def generate_workflow_name(self) -> str:
         """Generate a workflow name from the given URL to the zip archive.
 
-        The name of the zip archive is used as the name of the workflow.
+        The name of the zip archive is used as the name of the workflow, unless a custom
+        workflow name was specified when initiliazing the fetcher.
```
VMois, line 345:
```suggestion
        workflow name was specified when initializing the fetcher.
```

reana_server/fetcher.py
```diff
@@ -295,16 +297,24 @@ class WorkflowFetcherZip(WorkflowFetcherBase):
...
         :param parsed_url: Parsed URL of the workflow specification to fetch.
         :param output_dir: Directory where all the data will be saved to.
         :param spec: Optional path to the workflow specification.
+        :param workflow_name: Workflow name that overrides the workflow name generation.
         """
         super().__init__(parsed_url, output_dir, spec)
         self._archive_name = self._parsed_url.basename
+        self._workflow_name = None
+        if workflow_name:
+            self._workflow_name = self._clean_workflow_name(workflow_name)
```
VMois, line 317: *suggestion:* To have cleaning of workflow name in one place maybe it is better to move this part to `generate_workflow_name` method.

```python
if workflow_name:
    return self._clean_workflow_name(workflow_name)
return self._clean_workflow_name(self._parsed_url.basename_without_extension)
```

Or even better, in `init`:

```python
if workflow_name:
    self._workflow_name = self._clean_workflow_name(workflow_name)
else:
    self._workflow_name = self._clean_workflow_name(self._parsed_url.basename_without_extension)
```

mdonadoni:
> _minor:_ When a commit is referenced, the name because very long, and breaks UI layout. We can have maybe a shorter name when commit references are used. Not sure if it is better to tweak UI, or reduce the commit reference to a few first characters.

I think it is better to fix this issue directly in the UI, given that the name of a workflow can be chosen by the user and it can be very long (maximum length of 255 characters,  see [here](https://github.com/reanahub/reana-db/blob/master/reana_db/models.py#L443)).

What do you think?

VMois:
> I think it is better to fix this issue directly in the UI, given that the name of a workflow can be chosen by the user and it can be very long (maximum length of 255 characters, see [here](https://github.com/reanahub/reana-db/blob/master/reana_db/models.py#L443)). What do you think?

Yeah, agree. I didn't want to fix it in this PR anyway. Will create a new issue for UI.

mdonadoni:
> Yeah, agree. I didn't want to fix it in this PR anyway. Will create a new issue for UI.

This issue is already fixed in https://github.com/reanahub/reana-ui/pull/246 :tada:

VMois (approved):

reana_server/fetcher.py
```diff
@@ -409,12 +421,12 @@ def get_fetcher(
                 "The provided specification doesn't have a valid file extension"
             )
 
-    if parsed_url.extension == ".git":
+    if parsed_url.netloc == "github.com":
+        return _get_github_fetcher(parsed_url, output_dir, spec)
+    elif parsed_url.extension == ".git":
```
audrium, line 426: Just a quick question: is there still a use case for having this check? If we take an example of this url (`https://github.com/reanahub/reana-demo-root6-roofit.git`), in this case `_get_github_fetcher()` would be called. So is there a case in which `WorkflowFetcherGit()` would be returned here while `_get_github_fetcher()` would not being called?

VMois: I guess other repositories, for example, from GitLab.

mdonadoni: This check is still useful when dealing with Git repositories that are not on GitHub (e.g GitLab). However, in this case it is currently not possible to specify a particular commit/tag/branch to use, so perhaps we should improve this in the future.

audrium: Thanks, makes sense

PR #469, opened 2022-03-28
scheduler: improve logging while checking cluster readiness

audrium:


reana_server/scheduler.py
```diff
@@ -83,6 +92,10 @@ def doesnt_exceed_max_reana_workflow_count():
             .scalar()
         )
         if running_workflows >= REANA_MAX_CONCURRENT_BATCH_WORKFLOWS:
+            logging.info(
+                "Workflow could not be started due to reached upper limit on "
```
tiborsimko, line 96: We can print the detected values for easier deduction, such as:

Cannot start workflow 60515eff-33ca-468a-a55c-6cf72d72af5d because there are already 42 running workflows whilst the allowed maximum is 25.

reana_server/scheduler.py
```diff
@@ -60,6 +66,9 @@ def check_predefined_conditions():
             #              DiskPressure, PIDPressure,
             for condition in node.get("status", {}).get("conditions", {}):
                 if not condition.get("status"):
+                    logging.info(
+                        f"Workflow could not be started due to predefined condition: {condition}"
```
tiborsimko, line 70: Cannot start workflow 60515eff-33ca-468a-a55c-6cf72d72af5d due to predefined condition foo, bar and baz.

reana_server/scheduler.py
```diff
@@ -43,7 +43,13 @@ def check_memory_availability(workflow_min_job_memory):
     if not nodes:
         return True
     max_node_available_memory = max(nodes)
-    return max_node_available_memory >= workflow_min_job_memory
+    available = max_node_available_memory >= workflow_min_job_memory
+    if not available:
+        logging.info(
+            f"Workflow could not be started due to low memory. Available memory: "
```
tiborsimko, line 49: Cannot start workflow 60515eff-33ca-468a-a55c-6cf72d72af5d because it requires 12 GiB memory whilst only 8 GiB is available.

reana_server/scheduler.py
```diff
@@ -33,22 +33,39 @@
...
     REANA_SCHEDULER_REQUEUE_COUNT,
+    REANA_WORKFLOW_SCHEDULING_READINESS_CHECK,
 )
 from reana_server.status import NodesStatus
 
 
-def check_memory_availability(workflow_min_job_memory):
+def check_memory_availability(
+    workflow_min_job_memory: Optional[float],
+) -> Tuple[bool, Optional[str]]:
```
mdonadoni, line 43: What do you think about returning an `Optional[str]` instead of a tuple? `None` would mean that there are no errors, while a `str` would represent the error message.

In `reana_ready`:
```python
for check_condition in conditions:
    error = check_condition()
    if error is not None:
        return error
```

reana_server/scheduler.py
```diff
@@ -83,29 +103,44 @@ def doesnt_exceed_max_reana_workflow_count():
...
 
 
-def reana_ready(workflow_min_job_memory):
+def reana_ready(workflow_min_job_memory: Optional[float]) -> Tuple[bool, Optional[str]]:
     """Check if REANA can start new workflows."""
-    conditions = [check_predefined_conditions, doesnt_exceed_max_reana_workflow_count]
+    conditions_list = [
+        doesnt_exceed_max_reana_workflow_count,
+        partial(check_memory_availability, workflow_min_job_memory)
+        # check_predefined_conditions,
```
mdonadoni, line 125: Is this supposed to be commented? If we want to remove this check, I think it's better to just delete the whole function.

reana_server/scheduler.py
audrium: Thanks for a nice suggestion

reana_server/scheduler.py
audrium: I was hesitating to remove it in the beginning, since it was there for 3 years.. But indeed because of the reasons I explained in the MM I think it's better to get rid of this check

reana_server/scheduler.py
```diff
@@ -83,29 +103,44 @@ def doesnt_exceed_max_reana_workflow_count():
...
 
-    # Do not calculate memory availability on fifo strategy
-    if workflow_min_job_memory:
-        conditions.append(partial(check_memory_availability, workflow_min_job_memory))
+    conditions_map = {
+        0: [],
+        1: conditions_list[0],
+        2: conditions_list[1],
+        9: conditions_list,
+    }
```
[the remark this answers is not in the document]
audrium: In the end I did something similar to what you suggested and moved hardcoded value mapping to the `configs.py`. The tricky part here is that `check_memory_availability` needs an argument and we need to create a `partial` function for that. That's why this map could not be easily moved to the configs (also there would probably be circular imports error)

PR #473, opened 2022-04-04
rest: improve error messages in `/launch` endpoint

### LINKED ISSUE #472: fetcher: improve error messages
Currently, the following error is shown when launching a workflow using an invalid GitHub URL

screenshot

We should improve the error messages shown to the user when something goes wrong while fetching a remote workflow specification. It might also be a good idea to return a generic error [here](https://github.com/reanahub/reana-server/blob/a4c7f1b11bcab52f8a5fd9ba899b51bdfa9978dc/reana_server/rest/launch.py#L196), instead of the error message of the exception.

mdonadoni:
Closes #472

How to test:
Launch the following workflows
```
https://github.com/reanahub/reana-demo-root6-roofit/tree/invalid
https://github.com/reanahub/invalid
```
Expected result:
The workflows are not launched and the error messages shown to the user are short and user-friendly


VMois (approved):
Works as expected.

PR #477, opened 2022-04-06
setup: upgrade invenio dependencies

audrium:
closes https://github.com/reanahub/reana-server/issues/479

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,7 @@ Version 0.9.0 (UNRELEASED)
 - Adds REANA specification validation utilities.
 - Changes API rate limiter error messages to be more verbose.
 - Changes workflow scheduler to allow defining the checks needed to assess whether the cluster can start new workflows.
+- Changes the Invenio dependencies to the latest versions.
 
 Version 0.8.4 (2022-02-23)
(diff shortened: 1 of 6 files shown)


CHANGES.rst
```diff
@@ -10,6 +10,7 @@ Version 0.9.0 (UNRELEASED)
 - Adds REANA specification validation utilities.
 - Changes API rate limiter error messages to be more verbose.
 - Changes workflow scheduler to allow defining the checks needed to assess whether the cluster can start new workflows.
+- Changes to the latest versions of Invenio dependencies.
```
mdonadoni, line 13: What do you think about `Updates the Invenio dependencies to the latest versions`?

mdonadoni (approved):
LGTM! 🎉

mdonadoni:
There is also a little typo in the commit description: `#477` should be `#479`

CHANGES.rst
audrium: I agree that your suggestion sounds better, but there is a convention for the changelog to only use `changes/adds/fixes` as a prefix. @tiborsimko, I guess we should document it in REANA wiki?

PR #484, opened 2022-04-11
fetcher: add support for GitLab URLs

### LINKED ISSUE #483: launch: support launching workflows from GitLab repositories
**Current behaviour**

Launcher does not support launching workflows from GitLab repositories:

gl

**Expected behaviour**

GitLab should be supported just as well as GitHub.

**Notes**

Some repositories on GitLab may be fully public, some may be private (but in this case `reana` usually has `Reporter` access to them).

mdonadoni:
Closes #483

How to test:
- Create a new public repository and push a REANA demo to [gitlab.cern.ch](https://gitlab.cern.ch)
- Launch some workflows using the following URLs:
  - URL to the repository
  - URL to a specific tag/branch/commit
  - URL to the repository's zip archive

Expected result: The workflows are launched and executed correctly

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -274,3 +274,6 @@ FETCHER_ALLOWED_SCHEMES = ["https", "http"]
 
 FETCHER_REQUEST_TIMEOUT = 60
 """Timeout used when fetching workflow specifications."""
+
+FETCHER_ALLOWED_GITLAB_HOSTNAMES = ["gitlab.com", "gitlab.cern.ch"]
+"""GitLab instances allowed when fetching workflow specifications."""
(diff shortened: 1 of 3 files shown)


audrium (approved):
Great job, it works nicely!
One minor observation: if user tries to launch a private gitlab repo now the error message is a bit generic - `Cannot clone the given Git repository`, we could perhaps expand it a bit in this case to give a bit more context to the user.

mdonadoni:
> Great job, it works nicely! One minor observation: if user tries to launch a private gitlab repo now the error message is a bit generic - `Cannot clone the given Git repository`, we could perhaps expand it a bit in this case to give a bit more context to the user.

Done!

Issue #462, opened 2022-03-17
launcher: GitHub clone, directory is not found error


VMois:
No clear steps to reproduce a bug, but two developers already encountered it on a local installation of REANA.

Details:

![Screenshot 2022-03-17 at 13 19 23](https://user-images.githubusercontent.com/15694232/158816468-92a36a21-97b0-4bdf-adfd-80d45e1736d7.png)

Suggested by @audrium:

The problem seems to be coming from [this place](https://github.com/reanahub/reana-server/blob/master/reana_server/rest/launch.py#L125). Somehow `tmpdir` sometimes is not created because of the permission problems, could be umask issues or something like that.

tiborsimko:
I saw it in the past on LOCALHOST, where after letting system be idle for several minutes, every second launch led to the above problem. (Could be connected to kind volume mounting? Or uwsgi setup?) 

I cannot reproduce it on DEV though (using CephFS).

Working with:

```console
$ reana-dev git-status -c CLIENT -c CLUSTER -c reana | grep -v '@ master'
reana-workflow-controller @ pr-406 @ f07e502 workflow_run_manager: fix workspace group problem
reana-server @ pr-461 @ 06d773f rest: specify workflow path while loading spec in launch endpoint
reana-commons @ pr-342 @ f90a33d spec: improve spec loading functionality to support remote working directories
reana-ui @ pr-246 @ 84d64a9 wf-list: fix overflow when workflow names are long
```

tiborsimko:
I cannot reproduce on LOCALHOST anymore... Ditto @mdonadoni. I think we can close this, proceed with releasing alpha, and reopen the issue for further investigation should it reappear in the future.

PR #486, opened 2022-04-26
rest: load one specification at a time in `/launch`

mdonadoni:
Depends on reanahub/reana-commons#347
Related to #462

How to test:
Try to reproduce #462 by launching many workflows using Snakemake

### DIFF
--- a/reana_server/rest/launch.py
+++ b/reana_server/rest/launch.py
@@ -8,6 +8,9 @@
 
 import json
 import logging
+import os
+import shutil
+import threading
 import traceback
 
 from bravado.exception import HTTPError
@@ -36,12 +39,14 @@ from reana_server.utils import (
     mv_workflow_files,
     prevent_disk_quota_excess,
     publish_workflow_submission,
-    remove_fetched_workflows_dir,
 )
 from reana_server.validation import validate_workflow
 
 
 blueprint = Blueprint("launch", __name__)
+
+load_reana_spec_lock = threading.Lock()
+"""Lock used to make sure only one specification is loaded at a time."""
 
 
 @blueprint.route("/launch", methods=["POST"])
@@ -147,7 +152,10 @@ def launch(user, url, name="", parameters="{}", specification=None):
 
         # Load and validate the workflow spec
         spec_path = fetcher.workflow_spec_path()
-        reana_yaml = load_reana_spec(spec_path, workspace_path=tmpdir)
+        # FIXME: locking will not be needed when the loading and validation of
+        # specifications will be done inside an external sandbox
+        with load_reana_spec_lock:
+            reana_yaml = load_reana_spec(spec_path, workspace_path=tmpdir)
         input_parameters = json.loads(parameters)
         validate_workflow(reana_yaml, input_parameters)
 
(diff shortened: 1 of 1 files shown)


reana_server/rest/launch.py
```diff
@@ -147,7 +152,10 @@ def launch(user, url, name="", parameters="{}", specification=None):
 
         # Load and validate the workflow spec
         spec_path = fetcher.workflow_spec_path()
-        reana_yaml = load_reana_spec(spec_path, workspace_path=tmpdir)
+        # FIXME: locking will not be needed when the loading and validation of
+        # specifications will be done inside an external sandbox
+        with load_reana_spec_lock:
```
tiborsimko, line 157: Could this lead to situations with lock being stalled and never released, which could block further requests? I guess we can simply restart `r-server` pod when this happens, which is acceptable... but I was wondering whether we could do some gentle "auto-release-when-stalled" sort of handling here? (Not necessary to implement for the merge.)

reana_server/rest/launch.py
```diff
@@ -198,7 +206,18 @@ def launch(user, url, name="", parameters="{}", specification=None):
...
     finally:
-        remove_fetched_workflows_dir(tmpdir)
+        # FIXME: `load_reana_spec` is not thread-safe and it changes the cwd, so for the
+        # time being we can only delete the files inside the directory where the
+        # workflow is fetched.  Deleting the directory would result in a
+        # `FileNotFoundError` when calling `os.getcwd()`, due to the cwd not existing
+        # anymore.
+        #
+        # remove_fetched_workflows_dir(tmpdir)
+        for entry in os.scandir(tmpdir):
```
tiborsimko, line 216: I guess it should be OK to iterate through everything and delete one by one, since it is probable that there will be only a few spec files fetched there... But I'm wondering out loud whether this may become slow in certain situations (shared disk busy!) which could contribute to endpoint responsiveness troubles.

Perhaps we can make a length check to see how many files would be deleted, and do the deletions "live" only when there are less than N files? Or delete via one command from the root, such as `rm -rf ./workflowfetched/XXXX/*` in one fast operation?

Ideally, the REST API endpoint should be very fast to reply, so it'll be good not to do any potentially lengthy stuff "synchronously" there, and postpone deletions for overnight and do them "asynchronously". E.g. the REST API endpoint could simply tag this directory as having been submitted e.g. by leaving a file named "LAUNCHED.txt" or something in there, and then the nightly quota updater job would go through all user's `workflowfetched/XXXX` directories and would remove those that were launched already...

Could this "theoretical" concern be "potentially real" based on your scalability tests?

reana_server/rest/launch.py
mdonadoni: > Could this lead to situations with lock being stalled and never released, which could block further requests?

Yes, it might happen if `load_reana_spec` stalls or takes a long time to complete.  Putting a limit on the running time of `load_spec_reana` is for sure a good idea and it should be easy to do when we will load the specification in an external sandbox. Right now, I don't think we have any mechanism to control how much time `load_reana_spec` takes.

reana_server/rest/launch.py
mdonadoni: If the workflow is successfully launched, then the directory will be empty given that the files are moved from the temporary directory to the workflow's workspace (see [here](https://github.com/reanahub/reana-server/blob/4e8546726881ee6f525434e4089fc99d72d9967c/reana_server/rest/launch.py#L167)). In this case, we are already listing the contents of the directory and moving each entry to the workflow's workspace (see [here](https://github.com/reanahub/reana-server/blob/4e8546726881ee6f525434e4089fc99d72d9967c/reana_server/utils.py#L135)).

Should I investigate whether listing the contents of a directory can be a problem or should we open an issue about it?

tiborsimko: Since we are going to work on sandboxing soon, I think we can try in the current format and see in practice whether the shared disk slowness wouldn't be a noticeable problem, and come back to this directory tagging etc if necessary.

tiborsimko (approved):
Improves the situation nicely :+1:

PR #487, opened 2022-04-28
admin: add command to delete messages from queue

VMois:
closes reanahub/reana-message-broker#45

How to test?

0. Unit tests should cover the situation when you specify some `values-to-delete` (using the -v option). So, you should test the interactive way.

1. Bash into r-server `kubectl exec -it deployment/reana-server -- bash`. Run `flask reana-admin consume-queue -q workflow-submission -i`. This command will start listening to messages from the `workflow-submission` queue.

2. Submit some workflows. You will see our consumer picks up some messages (because the scheduler is working at the same time). Use

### DIFF
--- a/reana_server/reana_admin/__init__.py
+++ b/reana_server/reana_admin/__init__.py
@@ -1,0 +1,11 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2022 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA-Server admin tools."""
+
+from .cli import reana_admin
--- a/reana_server/reana_admin/cli.py
+++ b/reana_server/reana_admin/cli.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2020, 2021 CERN.
+# Copyright (C) 2020, 2021, 2022 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -13,6 +13,7 @@ import logging
 import secrets
 import sys
 import traceback
+from typing import List, Optional
 
 import click
 import tablib
@@ -639,3 +640,79 @@ def set_default_quota_limit(ctx):
             resource_type=resource.type_.name,
             limit=DEFAULT_QUOTA_LIMITS.get(resource.type_.name),
         )
+
+
+@reana_admin.command("queue-consume")
+@click.option(
+    "--queue-name",
+    "-q",
+    required=True,
+    type=str,
+    help="Name of the queue that will be consumed, e.g workflow-submission",
+)
+@click.option(
+    "--key",
+    "-k",
+    type=str,
+    help="Key of the property that will be used to filter the messages in the queue, e.g workflow_name_or_id",
+)
+@click.option(
+    "--values-to-delete",
+    "-v",
+    multiple=True,
+    help="List of property values used to filter messages that will be removed from the queue, e.g UUID of a workflow",
+)
+@click.option(
+    "-i",
+    "--interactive",
+    is_flag=True,
+    default=False,
+    help="Manually decide which messages to remove from the queue.",
+)
+def queue_consume(
+    queue_name: str,
+    key: Optional[str],
+    values_to_delete: List[str],
+    interactive: bool,
+):
+    """Start consuming specified queue and remove selected messages.
+
+    By default, you will need to specify either "-k" or "-i" options otherwise the command will return an error.
+
+    If -k option is specified, messages that have property values specified in -v will be deleted.
+
+    If -i option is specified, for every message, user will be asked what to do.
+
+    If -k and -i are specified together, for every message that matches property values in -v, user will be asked whether to delete it or not.
+    """
+    from reana_server.reana_admin.consumer import MessageConsumer
+
+    if key is None and not interactive:
+        click.secho(
+            "Please provide -k (with -v) or -i options. These options can be used together or separately.",
+            fg="red",
+        )
+        sys.exit(1)
+
+    if key and not values_to_delete:
+        click.secho(
+            f"Please provide a list of property values (using the '-v' option) to filter the messages that need to be removed from the {queue_name} queue.",
+            fg="red",
+        )
+        sys.exit(1)
+
+    try:
+        consumer = MessageConsumer(
+            queue_name=queue_name,
+            key=key,
+            values_to_delete=list(values_to_delete),
+            is_interactive=interactive,
+        )
+    except Exception as error:
+        click.secho(
+            "Error is raised during MessageConsumer initialization. Please, check if arguments are correct.",
+            fg="red",
+        )
+        logging.exception(error)
+    else:
+        consumer.run()
--- a/reana_server/reana_admin/consumer.py
+++ b/reana_server/reana_admin/consumer.py
@@ -1,0 +1,116 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2022 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA Admin MessageConsumer."""
+
+import json
+from enum import Enum, auto
+from typing import List, Optional
+
+import click
+
+from reana_commons.consumer import BaseConsumer
+
+
+class UserDecision(Enum):
+    """Possible user decisions in interactive consumer mode."""
+
+    DELETE_MESSAGE = auto()
+    KEEP_MESSAGE = auto()
+    STOP_CONSUMER = auto()
+
+
+class MessageConsumer(BaseConsumer):
+    """Consumer responsible for cleaning queues."""
+
+    def __init__(
+        self,
+        queue_name: str,
+        key: Optional[str],
+        values_to_delete: List[str],
+        is_interactive: bool = False,
+        **kwargs,
+    ):
+        """Initialise the class."""
+        super(MessageConsumer, self).__init__(queue=queue_name, **kwargs)
+        self.key = key
+        self.values_to_delete = values_to_delete
+        self.is_interactive = is_interactive
+
+        click.secho(
+            "MessageConsumer initialized with the following settings: \n"
+            f"  - queue_name: {queue_name}\n"
+            f"  - message_key: {self.key}\n"
+            f"  - values_to_delete: {self.values_to_delete}\n"
+            f"  - interactive: {self.is_interactive}"
+        )
+
+    def get_consumers(self, Consumer, channel):
+        """Implement providing kombu.Consumers with queues/callbacks."""
+        return [
+            Consumer(
+                queues=self.queue,
+                callbacks=[self.on_message],
+                accept=[self.message_default_format],
+                prefetch_count=1,
+            )
+        ]
+
+    def on_consume_ready(self, connection, channel, consumers, **kwargs):  # noqa: D102
+        click.secho(f"Starting consuming the {self.queue.name} queue", fg="green")
+        click.secho("If you want to stop the consumer, use Ctrl+C.\n", fg="yellow")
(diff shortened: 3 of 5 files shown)


tests/test_cli.py
```diff
@@ -254,3 +256,54 @@ def test_revoke_token(default_user, session):
...
+            messages_to_delete=["some_other_name"],
+        )
+        in_memory_wsp = WorkflowSubmissionPublisher(
+            connection=in_memory_queue_connection
+        )
+
+        in_memory_wsp.publish_workflow_submission("1", workflow_name, {})
+        consume_queue(consumer, limit=1)
+        assert not in_memory_queue_connection.channel().queues[queue_name].empty()
+        in_memory_queue_connection.channel().queues.clear()
```
VMois, line 284: *issue (non-blocking):* if in-memory queues are not cleared tests fail

The test fails for both MessageConsumer and scheduler. This is probably due to [in_memory_queue_connection having a "session" scope](https://github.com/reanahub/pytest-reana/blob/0055f5c72d55bdabe0e5c68bb714f5f177f95f18/pytest_reana/fixtures.py#L626). It means some messages can stay in the queue between tests and cause unexpected failures.

As for now, I solved it by cleaning messages at the end of each test. But, this is not a sustainable solution. It is not nice when you need to remember to clear stuff after the test. I propose to create an issue to address later.

WDYT?

reana_server/reana_admin/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+#
+# This file is part of REANA.
+# Copyright (C) 2022 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA-Server admin tools."""
+
+from .main import reana_admin
```
mdonadoni, line 11: Maybe we can call the file `cli.py` instead of `main.py`, like we do for [`reana-dev`](https://github.com/reanahub/reana/blob/7bddf3f23bb7afba3a363e0413d1ca6414e2b214/reana/reana_dev/cli.py)?

reana_server/reana_admin/consumer.py
```diff
@@ -0,0 +1,90 @@
...
+        should_remove_message = False
+
+        if self.is_interactive:
+            answer = input(
+                "Delete the above message? (Y - Yes, N - No, S - Stop consuming): "
+            )
+            if answer.startswith("Y"):
+                should_remove_message = True
+            elif answer.startswith("S"):
+                self.should_stop = True
```
mdonadoni, line 78: We should probably improve this a little bit: right now, if a user type a lowercase `s` or `y`, the answer is simply ignored and considered as a `No`. In particular:
1. If the command receives an invalid answer, it should ask again to the user what to do
2. It should accept both lowercase and uppercase answers
3. It should also accept `yes`/`no`/`stop` as an answer

One way to do it: `answer.lower() in ["yes", "y"]`

What do you think?

reana_server/reana_admin/main.py
```diff
@@ -639,3 +640,61 @@ def set_default_quota_limit(ctx):
...
         )
+
+
+@reana_admin.command("consume-queue")
+@click.option(
+    "--queue-name",
+    "-q",
+    required=True,
+    type=str,
+    help="Name of the queue that will be consumed, e.g workflow-submissions",
```
mdonadoni, line 651: typo: `workflow-submissions` -> `workflow-submission`

reana_server/reana_admin/main.py
```diff
@@ -639,3 +640,61 @@ def set_default_quota_limit(ctx):
...
+)
+@click.option(
+    "--key",
+    "-k",
+    type=str,
+    help="Key that will be used to get the id value from the message body, e.g workflow_name_or_id",
```
mdonadoni, line 657: What about `Key of the property that will be used to filter the messages in the queue, e.g. workflow_name_or_id`?

reana_server/reana_admin/main.py
```diff
@@ -639,3 +640,61 @@ def set_default_quota_limit(ctx):
...
+)
+@click.option(
+    "--messages-to-delete",
+    "-m",
+    multiple=True,
+    help="List of message ids to remove from the queue, e.g UUID of a workflow",
+)
```
mdonadoni, line 664: I think it is better if we refer to these as `values` instead of `message ids`. WDYT?

reana_server/reana_admin/main.py
```diff
@@ -639,3 +640,61 @@ def set_default_quota_limit(ctx):
...
+def consume_queue(
+    queue_name: str,
+    key: Optional[str],
+    messages_to_delete: List[str],
+    interactive: bool,
+):
+    """Start consuming specified queue and remove selected messages."""
+    from reana_server.reana_admin.consumer import MessageConsumer
+
+    if key and interactive:
```
mdonadoni, line 681: I think it might be useful to be able to specify both at the same time. In this way, the administrators can filter the messages based on some conditions (e.g. `user` is `xyz`), but they can still decide manually whether to remove each of those message from the queue. The other messages are simply requeued.

reana_server/reana_admin/main.py
```diff
@@ -639,3 +640,61 @@ def set_default_quota_limit(ctx):
...
+            fg="red",
+        )
+        return
+
+    consumer = MessageConsumer(
+        queue_name=queue_name,
+        message_key=key,
+        messages_to_delete=list(messages_to_delete),
+        is_interactive=interactive,
+    )
```
mdonadoni, line 699: I think we should improve the error handling when the queue specified by the user doesn't exist. Right now, the command only shows a stack trace.

tests/test_cli.py
```diff
@@ -254,3 +256,54 @@ def test_revoke_token(default_user, session):
         ],
     )
     assert "does not have an active access token" in result.output
+
+
+class TestMessageConsumer:
```
mdonadoni, line 261: minor: we need to add `2022` to the copyright notice of this file

tests/test_cli.py
mdonadoni: Yes, I think that opening an issue is a good idea!

reana_server/reana_admin/main.py
```diff
@@ -639,3 +640,61 @@ def set_default_quota_limit(ctx):
...
+    help="List of message ids to remove from the queue, e.g UUID of a workflow",
+)
+@click.option(
+    "-i",
+    "--interactive",
+    is_flag=True,
+    default=False,
+    help="Manually decide which messages to remove from the queue.",
+)
+def consume_queue(
```
mdonadoni, line 672: Some of the other `reana-admin` commands require the admin access token to be provided by the user (e.g. [`token-grant`](https://github.com/reanahub/reana-server/blob/a42b1efeb722c19181d6fe016c0e308cbafa134c/reana_server/reana_admin.py#L215)), but not all of them. Should we check it in every command? In this case, we can open an issue. /cc @tiborsimko

mdonadoni (commented):
Nice work! The core functionality works, but the PR needs to be rebased and there are a few comments to address.

One question: if the scheduler is still running when this command executes, some messages are sent directly to the scheduler and are not received by `consume-queue`. Is there something we can do about that?

VMois:
> One question: if the scheduler is still running when this command executes, some messages are sent directly to the scheduler and they are not received by consume-queue. Is there something we can do about that?

This is a good question. I have tried to disable the scheduler but didn't succeed in finding a way. 

It doesn't look like you can stop the container inside the pod without stopping the pod itself. Ideas on how to pause `scheduler` are appreciated.

reana_server/reana_admin/__init__.py
VMois: Really good suggestion. I will change to `cli.py`

reana_server/reana_admin/main.py
VMois: Yes. It is more general and better describes.

Should we still keep `--messages-to-delete`? Maybe we should use `--values-to-delete`? WDYT?

reana_server/reana_admin/main.py
VMois: Can you describe more what it will look like from the administrator's perspective? Like what the input and output of the script will look like.

reana_server/reana_admin/main.py
mdonadoni: I agree that `--values-to-delete` is better. It should be changed also in the rest of the code.

reana_server/reana_admin/main.py
mdonadoni: For messages that do not match the given filters, we can either display nothing or show the received message along with `Message is ignored and re-queued`. If a message matches the given filters, we ask the user whether he wants to drop it or not.

In summary:
- if only `-k` (and `--values-to-delete`) is specified, the script drops all the messages that match the given filter
- if only `-i` is specified, the user decides what to do with each message
- if both `-i` and `-k` are specified, the user decides what to do with each message *that matches the given filter*

Do you think this is a good approach?

reana_server/reana_admin/consumer.py
VMois: Sounds better for UX. I will change it to what you suggested.

reana_server/reana_admin/main.py
VMois: I think this approach makes sense. I will change it.

tests/test_cli.py
VMois: It can be a nice "good first issue"

reana_server/reana_admin/main.py
VMois: I have no idea. The `consume-queue` command doesn't require `admin token` so I did not add it.

reana_server/reana_admin/main.py
VMois: Done.

reana_server/reana_admin/consumer.py
VMois: Done

reana_server/reana_admin/cli.py
```diff
@@ -639,3 +640,77 @@ def set_default_quota_limit(ctx):
...
+            key=key,
+            values_to_delete=list(values_to_delete),
+            is_interactive=interactive,
+        )
+    except Exception as error:
+        click.secho(
+            "Error is raised during MessageConsumer initialization. Please, check if arguments are correct.",
+            fg="red",
+        )
+        logging.exception(error)
```
mdonadoni, line 714: Do you think we need to show the stacktrace, or is the exception message enough?

reana_server/reana_admin/consumer.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+
+    def get_consumers(self, Consumer, channel):
+        """Implement providing kombu.Consumers with queues/callbacks."""
+        return [
+            Consumer(
+                queues=self.queue,
+                callbacks=[self.on_message],
+                accept=[self.message_default_format],
+                prefetch_count=50,
```
mdonadoni, line 60: I have noticed that when the consumer is waiting for the user's decision, it might happen that multiple messages are received by the consumer. In this case, using `Ctrl+C` or replying with `stop` does not actually stop the consumer, but it will show the next message in the queue. Would `prefetch_count=1` fix this issue? Are there any cons with this approach?

VMois: Honestly, no. It was `1` before. I will change it back.

reana_server/reana_admin/cli.py
VMois: What do you mean by stacktrace? `logging.exception` is already printing stacktrace (and not a very helpful message 😄 )

mdonadoni: Sorry, I was not clear. I was wondering if it might be better not to show the stacktrace :smile:. What do you think about showing only the exception message instead of the full stacktrace?

reana_server/reana_admin/consumer.py
mdonadoni: Works well now! 🚀

reana_server/reana_admin/main.py
mdonadoni: Related issue: https://github.com/reanahub/reana-server/issues/489

mdonadoni (approved):
Works very well, good job!

mdonadoni:
I just noticed that the PR needs to be rebased. Could you also update `CHANGES.rst`?

PR #493, opened 2022-05-16
admin: add command to detect out-of-sync workflows

### LINKED ISSUE #491: reana-admin check-workflow-runs
Observed on PROD:

- 1 workflow in `queued` status
- no log entries in r-server scheduler or r-w-controller job-status-consumer
- no unacked messages in MQ
- no k8s jobs for this workflow
- no k8s pods for this workflow

Since REANA relies on DB as SSOT about running workflows, it would be good to write a script to detect cases of this kind.

The script would go through workflow runs in the DB (either all, or workflows by a certain user U1, or workflows created since not longer than D1 days, etc) and it would check what is the DB status, and do we have k8s jobs or k8s pods and does the status match, and do we have anything in MQ, and do we see anything in the reana-server or reana-workflow-controller logs about this workflow.  Basically, it would check DB <-> MQ <-> K8S statuses to see whether we have an agreement or whether we have a discrepancy between them.

The script would then simply report about any observed inconsistencies, e.g. if DB says "finished" but the pod is still running, e.g. if DB says "queued" but there is no message to it up for running, e.g. DB says "pending" but there is no k8s job in the list, etc.

MVP: the script would simply report findings without fixing them. Admins would then run it say weekly to find about any problems on the deployment, or say when a user reports troubles, etc.

Beyond MVP: when the script would detect some very obvious inconsistency situation, such as the one listed above, it could propose "Shall I re-submit the workflow anew for execution? (Y/N)" or "Shall I mark the workflow as failed? (Y/N)" etc. This could be a possible future "interactive" extension.

Possible CLI API: (to check last 50 workflows from John Doe):

```
$ kubectl exec -i -t deployment/reana-server -c scheduler -- flask reana-admin check-workflow-runs --user john.doe@cern.ch --size 50
```

And, for each workflow run ID, the script would report DB status and whether things are OK or whether there are any inconsistencies; and show them if any.

VMois:
closes #491

## How to test

You will need to reproduce a few scenarios where workflows end up in sync situations. There are multiple ways to do it and probably some ways we didn't discover yet. Below you can find *one* way to do it for each state.

### 1. Queued

The message is not published to the `workflow-submission` queue. To do this go to `values-dev.yaml`. Enable "balanced" strategy, and set `kubernetes_jobs_max_user_memory_limit`:

```yaml
components:
    reana_server:
      image: reanahub/reana-server
      environment:
        ...
        REANA_WORKFLOW_SCHEDULING_POLICY: "balanced"
...
kubernetes_jobs_max_user_memory_limit: "4Gi"
```

Go to `helloworld` workflow, set `kubernetes_memory_limit`higher than allowed - `8Gi`. Run the workflow using `reana-client run`. Submission will fail right away but the workflow will end up in the queued state. **Revert your change in `reana.yaml`**

This is an actual bug that needs a separate issue :)

### 2. Pending

This is harder to reproduce without breaking code explicitly. Go to `r-w-controller/rest/utils.py` to `start_workflow` function. Replace:

```python
  try:
      kwrm.start_batch_workflow_run(
          overwrite_input_params=parameters.get("input_parameters"),
          overwrite_operational_options=parameters.get("operational_options"),
      )
      _start_workflow_db(workflow, parameters)
  except SQLAlchemyError as e:
  ...
```

with:

```python
 try:
      _start_workflow_db(workflow, parameters)
      raise ApiException("hello")
      kwrm.start_batch_workflow_run(
          overwrite_input_params=parameters.get("input_parameters"),
          overwrite_operational_options=parameters.get("operational_options"),
      )
  except SQLAlchemyError as e:
  ...
```

Run `helloworld`. It will stay pending forever.
**Revert code changes you made.**

### 3. Running

Go to `r-w-controller/consumer.py` and add this line:

```python
def _update_workflow_status(workflow, status, logs):
    """Update workflow status in DB."""
    if workflow.status != status:
        if status == RunStatus.finished:
            raise Exception("Something abd with db")
        ...
```

Rebuild image, load to kind, delete current pod and start a workflow. It will be stuck in running state.

### 4. Finished/Failed

You can make any workflow fail or finish while having `REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES: failed` set up in `values-dev.yaml`. Similar to finished.

### 5. Run one workflow correctly

Run a plain `helloworld` demo with no memory requirements. Let it finish.

### 6. Run `reana-admin check-workflows` command.

- ssh into scheduler container using `kubectl exec -it deployment/reana-server -c scheduler -- bash`

- run `flask reana-admin check-workflows` command. You will get a report. The first 4 workflows should be out of sync, the last one should be fine.

### DIFF
--- a/reana_server/reana_admin/check_workflows.py
+++ b/reana_server/reana_admin/check_workflows.py
@@ -1,0 +1,218 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2022 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA Server administrative tool - check workflows command."""
+
+
+import datetime
+import logging
+import sys
+from typing import Optional, List, Dict, Tuple
+from functools import partial
+
+import click
+from reana_commons.k8s.api_client import current_k8s_corev1_api_client
+from reana_commons.config import REANA_RUNTIME_KUBERNETES_NAMESPACE
+from reana_db.database import Session
+from reana_db.models import (
+    Workflow,
+    RunStatus,
+)
+from kubernetes.client import V1Pod
+from kubernetes.client.rest import ApiException
+from sqlalchemy.orm import Query
+
+from reana_server.reana_admin.consumer import CollectingConsumer
+
+
+class WorkflowCheckFailed(Exception):
+    """Error when workflow doesn't pass some check."""
+
+
+WorkflowCheckResult = Tuple[Workflow, List[WorkflowCheckFailed]]
+
+
+def _display(validation_results: List[WorkflowCheckResult]):
+    for workflow_result in validation_results:
+        workflow = workflow_result[0]
+        click.secho(
+            f"- id: {str(workflow.id_)}, name: {workflow.name}, user: {workflow.owner.email}, status: {workflow.status}"
+        )
+
+        failed_checks = workflow_result[1]
+        for check_error in failed_checks:
+            click.secho(f"   - ERROR: {check_error}\n", fg="red")
+
+
+def _get_all_pods() -> List[V1Pod]:
+    try:
+        response = current_k8s_corev1_api_client.list_namespaced_pod(
+            namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE
+        )
+        return response.items
+    except ApiException as error:
+        click.secho(
+            "Couldn't fetch list of pods due to an error.",
+            fg="red",
+        )
+        logging.exception(error)
+        return []
+
+
+def _collect_messages_from_scheduler_queue(filtered_workflows: Query) -> Dict:
+    scheduler_messages_collector = CollectingConsumer(
+        queue_name="workflow-submission",
+        key="workflow_id_or_name",
+        values_to_collect=[str(workflow.id_) for workflow in filtered_workflows],
+    )
+    scheduler_messages_collector.run()
+    return scheduler_messages_collector.messages
+
+
+def _message_is_in_scheduler_queue(workflow, pods, scheduler_messages):
+    if workflow.id_ not in scheduler_messages:
+        raise WorkflowCheckFailed("Message is not found in workflow-submission queue.")
+
+
+def _pods_dont_exist(workflow, pods, scheduler_messages):
+    if len(pods) > 0:
+        raise WorkflowCheckFailed("Some pods still exist for the workflow.")
+
+
+def _pods_exist(workflow, pods, scheduler_messages):
+    if len(pods) == 0:
+        raise WorkflowCheckFailed("No pods found for the workflow.")
+
+
+def _all_batch_pods_have_phase(workflow, pods, scheduler_messages, phase):
+    if any(
+        [pod.status.phase != phase for pod in pods if "batch-pod" in pod.metadata.name]
+    ):
+        raise WorkflowCheckFailed(f"Some pods are not in {phase}.")
+
+
+def _no_batch_pods_are_in_notready_state(workflow, pods, scheduler_messages):
+    for pod in pods:
+        if "batch-pod" in pod.metadata.name:
+            for container in pod.status.container_statuses:
+                if container.state.terminated:
+                    raise WorkflowCheckFailed(
+                        f"{pod.metadata.name} pod is in NotReady state."
+                    )
+
+
+validation_map = {
+    RunStatus.queued: [
+        _message_is_in_scheduler_queue,
+    ],
+    RunStatus.pending: [
+        _pods_exist,
+        partial(_all_batch_pods_have_phase, phase="Pending"),
+    ],
+    RunStatus.running: [
+        _pods_exist,
+        partial(_all_batch_pods_have_phase, phase="Running"),
+        _no_batch_pods_are_in_notready_state,
+    ],
+    RunStatus.finished: [
+        _pods_dont_exist,
+    ],
+    RunStatus.failed: [
+        _pods_dont_exist,
+    ],
+}
+
+
+def check_workflows(
+    date_start: datetime.datetime, date_end: Optional[datetime.datetime]
+) -> None:
+    """Check if selected workflows are in sync with database according to predefined rules."""
+    # filter workflows that need to be checked
+    statuses_to_check = validation_map.keys()
+    query_filter = [
+        Workflow.status.in_(statuses_to_check),
+        Workflow.created >= date_start,
+    ]
+
+    if date_end:
+        query_filter.append(Workflow.created < date_end)
+
+    filtered_workflows = Session.query(Workflow).filter(*query_filter)
+
+    time_range_message = f"{date_start} - {date_end if date_end else 'now'}"
+
+    if filtered_workflows.count() == 0:
+        click.secho(
+            f"No workflows found matching time range {time_range_message}.\nPlease, adjust your filter conditions.",
+            fg="red",
+        )
+        sys.exit(1)
+    click.secho(
+        f"Found {filtered_workflows.count()} workflow(s) matching time range {time_range_message}.\n",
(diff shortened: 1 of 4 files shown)


reana_server/reana_admin/cli.py
```diff
@@ -716,3 +718,159 @@ def queue_consume(
...
+        RunStatus.pending,
+        RunStatus.running,
+        RunStatus.finished,
+        RunStatus.failed,
+    ]
+
+    query_filter = [Workflow.status.in_(statuses_to_check)]
+
+    if date_start:
+        query_filter.append(Workflow.created >= date_start)
```
VMois, line 755: *suggestion:* by default we can set 24 hours back from now.

reana_server/reana_admin/cli.py
```diff
@@ -716,3 +718,159 @@ def queue_consume(
...
+    if filtered_workflows.count() == 0:
+        click.secho(
+            "No workflows found. Please, adjust your filter conditions.",
+            fg="red",
+        )
+        return
+
+    click.secho(
+        f"Found {filtered_workflows.count()} workflows. Starting to check them...",
+    )
```
VMois, line 771: *suggestion:* Maybe before proceeding we should ask users whatever they want to analyze workflows? Imagine having 100+ workflows. It might take a while (or not).

reana_server/reana_admin/consumer.py
```diff
@@ -114,3 +114,56 @@ def on_message(self, body, message):
...
+        return [
+            Consumer(
+                queues=self.queue,
+                callbacks=[self.on_message],
+                accept=[self.message_default_format],
+                prefetch_count=1,
+            )
+        ]
+
+    def on_consume_ready(self, connection, channel, consumers, **kwargs):  # noqa: D102
```
audrium, line 146: Minor: I guess `# noqa: D102` is not needed here?

reana_server/reana_admin/cli.py
```diff
@@ -716,3 +718,159 @@ def queue_consume(
...
+        )
+        pods = response.items
+    except ApiException:
+        pods = []
+
+    # validate if workflows are in sync with DB
+    in_sync_workflows = []
+    out_of_sync_workflows = []
+
+    for workflow in filtered_workflows:
```
audrium, line 803: It would be nice to somehow increase the readability of this peace of code (line 803-860). Perhaps breaking it down in small functions would help? Not easy to follow the logic with nested if statements, etc

audrium (commented):
Nice work! I left some minor cosmetic comments, but it works without any issues

reana_server/reana_admin/cli.py
VMois: Agree. I will work on that. Any suggestions on how to break it?

audrium: It's a pity that we can't still use [structural pattern matching](https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching) which would play nicely here, but I think breaking it down in small functions and placing it in some utils file would help. Also we could use dictionary lookups instead of if statements for all the run statuses to point to these functions

reana_server/reana_admin/cli.py
```diff
@@ -716,3 +718,159 @@ def queue_consume(
...
+        scheduler_messages_collector.run()
+
+    # get snapshot of all present pods
+    try:
+        response = current_k8s_corev1_api_client.list_namespaced_pod(
+            namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE
+        )
+        pods = response.items
+    except ApiException:
+        pods = []
```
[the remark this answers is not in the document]
VMois: Good point. Maybe, we should even throw an error here. Cause without a pod list we cannot check properly most of the cases.

reana_server/reana_admin/cli.py
```diff
@@ -716,3 +716,23 @@ def queue_consume(
         logging.exception(error)
     else:
         consumer.run()
+
+
+@reana_admin.command("check-workflows")
+@click.option(
+    "--date-start",
```
audrium, line 723: Minor: we could specify that 24 hours are set by default

reana_server/reana_admin/check_workflows.py
```diff
@@ -0,0 +1,218 @@
...
+    RunStatus.finished: [
+        _pods_dont_exist,
+    ],
+    RunStatus.failed: [
+        _pods_dont_exist,
+    ],
+}
+
+
+def check_workflows(  # noqa: C901
```
audrium, line 140: Very minor: I think `# noqa: C901` is no longer needed

reana_server/reana_admin/check_workflows.py
```diff
@@ -0,0 +1,218 @@
...
+            "Error is raised when collecting scheduler messages.",
+            fg="red",
+        )
+        logging.exception(error)
+        sys.exit(1)
+
+    pods = _get_all_pods()
+    if not pods:
+        click.secho(
+            "List of active pods is empty. It is required to validate worfklows. Exiting..",
```
audrium, line 182: Very minor: typo: `worfklows`

reana_server/reana_admin/check_workflows.py
```diff
@@ -0,0 +1,218 @@
...
+
+        for check in checks:
+            check_results.append(check(workflow, filtered_pods, scheduler_messages))
+
+        workflow_is_not_in_sync = any([not result[0] for result in check_results])
+
+        if workflow_is_not_in_sync:
+            out_of_sync_workflows.append((workflow, check_results))
+        else:
+            in_sync_workflows.append((workflow, check_results))
```
audrium, line 203: I think that it would be enough just to print a compact list of `in_sync_workflows` ids and we could get rid of all the other detailed info, since using this script on production with larger periods of time or even last 24 hours could result in lots of workflow output. I've simply tested with 15 workflows and it was too much output to check.
Also, the most interesting part is `out_of_sync_workflows`, so I think there is not much value to check `in_sync_workflows` output? I would also suggest to try to compact the output, to trim all the new lines and to fit as much information to the screen as possible (just to prepare for large number of workflows to parse).

reana_server/reana_admin/check_workflows.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def _message_is_in_scheduler_queue(workflow, pods, scheduler_messages) -> CheckResult:
+    if workflow.id_ not in scheduler_messages:
+        return False, "Message is not found in workflow-submission queue."
+    else:
+        return True, "Message is found in workflow-submission queue."
+
+
+def _pods_dont_exist(workflow, pods, scheduler_messages) -> CheckResult:
```
audrium, line 82: This comment is related to my comment below regarding `in_sync_workflows`. I think we could return only the errors in these `check` functions, since as I mentioned below successful output may not be that relevant. Also it would increase the readability of these functions, because now after a quick look into these functions it's not clear which return message is the error and which is not

audrium (commented):
Looks much better then before! I've left few more cosmetic comments

reana_server/reana_admin/check_workflows.py
VMois: I made changes. Please, check if those are good enough.

reana_server/reana_admin/check_workflows.py
VMois: I replace booleans with exceptions. I think this will be more intuitive for whoever reads a code. What do you think?

audrium (approved):
LGTM! Nice work!

P.S. Rebase is needed

reana_server/reana_admin/check_workflows.py
audrium: Looks much better indeed!

reana_server/reana_admin/check_workflows.py
audrium: IMO it's very compact and nice now, thanks

tiborsimko (approved):

PR #494, opened 2022-05-18
launcher: remove unnecessary files after fetching

### LINKED ISSUE #492: launcher: remove unnecessary files after cloning
**Current behaviour**

When launching workflow from GitHub, e.g. using "Launch-on-REANA@CERN" badge at https://github.com/reanahub/reana-demo-root6-roofit/, the whole source code repository is cloned and the workflow is launched.

This has a drawback that if we are launching say Snakemake workflow, the workspace has a lot of Yadage etc unnecessary files.

This may be confusing to the users e.g. why are these files there? And why there are not there when users use `reana-client`?

**Expected behaviour**

Since users always declare which files are necessary for analysis via `reana.yaml`'s `inputs.files` and `inputs.directories` clause, we can actually "clean" the workspace right after cloning and delete all the undeclared files. (Though beware about `reana.yaml`, it may be nice to keep this one.)

This will make the workspace-when-launched-via-badge and workspace-when-started-via-cli identical.

mdonadoni:
Closes #492

How to test:
1. Launch one of the demo examples, e.g. `reana-demo-helloworld`
2. Check that only the files and directories specified in `reana.yaml` are present in the workspace

### DIFF
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -10,13 +10,15 @@
 import base64
 import csv
 import io
+import itertools
 import json
 import logging
 import os
+import pathlib
 import secrets
 import sys
 import shutil
-from typing import Optional, Dict
+from typing import Dict, List, Optional, Sequence, Union
 from uuid import UUID, uuid4
 
 import click
@@ -29,7 +31,7 @@ from marshmallow.validate import Email
 
 from reana_commons.config import REANAConfig, REANA_WORKFLOW_UMASK, SHARED_VOLUME_PATH
 from reana_commons.email import send_email
-from reana_commons.errors import REANAQuotaExceededError
+from reana_commons.errors import REANAQuotaExceededError, REANAValidationError
 from reana_commons.k8s.secrets import REANAUserSecretsStore
 from reana_commons.yadage import yadage_load_from_workspace
 from reana_db.database import Session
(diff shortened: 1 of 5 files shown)


reana_server/utils.py
```diff
@@ -130,10 +132,57 @@ def remove_fetched_workflows_dir(tmpdir: str) -> None:
...
+                raise REANAValidationError(f"Path cannot contain '..': {path}")
+            if path in unique_paths:
+                raise REANAValidationError(f"Path declared multiple times: {path}")
+            unique_paths.add(path)
+
+        for x, y in itertools.permutations(paths, r=2):
+            if is_relative_to(x, y):
+                raise REANAValidationError(f"Invalid paths: '{y}' is a prefix of '{x}'")
+
+        return paths
```
mdonadoni, line 171: I think we should validate the paths present in `inputs.directories` and `inputs.files` even when using reana-client. Should I open an issue to move this logic to reana-commons and to use it in both reana-client and reana-server?

VMois (commented):
Works nicely. Tried with `rootfit`. I really like how you use `Path` in the code.

I found one issue. Not sure how important it is to address rn. If input validation fails, workflow is still created but has no files.

If this issue is important to solve now, one way is to separate validation from `mv_workflow_files` so that validation is done before workflow creation.

reana_server/utils.py
```diff
@@ -130,10 +132,57 @@ def remove_fetched_workflows_dir(tmpdir: str) -> None:
...
-    """Move files from one directory to another."""
-    for entry in os.listdir(source):
-        shutil.move(os.path.join(source, entry), target)
+def mv_workflow_files(source: str, target: str, reana_spec: Dict) -> None:
+    """Move workflow's input files from one directory to another.
+
+    :param source: Directory from which the files will be moved.
+    :param target: Directory where the files will be moved to.
+    :param reana_spec: REANA specification used to decide which files to move.
+    """
```
VMois, line 141: In addition to moving files, this function also validates them. Maybe call it `validate_and_move_workflow_files`?

mdonadoni:
> I found one issue. Not sure how important it is to address rn. If input validation fails, workflow is still created but has no files.

Nice catch! There was also another issues with the usage quotas, as they were updated before removing the files.

Can you please check if now everything works as expected?

reana_server/validation.py
```diff
@@ -92,3 +124,4 @@ def validate_workflow(reana_yaml: Dict, input_parameters: Dict) -> None:
     validate_input_parameters(input_parameters, original_parameters)
     validate_compute_backends(reana_yaml)
     validate_workspace_path(reana_yaml)
+    validate_inputs(reana_yaml)
```
VMois, line 127: Nice! Now input validation is a part of our validation.

reana_server/utils.py
```diff
@@ -136,6 +138,68 @@ def mv_workflow_files(source: str, target: str) -> None:
...
+            path.unlink()
+        else:
+            shutil.rmtree(path)
+
+    # Move the contents of the temporary `filtered` directory to the workspace
+    for path in filtered.iterdir():
+        path.replace(workspace / path.name)
+
+    # Remove temporary `filtered` directory
+    filtered.rmdir()
```
VMois, line 200: *note(non-blocking):* Initially, I was thinking maybe it is better to just delete files/directories that are not in `input` instead of moving them to `filtered-<uuid>` and back. I came up with this:

```python
    for path in workspace.iterdir():
        if any([is_relative_to(workspace / input_path, path) for input_path in paths]):
            continue
        if path.is_file() or path.is_symlink():
            path.unlink()
        else:
            shutil.rmtree(path)
```

But my code has an issue, it can copy unnecessary files. I think it is possible to do it with `os.walk` but I am not sure it will give any performance benefits. Move operations are pretty fast. Anyway, leaving it here as a learning/observation.

VMois (approved):
Nice work on fixing issues!

PR #499, opened 2022-05-30
rest: fail Gitlab commit if quota is exceeded

### LINKED ISSUE #495: gitlab: verify behaviour when user exceeds quota
When a user exceeds disk quota, the CLI way of submitting workflows will prevent user from creating new workflow runs.

The behaviour was verified to work well. However, some improvements were proposed in https://github.com/reanahub/reana-client/issues/607.

This goal of this issue is to do the same verification for GitLab <-> REANA way of submitting workflows:

- will the workflow be prevented from starting?
- will the error message back to GitLab web informative enough?
- will the user sees the core of the issue on the GitLab side?
- will the user know what to do next when this happens?

VMois:
closes #495

### How to test

We do not have a local Gitlab integration, so you cannot test. Currently, this branch is deployed on production, so you can try to run some Gitlab workflow ([how to](https://docs.reana.io/advanced-usage/code-repositories/gitlab/)).

Test if `r-client` works as it was working before.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,7 @@ Version 0.9.0 (UNRELEASED)
 - Changes workflow scheduler to allow defining the checks needed to assess whether the cluster can start new workflows.
 - Changes the Invenio dependencies to the latest versions.
 - Changes OAuth configuration to enable the new CERN SSO.
+- Fixes GitLab integration error reporting in case user exceeds CPU or Disk quota usage limits.
 
 Version 0.8.4 (2022-02-23)
 --------------------------
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -22,18 +22,17 @@ from reana_commons.validation.operational_options import validate_operational_op
 from reana_commons.validation.utils import validate_workflow_name
 from reana_commons.specification import load_reana_spec
 from reana_db.database import Session
-from reana_db.models import (
-    InteractiveSessionType,
-    RunStatus,
-)
+from reana_db.models import InteractiveSessionType, RunStatus
 from reana_db.utils import _get_workflow_with_uuid_or_name
 from webargs import fields, validate
 from webargs.flaskparser import use_kwargs
 
 from reana_server.api_client import current_rwc_api_client
+from reana_server.config import REANA_HOSTNAME
 from reana_server.decorators import check_quota, signin_required
 from reana_server.validation import validate_workspace_path
 from reana_server.utils import (
+    _fail_gitlab_commit_build_status,
     RequestStreamWithLen,
     _load_and_save_yadage_spec,
     _get_reana_yaml_from_gitlab,
@@ -274,7 +273,6 @@ def get_workflows(user, **kwargs):  # noqa
 
 @blueprint.route("/workflows", methods=["POST"])
 @signin_required(include_gitlab_login=True)
-@check_quota
 def create_workflow(user):  # noqa
     r"""Create a workflow.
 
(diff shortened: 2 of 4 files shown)


reana_server/rest/workflows.py
```diff
@@ -365,57 +363,70 @@ def create_workflow(user):  # noqa
...
+                git_branch,
+                git_commit_sha,
+            ) = _get_reana_yaml_from_gitlab(request.json, user.id_)
+            git_data = {
+                "git_url": git_url,
+                "git_branch": git_branch,
+                "git_commit_sha": git_commit_sha,
+            }
+        elif request.args.get("spec"):
+            return jsonify("Not implemented"), 501
```
VMois, line 386: *suggestion(non-blocking):* I don't think this code is used at all. Or prevents something from happening. It has been carried here for 5 years. Maybe we should delete it (in a separate commit)?

VMois:
Screenshot 2022-06-01 at 10 31 52

VMois:
CI issue with docs-sphinx is fixed in https://github.com/reanahub/reana-server/pull/493

CHANGES.rst
```diff
@@ -4,6 +4,7 @@ Changes
 Version 0.9.0 (UNRELEASED)
 --------------------------
 
+- Fixes when Gitlab hook is received and user exceeds quota limit, commit build status is properly failed.
```
tiborsimko, line 7: The order of news items is usually Adds/Changes/Fixes, to announce new features before feature changes and bug fixes. (Although we don't always respect it everywhere!)  You can also reformulate the text to read like a user-oriented phrase, perhaps:

- Fixes GitLab integration error reporting in case user exceeds CPU or Disk quota usage limits.

reana_server/rest/workflows.py
```diff
@@ -22,18 +22,17 @@
...
-    RunStatus,
-)
+from reana_db.models import InteractiveSessionType, RunStatus
 from reana_db.utils import _get_workflow_with_uuid_or_name
 from webargs import fields, validate
 from webargs.flaskparser import use_kwargs
 
 from reana_server.api_client import current_rwc_api_client
 from reana_server.decorators import check_quota, signin_required
+from reana_server.config import REANA_HOSTNAME
```
tiborsimko, line 32: Move one line up, `c..` imports before `d...` imports.

reana_server/rest/workflows.py
```diff
@@ -365,57 +363,67 @@ def create_workflow(user):  # noqa
...
-            workflow_engine = reana_spec_file["workflow"]["type"]
-        elif request.args.get("spec"):
-            return jsonify("Not implemented"), 501
-        else:
+        json_body_is_not_present = not request.json
+        if json_body_is_not_present:
             raise Exception(
-                "Either remote repository or a reana spec need to \
-            be provided"
+                "Either remote repository or a reana specification needs to be provided"
```
tiborsimko, line 369: Either remote repository or REANA specification needs to be provided.

reana_server/rest/workflows.py
```diff
@@ -365,57 +363,67 @@ def create_workflow(user):  # noqa
...
+            raise Exception("Unknown workflow type.")
+
+        operational_options = validate_operational_options(
+            workflow_engine, reana_spec_file.get("inputs", {}).get("options", {})
+        )
+
         workflow_dict = {
             "reana_specification": reana_spec_file,
             "workflow_name": workflow_name,
+            "operational_options": operational_options,
```
tiborsimko, line 413: Hmm, haven't tested the part concerning operational options. Do we have an example?

VMois: Why do you need to test it? This PR doesn't change how we handle `operational_options`. I just moved it into a separate variable to make dictionary creation nicer.

reana_server/rest/workflows.py
tiborsimko: I think we should keep it. We do have "spec" in create_workflow OpenAPI contract and it is obviously also exposed  to clients via `r-commons` and such:

```json
          {
            "description": "Remote repository which contains a valid REANA specification.",
            "in": "query",
            "name": "spec",
            "required": false,
            "type": "string"
          },
```

So, if some client decides to pass it along, I think it is safe to keep returning "Not implemented".

(IOW, we should not remove the implementation whilst keeping the interface. An alternative would be to remove it from OpenAPI everywhere, if it is never used, but then this would create unnecessary complications for us to handle incompatibilities with API versioning. It's simpler to keep both interface and implementation and return "Not implemented" in case somebody passes it along.)

tiborsimko (approved):
Works nicely 👍 Tested both locally and with GitLab@CERN.

PR #500, opened 2022-06-02
docs: set language to English

VMois:


docs/conf.py
```diff
@@ -95,7 +95,7 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
```
tiborsimko, line 98: Please remove the Sphinx pin in `setup.py` if it's no longer needed.

tiborsimko (approved):

PR #501, opened 2022-06-02
rest: add docs link to user quota exceeded error output

audrium:
closes https://github.com/reanahub/reana-client/issues/607
- reana-client run
```
...
==> ERROR: Cannot create workflow :
User quota exceeded.
Resource: disk, usage: 16.97 MiB out of 1.91 MiB used (890%)
Please find more details here: https://docs.reana.io/advanced-usage/user-quotas
```
Screenshot from 2022-06-02 16-39-20


reana_server/utils.py
```diff
@@ -97,6 +99,21 @@ def get_fetched_workflows_dir(user_id: str) -> str:
...
+    quota = user.get_quota_usage()
+    message = "User quota exceeded.\n"
+    for resource_type, resource in quota.items():
+        limit = resource.get("limit", {}).get("raw", 0)
+        usage = resource.get("usage", {}).get("raw", 0)
+        if 0 < limit <= usage:
+            resource_usage, _ = get_quota_resource_usage(resource, "human_readable")
+            message += f"Resource: {resource_type}, usage: {resource_usage}\n"
+
+    message += f"Please find more details here: {REANA_QUOTAS_DOCS_URL}"
```
tiborsimko, line 113: Instead of "Please find more details...", perhaps write simply: "Please see https://...". It is shorter and perhaps more inciting to take an action than just "find more information".

tiborsimko: (or "Please see ... to resolve the situation." or some such... you know what I mean)

VMois (commented):
The "nice" message doesn't work with the launcher. For example, when initially quota is okay but after downloading the remote repository it exceeds the limits, and the launcher fails.

Screenshot 2022-06-07 at 10 06 56

You will need to modify `prevent_disk_quota_excess` in `reana_server/utils.py`. Also, would be good to add an exception in `launcher` to catch and properly return an error message in the case when `prevent_disk_quota_excess` is triggered.

reana_server/rest/launch.py
```diff
@@ -217,7 +217,13 @@ def launch(user, url, name="", parameters="{}", specification=None):
...
-    except (REANAFetcherError, REANAValidationError, ValueError, ValidationError) as e:
+    except (
+        REANAFetcherError,
+        REANAValidationError,
+        REANAQuotaExceededError,
+        ValueError,
+        ValidationError,
+    ) as e:
         logging.error(traceback.format_exc())
         return jsonify({"message": str(e)}), 400
```
VMois, line 228: *suggestion:* Here return code is 400. In `check_quota` it is 403. Let's have a separate catch for quota that returns HTTP 403.

reana_server/utils.py
```diff
@@ -122,7 +142,7 @@ def prevent_disk_quota_excess(user, bytes_to_sum: int, action=Optional[str]):
             action = "This action"
         raise REANAQuotaExceededError(
             f"{action} would exceed the disk quota limit "
-            f"({human_readable_limit}). Aborting."
+            f"({human_readable_limit}). Aborting. Please see: {REANA_QUOTAS_DOCS_URL}"
```
VMois, line 145: *question(non-blocking):* will it make sense to unify the error messages here and use `get_quota_excess_message` everywhere?

VMois (approved):
Good job. Now when the quota was still okay before downloading the repo but exceed after users get this error:

Screenshot 2022-06-07 at 14 36 02

PR #503, opened 2022-06-13
config: fix REANA_SCHEDULER_REQUEUE_SLEEP for large numbers

### LINKED ISSUE #502: high REANA_SCHEDULER_REQUEUE_COUNT values lead to deployment troubles
**Current behaviour**

When a REANA deployment uses low requeue sleep configuration (e.g. 1 sec), they may want to use high requeue count limits so that workflows would not be failed due to inability to get started when the cluster is busy.

Using a high requeue count limit value such as `1000000` leads to helm deployment change for the corresponding environment variable to be:

```diff
            - name: REANA_SCHEDULER_REQUEUE_COUNT
-             value: "200"
+             value: "1e+06"
```

This is problematic, because the environment variable is indeed attempted to be equal to 1e+06, which leads to the deployment crash `CrashLoopBackOff` because the variable is expected to be numerical when the pod is started. 

Note that value of 10000 is OK, which may be enough for many installations... But we should fix the crash for big values.

**Expected behaviour**

We should accept arbitrarily big number that user may want to pass in `REANA_SCHEDULER_REQUEUE_COUNT`.  

Note that `int(1e+06)` would have worked, whilst `int('1e+06')` leads to crash.

**How to test**

```
$ vim helm/configurations/values-dev.yaml
$ git diff
diff --git a/helm/configurations/values-dev.yaml b/helm/configurations/values-dev.yaml
index 94e70de..6d4707d 100644
--- a/helm/configurations/values-dev.yaml
+++ b/helm/configurations/values-dev.yaml
@@ -5,6 +5,7 @@ components:
       image: reanahub/reana-server
       environment:
         REANA_SCHEDULER_REQUEUE_SLEEP: 2
+        REANA_SCHEDULER_REQUEUE_COUNT: 100000
         REANA_RATELIMIT_SLOW: "5 per second"
     reana_workflow_controller:
       image: reanahub/reana-workflow-controller
$ helm diff upgrade reana helm/reana  -f helm/configurations/values-dev.yaml
$ helm upgrade reana helm/reana  -f helm/configurations/values-dev.yaml
```

**Notes**

In addition to fixing the deployment crash for big numbers, we may perhaps want to advertise a special value of `-1` or `0` that would mean "infinity", effectively removing the requeue count functionality.

Also, we may want to check how other variables such as `REANA_SCHEDULER_REQUEUE_SLEEP` behave when passing "unexpected" values.

VMois:
- replace int() with float() to allow "1e+06" liek numbers and "infinity"

closes #502

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -264,7 +264,7 @@ ADMIN_EMAIL = os.getenv("REANA_EMAIL_SENDER", "CHANGE_ME")
 REANA_SCHEDULER_REQUEUE_SLEEP = float(os.getenv("REANA_SCHEDULER_REQUEUE_SLEEP", "15"))
 """How many seconds to wait between consuming workflows."""
 
-REANA_SCHEDULER_REQUEUE_COUNT = int(os.getenv("REANA_SCHEDULER_REQUEUE_COUNT", "200"))
+REANA_SCHEDULER_REQUEUE_COUNT = float(os.getenv("REANA_SCHEDULER_REQUEUE_COUNT", "200"))
 """How many times to requeue workflow, in case of error or busy cluster, before failing it."""
 
 # Workflow fetcher
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,7 @@ Version 0.9.0 (UNRELEASED)
 - Changes workflow scheduler to allow defining the checks needed to assess whether the cluster can start new workflows.
 - Changes the Invenio dependencies to the latest versions.
 - Changes OAuth configuration to enable the new CERN SSO.
+- Fixes issue when irregular number formats are passed to ``REANA_SCHEDULER_REQUEUE_COUNT``.
 - Fixes GitLab integration error reporting in case user exceeds CPU or Disk quota usage limits.
 
 Version 0.8.4 (2022-02-23)


reana_server/config.py
```diff
@@ -264,7 +264,7 @@ def _get_rate_limit(env_variable: str, default: str) -> str:
 REANA_SCHEDULER_REQUEUE_SLEEP = float(os.getenv("REANA_SCHEDULER_REQUEUE_SLEEP", "15"))
 """How many seconds to wait between consuming workflows."""
 
-REANA_SCHEDULER_REQUEUE_COUNT = int(os.getenv("REANA_SCHEDULER_REQUEUE_COUNT", "200"))
+REANA_SCHEDULER_REQUEUE_COUNT = float(os.getenv("REANA_SCHEDULER_REQUEUE_COUNT", "200"))
```
tiborsimko, line 267: What about using `int(float("1e+06"))`? Just to make sure that the final result is well an integer as the main business logic expects.  

It does not make much difference here, but it could for other variables.

(We could probably use this solution for all variables where we expect an integer and where the passed value may be too huge... Unless we solve the core issue on the helm level already.)

VMois: `int(float("1e+06"))` should work. I added `float` because you want to introduce a special value to disable REQUEUE_COUNT and float supports "infinity" as a parameter.

VMois: If we will have some other special value like zero or -1 to disable REQUEUE, we will need to add some new logic in the code to account for that. It was just simpler with float(). 

If "infinity" is not a good name, we can have something like "disabled" or "none".

tiborsimko: Ah, I see; here "infinity" seems to be a handy implementation trick indeed; although "count" being "float" looks a bit weird conceptually. We can keep it here, and elsewhere for other variables (where we don't allow infinity and such) we can cast back to int.

VMois: So to confirm, if the admin wants to disable the re-queue count check they should use "infinity" like:

```yaml
REANA_SCHEDULER_REQUEUE_COUNT: infinity
```

audrium (approved):
LGMT!

PR #506, opened 2022-06-28
rest: populate workspace retention rules when creating new workflow

audrium:
closes https://github.com/reanahub/reana-workflow-controller/issues/458


CHANGES.rst
```diff
@@ -9,6 +9,8 @@ Version 0.9.0 (UNRELEASED)
 - Adds new ``/api/launch`` endpoint that allows running workflows from remote sources.
 - Adds the possibility to fetch the workflow specification from a remote URL.
 - Adds REANA specification validation utilities.
+- Changes workflow create endpoint to populate workspace retention rules for the workflow.
```
mdonadoni, line 12: I think we should move this line to the first commit

reana_server/config.py
```diff
@@ -301,11 +301,11 @@ def _get_rate_limit(env_variable: str, default: str) -> str:
 # Workspace retention rules
 # ==================
 
-MAXIMUM_WORKSPACE_RETENTION_PERIOD = 365  # TODO: specify this through Helm
+WORKSPACE_RETENTION_PERIOD = int(os.getenv("WORKSPACE_RETENTION_PERIOD", "365"))
```
mdonadoni, line 304: I think we should move all the changes related to the renaming of this variable to the first commit (both in this file and in `utils.py`)

reana_server/rest/workflows.py
```diff
@@ -997,6 +1002,7 @@ def start_workflow(workflow_id_or_name, user):  # noqa
                 .get("workflow", {})
                 .get("type", None)
             )
+            workflow.inactivate_workspace_retention_rules()
```
mdonadoni, line 1010: What is your opinion on moving this line inside `clone_workflow`? In this way, all the logic to clone a workflow is in a single function

mdonadoni (approved):
Works nicely!

PR #524, opened 2022-07-15
validate retention rules

VMois:


reana_server/utils.py
```diff
@@ -160,16 +161,6 @@ def mv_workflow_files(source: str, target: str) -> None:
         shutil.move(os.path.join(source, entry), target)
 
 
-# FIXME: use `is_relative_to` from the standard library when moving to Python 3.9
-def is_relative_to(path: pathlib.Path, base: pathlib.Path) -> bool:
```
VMois, line 164: `is_relative_to` was causing circular import issues because it is also used in `validation.py`. I moved it to `validation.py` to prevent an error.

*question:* maybe, it makes sense to place `is_relative_to` to some other file? If you have ideas, please, share.

reana_server/validation.py
```diff
@@ -125,3 +134,20 @@ def validate_workflow(reana_yaml: Dict, input_parameters: Dict) -> None:
...
+
+    :raises reana_commons.errors.REANAValidationError: if rule is not valid
+    """
+    rule_path = pathlib.Path(rule)
+    if rule_path.is_absolute():
+        raise REANAValidationError(f"Retention rule {rule} cannot be an absolute path")
+    if not rule_path.parts:
+        raise REANAValidationError(f"Retention rule {rule} cannot be empty")
+    if ".." in rule_path:
+        raise REANAValidationError(f"Retention rule {rule} cannot contain '..'")
```
VMois, line 153: This part is similar to `validate_inputs`. Do you think extracting this block to a common function like `validate_local_path` makes sense?

mdonadoni: I personally think that it does not make too much sense to generalize this, given that in one case we are validating paths, and in the other we are validating patterns with wildcards

reana_server/validation.py
```diff
@@ -125,3 +134,20 @@ def validate_workflow(reana_yaml: Dict, input_parameters: Dict) -> None:
...
+
+def validate_retention_rule(rule: str) -> None:
+    """Validate retention rule.
+
+    :param rule: retention rule
+    :type rule: string
+
+    :raises reana_commons.errors.REANAValidationError: if rule is not valid
+    """
+    rule_path = pathlib.Path(rule)
```
mdonadoni, line 147: Do you think using `pathlib` can cause some problems given that we are validating patterns and not paths? In particular, I don't know exactly how the library manipulates paths and whether this can cause undesired behaviours when passing patterns with wildcards.

reana_server/utils.py
```diff
@@ -570,7 +574,7 @@ def get_workspace_retention_rules(
                     "Maximum workflow retention period was reached. "
                     f"Please use less than {WORKSPACE_RETENTION_PERIOD} days."
                 )
-            # TODO: add validation for rules
+            validate_retention_rule(rule)
```
mdonadoni, line 577: (minor) what do you think about moving these lines inside `validate_retention_rule`?
https://github.com/reanahub/reana-server/blob/1ac23c4729bc2b2443ddf1b1f6dbb6d14edbc1a8/reana_server/utils.py#L568-L572

mdonadoni (approved):
LGTM!

PR #525, opened 2022-07-19
reana-admin: add `retention-rules-apply` command

mdonadoni:
Add command to apply pending retention rules, deleting the necessary files or directories.

How to test:
1. Modify one of the demo examples, defining some retention rules
2. Fake a different current time (e.g. one day in the future)
```diff
--- a/reana_server/reana_admin/cli.py
+++ b/reana_server/reana_admin/cli.py
@@ -748,6 +748,7 @@ def check_workflows(
 def retention_rules_apply() -> None:
     """Apply pending retentions rules."""
     current_time = datetime.datetime.now()
+    current_time = current_time + datetime.timedelta(days=1)
     pending_rules = WorkspaceRetentionRule.query.filter(
         WorkspaceRetentionRule.status == WorkspaceRetentionRuleStatus.active,
         WorkspaceRetentionRule.apply_on < current_time,
```
3. Run the `retention-rules-apply` command
```
kubectl exec -it deployments/reana-server -- flask reana-admin retention-rules-apply
```

Expected result: The retention rules are applied correctly and the correct files are deleted.

### DIFF
--- a/reana_server/reana_admin/retention_rule_deleter.py
+++ b/reana_server/reana_admin/retention_rule_deleter.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2022 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+"""REANA utilities to apply retention rules."""
+
+
+from pathlib import Path
+from typing import Dict, Union
+
+import click
+from reana_db.models import WorkspaceRetentionRule
+
+from reana_server.utils import is_relative_to
+
+
+class RetentionRuleDeleter:
+    """Delete files and directories matching the given retention rule."""
+
+    def __init__(self, rule: WorkspaceRetentionRule):
+        """Initialize the RetentionRuleDeleter.
+
+        :param rule: Retention rule to be applied.
+        """
+        self.rule_id = rule.id_
+        self.specification = rule.workflow.reana_specification
+        self.workflow_id = rule.workflow.id_
+        self.workspace = Path(rule.workflow.workspace_path)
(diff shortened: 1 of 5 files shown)


reana_server/reana_admin/retention_rule_deleter.py
```diff
@@ -0,0 +1,78 @@
...
+        :param file_or_dir: Path to the file/directory.
+        """
+        if self.is_input_output(file_or_dir):
+            click.echo(f"Preserved in/out: {file_or_dir}")
+        elif file_or_dir.is_file() or file_or_dir.is_symlink():
+            file_or_dir.unlink()
+            click.echo(f"Deleted file: {file_or_dir}")
+        else:
+            for child in file_or_dir.iterdir():
+                self.delete_keeping_inputs_outputs(child)
```
VMois, line 62: *note(non-blocking)*: usually, some deeply nested recursions can suffer from running out of stack memory due to many calls. In our case, here is all good. I just wanted to share the info :)

VMois (commented):
Very nice job!

One small thing to do is to add a line to `CHANGES`.

reana_server/reana_admin/retention_rule_deleter.py
mdonadoni: Do you think I should convert this to be an iterative function?

VMois: No, no. It is good, I think. The recursion limit in Python is around 1000 ([source](https://stackoverflow.com/questions/71317463/recursionerror-maximum-recursion-depth-exceeded-in-comparison-geeksforgeeks-i)). So unless you have thousand nested directories (which I doubt anyone has, or even filesystem supports), we should be fine.

VMois (approved):
Nice job!

PR #527, opened 2022-07-20
admin: detect out-of-sync interactive sessions

VMois:
- add possibility to "check-workflows" command to detect out-of-sync interactive sessions.

### Intro

Session has `created` status in DB when they are open. When it is closed, the record is deleted from the database. It differs from workflow, where deleted workflow is kept in the database.

### How to test

There are three scenarios:

1. Interactive session is in database, open, and pod exist:

- run some workflow like `helloworld`
- open a session (`reana-client open -w <workflow-name>`)
- ssh into `r-server` using `kubectl exec -i -t <r-server-pod-id> -- /bin/bash`
- run check command `flask reana-admin check-workflows`
- it should pass

2. Interactive session is in DB but pod doesn't exist:

- reuse a session from before
- run `kubectl get deployments` to locate Deployment for interactive session
- delete deployment using `kubectl delete deployment <id>`
- run check command like before, it should fail

3. Interactive session is not in DB (closed), but the pod still exists:

- open a new session
- ssh into r-server:
```bash
root@reana-server-f5976658d-hv8fw:/code# python
Python 3.8.13 (default, Jul 13 2022, 05:54:24)
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from reana_db.database import Session
>>> from reana_db.models import InteractiveSession
>>> sessions = InteractiveSession.query.all()
>>> sessions
[<InteractiveSession 'reana-run-session-d579f1b3-7e88-4143-b188-2a081c8db936'>]
>>> [Session.delete(session) for session in sessions]
[None]
>>> Session.commit()
>>> sessions = InteractiveSession.query.all()
>>> exit()
root@reana-server-f5976658d-hv8fw:/code# flask reana-admin check-workflows  # should fail
```

### DIFF
--- a/reana_server/reana_admin/check_workflows.py
+++ b/reana_server/reana_admin/check_workflows.py
@@ -10,9 +10,8 @@
 
 
 import datetime
-import logging
-import sys
 from typing import Optional, List, Dict, Tuple
+from dataclasses import dataclass
 from functools import partial
 
 import click
@@ -20,6 +19,7 @@ from reana_commons.k8s.api_client import current_k8s_corev1_api_client
 from reana_commons.config import REANA_RUNTIME_KUBERNETES_NAMESPACE
 from reana_db.database import Session
 from reana_db.models import (
+    InteractiveSession,
     Workflow,
     RunStatus,
 )
@@ -30,38 +30,55 @@ from sqlalchemy.orm import Query
 from reana_server.reana_admin.consumer import CollectingConsumer
 
 
-class WorkflowCheckFailed(Exception):
-    """Error when workflow doesn't pass some check."""
+class CheckFailed(Exception):
+    """Error when a check for workflow or session fails."""
 
 
-WorkflowCheckResult = Tuple[Workflow, List[WorkflowCheckFailed]]
+class InfoCollectionError(Exception):
+    """Error when check functions fails to query workflows/sessions, collect pods or queue messages."""
 
 
-def _display(validation_results: List[WorkflowCheckResult]):
-    for workflow_result in validation_results:
-        workflow = workflow_result[0]
+@dataclass
+class CheckSource:
+    """Contains metadata about workflow or session."""
+
+    id: str
+    name: str
+    user: Optional[str]
+    status: Optional[RunStatus]
+
+
+@dataclass
+class CheckResult:
+    """Contains information about source and its errors (if any)."""
+
+    source: CheckSource
+    errors: List[CheckFailed]
+
+
+def display_results(validation_results: List[CheckResult]):
+    """Display results from a list of CheckResults."""
+    for result in validation_results:
+        source = result.source
         click.secho(
-            f"- id: {str(workflow.id_)}, name: {workflow.name}, user: {workflow.owner.email}, status: {workflow.status}"
+            f"- id: {source.id}, name: {source.name}, user: {source.user}, status: {source.status}"
         )
 
-        failed_checks = workflow_result[1]
+        failed_checks = result.errors
         for check_error in failed_checks:
             click.secho(f"   - ERROR: {check_error}\n", fg="red")
 
 
 def _get_all_pods() -> List[V1Pod]:
-    try:
-        response = current_k8s_corev1_api_client.list_namespaced_pod(
-            namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE
-        )
-        return response.items
-    except ApiException as error:
-        click.secho(
-            "Couldn't fetch list of pods due to an error.",
-            fg="red",
-        )
-        logging.exception(error)
-        return []
+    """Collect information about Pods.
+
+    :raises kubernetes.client.rest.ApiException: in case of a k8s server error.
+    """
+    click.secho("Collecting information about Pods...")
+    response = current_k8s_corev1_api_client.list_namespaced_pod(
+        namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE
+    )
+    return response.items
 
 
 def _collect_messages_from_scheduler_queue(filtered_workflows: Query) -> Dict:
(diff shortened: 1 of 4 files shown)


mdonadoni (commented):
Could you please add some tests for both workflows and sessions, checking the various scenarios?

reana_server/reana_admin/check_workflows.py
```diff
@@ -212,7 +221,121 @@ def check_workflows(
         _display(out_of_sync_workflows)
 
     if out_of_sync_workflows:
-        click.secho("\nFAILED")
+        return False
+    else:
+        return True
```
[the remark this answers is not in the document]
VMois: Good catch! Ahaha. I was moving stuff around and haven't noticed this stupid piece.

reana_server/reana_admin/check_workflows.py
```diff
@@ -212,7 +221,121 @@ def check_workflows(
...
+        if is_it_session_pod:
+            pod_doesnt_have_session = (
+                len(
+                    [
+                        session
+                        for session in filtered_sessions
+                        if str(session.name) in pod.metadata.name
+                    ]
+                )
+                == 0
```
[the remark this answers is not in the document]
VMois: Yep. That looks much better.

reana_server/reana_admin/check_workflows.py
```diff
@@ -212,7 +221,121 @@ def check_workflows(
...
         sys.exit(1)
+
+    in_sync_sessions: List[CheckResult] = []
+    out_of_sync_sessions: List[CheckResult] = []
+    pods_without_session: List[CheckResult] = []
+
+    # check if some session pods don't have corresponding session entry in the database
+    for pod in pods:
+        is_it_session_pod = "run-session" in pod.metadata.name
+        if is_it_session_pod:
```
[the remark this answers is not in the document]
VMois: Good idea! It makes the code cleaner

reana_server/reana_admin/check_workflows.py
```diff
@@ -30,34 +32,36 @@
...
 
-class WorkflowCheckFailed(Exception):
-    """Error when workflow doesn't pass some check."""
+class CheckFailed(Exception):
+    """Error when check fails."""
 
 
-WorkflowCheckResult = Tuple[Workflow, List[WorkflowCheckFailed]]
+CheckSource = namedtuple("CheckSource", "id name user status")
+CheckResult = Tuple[CheckSource, List[CheckFailed]]
```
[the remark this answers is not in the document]
VMois: Because I also want types, I replaced it with `dataclass` ([python docs](https://docs.python.org/3.8/library/dataclasses.html), [StackOverflow](https://stackoverflow.com/questions/34269772/type-hints-in-namedtuple)). Learned something new :)

VMois:
> Could you please add some tests for both workflows and sessions, checking the various scenarios?

Can you please provide examples of scenarios you want to test?

reana_server/reana_admin/check_workflows.py
```diff
@@ -165,54 +181,135 @@ def check_workflows(
             fg="red",
         )
         logging.exception(error)
-        sys.exit(1)
+        return [], [], total_workflows
```
mdonadoni, line 184: Returning empty arrays causes the command to print `OK` and exit with code zero. What do you think about throwing exceptions (or not catching them)?

reana_server/reana_admin/check_workflows.py
```diff
@@ -75,35 +93,35 @@ def _collect_messages_from_scheduler_queue(filtered_workflows: Query) -> Dict:
...
+        raise CheckFailed("Some pods still exist.")
 
 
 def _pods_exist(workflow, pods, scheduler_messages):
     if len(pods) == 0:
-        raise WorkflowCheckFailed("No pods found for the workflow.")
+        raise CheckFailed("No pods found.")
+
+
+def _only_one_pod_exist(workflow, pods, scheduler_messages):
```
mdonadoni, line 110:
```suggestion
def _only_one_pod_exists(workflow, pods, scheduler_messages):
```

reana_server/reana_admin/check_workflows.py
```diff
@@ -146,73 +164,144 @@ def check_workflows(
...
+    total_workflows = filtered_workflows.count()
+
+    if total_workflows == 0:
         click.secho(
-            f"No workflows found matching time range {time_range_message}.\nPlease, adjust your filter conditions.",
+            f"No workflows found matching time range {time_range_message}.",
             fg="red",
         )
-        sys.exit(1)
+        return [], [], total_workflows
```
mdonadoni, line 174: Before we were exiting with status code 1, while now we don't. Is this supposed to be treated as an error or not? What do you think?

reana_server/reana_admin/cli.py
```diff
@@ -736,24 +751,79 @@ def queue_consume(
...
+    sessions_in_sync = True
+    try:
+        (
+            in_sync_sessions,
+            out_of_sync_sessions,
+            pods_without_session,
+            total_sessions,
+        ) = check_interactive_sessions()
+    except FailedToCollect as error:
+        workflows_in_sync = False
```
mdonadoni, line 799:
```suggestion
        sessions_in_sync = False
```

reana_server/reana_admin/check_workflows.py
```diff
@@ -30,38 +31,55 @@
...
 
 
-class WorkflowCheckFailed(Exception):
-    """Error when workflow doesn't pass some check."""
+class CheckFailed(Exception):
+    """Error when a check for workflow or session fails."""
 
 
-WorkflowCheckResult = Tuple[Workflow, List[WorkflowCheckFailed]]
+class FailedToCollect(Exception):
```
mdonadoni, line 38:
```suggestion
class InfoCollectionError(Exception):
```

What do you think about this?

reana_server/reana_admin/check_workflows.py
```diff
@@ -146,73 +164,144 @@ def check_workflows(
...
 
     try:
         click.secho("Collecting MQ messages...")
         scheduler_messages = _collect_messages_from_scheduler_queue(filtered_workflows)
     except Exception as error:
-        click.secho(
-            "Error is raised when collecting scheduler messages.",
-            fg="red",
-        )
         logging.exception(error)
```
mdonadoni, line 183: What about moving all the `logging.exception(...)` in `check_workflows` and `check_interactive_sessions` to the main command function, where we are already catching and printing this exceptions? WDYT?

reana_server/reana_admin/check_workflows.py
VMois: It should be an error. If no workflows are found, it can mean that no workflows were run during that time. If we use this tool regularly, it might return "FAILED" when there were just zero workflows run.

reana_server/reana_admin/check_workflows.py
VMois: I removed click.secho in the `exception` clause as it is not helpful. Now, it will just log an exception that has traceback and message.

reana_server/reana_admin/check_workflows.py
```diff
@@ -146,73 +163,141 @@ def check_workflows(
...
+                check(session, filtered_pods, [])
+            except CheckFailed as error:
+                failed_checks.append(error)
+                session_is_in_sync = False
+                break
+
+        check_source = CheckSource(
+            id=str(session.id_),
+            name=session.name,
+            user=str(session.owner_id),
```
mdonadoni, line 294:
```suggestion
            user=str(session.workflow[0].owner.email),
```
WDYT? This is not the best, but it should work fine as a session is always attached to a workflow.

mdonadoni (approved):
Well done! Works as expected :smile:

PR #529, opened 2022-07-28
openapi: add missing schemas to error responses

BrunoRosendo:
closes https://github.com/reanahub/reana-client-go/issues/16

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -56,6 +56,14 @@
               "application/json": {
                 "message": "Internal server error."
               }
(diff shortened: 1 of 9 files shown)


VMois (approved):
Looks okay. Quite a lot of changes. It is hard to check every single one. I think we can merge them for now and see if, in the future, any issues will appear. Then we can fix them.

mdonadoni:
Just one last thing: could you please change `added` with `add` in the commit message?

PR #530, opened 2022-07-29
rest: validate REANA specification

### LINKED ISSUE #468: rest: start_workflow does not validate workflow
Currently, the `launch` endpoint is validating workflow:

https://github.com/reanahub/reana-server/blob/9d047fc9cac520dc8f539707390b8071e791e686/reana_server/rest/launch.py#L152

The `start_workflow` endpoint does not validate workflow. As far as I understood, workflow is checked before using reana-client CLI. But, is it enough? Creating this issue for future discussions.

VMois:
- before, "create_workflow" endpoint was not validating REANA specification; this commit adds validation.
- it also fixes some tests as they incorrectly put "workflow_name" in JSON body instead of query string.

closes #468

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -16,6 +16,8 @@ import requests
 from bravado.exception import HTTPError
 from flask import Blueprint, Response
 from flask import jsonify, request, stream_with_context
+from jsonschema.exceptions import ValidationError
+
 from reana_commons.config import REANA_WORKFLOW_ENGINES
 from reana_commons.errors import REANAQuotaExceededError, REANAValidationError
 from reana_commons.validation.operational_options import validate_operational_options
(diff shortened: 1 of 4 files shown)


reana_server/rest/workflows.py
```diff
@@ -468,6 +470,9 @@ def create_workflow(user):  # noqa
         if is_uuid_v4(workflow_name):
             return jsonify({"message": "Workflow name cannot be a valid UUIDv4."}), 400
 
+        validate_workflow(reana_spec_file, input_parameters={})
+        workspace_root_path = reana_spec_file.get("workspace", {}).get("root_path")
```
VMois, line 474: *note:* validation part is messy. `validate_workflow` also validates workspace_path, which is okay. But, for example, `validate_operational_options,` in addition to validating, is also formatting and returning the output so I cannot remove it.

audrium (approved):
Looks good! Needs a rebase and a changelog message

VMois:
I would wait to merge this PR. @audrium made a good point that right now, we will have validation happening on both the client and server sides. In addition, for example, users can now disable validation on the client side, but this PR makes it mandatory. It would be good to have a clear strategy for approaching the validation of workflows

PR #540, opened 2022-08-11
rest: add new default values to "info" endpoint output

VMois:
- add kubernetes default and max memory limits;
- add default workspace retention period.

closes reanahub/reana-client#622

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -380,9 +380,29 @@
                     "slurmcern"
                   ]
                 },
+                "default_kubernetes_jobs_timeout": {
+                  "title": "Default timeout for Kubernetes jobs",
+                  "value": "604800"
+                },
                 "default_workspace": {
                   "title": "Default workspace",
                   "value": "/usr/share"
+                },
+                "kubernetes_max_memory_limit": {
+                  "title": "Maximum allowed memory limit for Kubernetes jobs",
+                  "value": "10Gi"
+                },
+                "kubernetes_memory_limit": {
+                  "title": "Default memory limit for Kubernetes jobs",
+                  "value": "3Gi"
+                },
+                "maximum_kubernetes_jobs_timeout": {
+                  "title": "Maximum timeout for Kubernetes jobs",
+                  "value": "1209600"
+                },
+                "maximum_workspace_retention_period": {
+                  "title": "Maximum retention period in days for workspace files",
+                  "value": "3650"
                 },
                 "workspaces_available": {
                   "title": "List of available workspaces",
@@ -410,6 +430,28 @@
                   },
                   "type": "object"
                 },
+                "default_kubernetes_jobs_timeout": {
+                  "properties": {
+                    "title": {
+                      "type": "string"
+                    },
+                    "value": {
+                      "type": "string"
+                    }
+                  },
+                  "type": "object"
+                },
+                "default_kubernetes_memory_limit": {
+                  "properties": {
+                    "title": {
+                      "type": "string"
+                    },
+                    "value": {
+                      "type": "string"
+                    }
+                  },
+                  "type": "object"
+                },
                 "default_workspace": {
                   "properties": {
                     "title": {
@@ -417,6 +459,41 @@
                     },
                     "value": {
                       "type": "string"
+                    }
+                  },
+                  "type": "object"
+                },
+                "kubernetes_max_memory_limit": {
+                  "properties": {
+                    "title": {
+                      "type": "string"
+                    },
+                    "value": {
+                      "type": "string",
+                      "x-nullable": true
+                    }
+                  },
+                  "type": "object"
+                },
+                "maximum_kubernetes_jobs_timeout": {
+                  "properties": {
+                    "title": {
+                      "type": "string"
+                    },
+                    "value": {
+                      "type": "string"
+                    }
+                  },
+                  "type": "object"
+                },
+                "maximum_workspace_retention_period": {
+                  "properties": {
+                    "title": {
+                      "type": "string"
+                    },
+                    "value": {
+                      "type": "string",
+                      "x-nullable": true
                     }
                   },
                   "type": "object"
--- a/reana_server/rest/info.py
+++ b/reana_server/rest/info.py
@@ -16,7 +16,14 @@ from marshmallow import Schema, fields
 
 from reana_commons.config import DEFAULT_WORKSPACE_PATH, WORKSPACE_PATHS
 
-from reana_server.config import SUPPORTED_COMPUTE_BACKENDS
+from reana_server.config import (
+    SUPPORTED_COMPUTE_BACKENDS,
+    WORKSPACE_RETENTION_PERIOD,
+    REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT,
+    REANA_KUBERNETES_JOBS_MEMORY_LIMIT,
+    REANA_KUBERNETES_JOBS_TIMEOUT_LIMIT,
+    REANA_KUBERNETES_JOBS_MAX_USER_TIMEOUT_LIMIT,
+)
 from reana_server.decorators import signin_required
 
 blueprint = Blueprint("info", __name__)
@@ -45,34 +52,7 @@ def info(user, **kwargs):  # noqa
         200:
           description: >-
             Request succeeded. The response contains general info about the cluster.
-          schema:
-            type: object
-            properties:
-              workspaces_available:
-                type: object
-                properties:
-                  title:
-                    type: string
-                  value:
-                    type: array
-                    items:
-                      type: string
-              default_workspace:
-                type: object
-                properties:
-                  title:
-                    type: string
-                  value:
-                    type: string
-              compute_backends:
-                type: object
-                properties:
-                  title:
-                    type: string
-                  value:
-                    type: array
-                    items:
-                      type: string
+          schema: InfoSchema
           examples:
             application/json:
               {
@@ -91,7 +71,27 @@ def info(user, **kwargs):  # noqa
                         "htcondorcern",
                         "slurmcern"
                     ]
-              }
+                },
+                "kubernetes_memory_limit": {
+                    "title": "Default memory limit for Kubernetes jobs",
+                    "value": "3Gi"
+                },
+                "kubernetes_max_memory_limit": {
+                    "title": "Maximum allowed memory limit for Kubernetes jobs",
+                    "value": "10Gi"
+                },
+                "maximum_workspace_retention_period": {
+                    "title": "Maximum retention period in days for workspace files",
+                    "value": "3650"
+                },
+                "default_kubernetes_jobs_timeout": {
+                    "title": "Default timeout for Kubernetes jobs",
+                    "value": "604800"
+                },
+                "maximum_kubernetes_jobs_timeout": {
+                    "title": "Maximum timeout for Kubernetes jobs",
+                    "value": "1209600"
+                },
               }
         500:
           description: >-
@@ -108,7 +108,7 @@ def info(user, **kwargs):  # noqa
               }
     """
     try:
-        info = dict(
+        cluster_information = dict(
             workspaces_available=dict(
                 title="List of available workspaces",
                 value=list(WORKSPACE_PATHS.values()),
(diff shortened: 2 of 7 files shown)


reana_server/rest/info.py
```diff
@@ -120,8 +125,20 @@ def info(user, **kwargs):  # noqa
...
+                title="Maximum allowed memory limit for Kubernetes backend jobs",
+                value=REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT,
+            ),
+            default_workspace_retention_period=dict(
+                title="Default number of days to retain workspace files",
+                value=WORKSPACE_RETENTION_PERIOD,
+            ),
         )
-        return InfoSchema().dump(info)
+        return InfoSchema().dump(cluster_information)
```
mdonadoni, line 141: The OpenAPI specification needs to be updated to include the changes to the response

mdonadoni (commented):
Works well!

What do you think about improving also reana-client? In particular, it would be nice to print the  cluster info in a different order (e.g. keeping the default workspace and the list of available workspaces close together) and to print the health status of the cluster.

reana_server/rest/info.py
```diff
@@ -120,8 +125,20 @@ def info(user, **kwargs):  # noqa
...
+                value=REANA_KUBERNETES_JOBS_MEMORY_LIMIT,
+            ),
+            kubernetes_max_memory_limit=dict(
+                title="Maximum allowed memory limit for Kubernetes backend jobs",
+                value=REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT,
+            ),
+            default_workspace_retention_period=dict(
+                title="Default number of days to retain workspace files",
+                value=WORKSPACE_RETENTION_PERIOD,
+            ),
```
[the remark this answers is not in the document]
VMois: Yep. Totally forgot.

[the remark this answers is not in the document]
VMois: It will also require some changes in the `reana` repo to pass ENV to the reana-server.

docs/openapi.json
```diff
@@ -384,6 +384,18 @@
                   "title": "Default workspace",
                   "value": "/usr/share"
                 },
+                "kubernetes_max_memory_limit": {
+                  "title": "Maximum allowed memory limit for Kubernetes jobs",
+                  "value": "4Gi"
```
tiborsimko, line 389: 10Gi

It'll be good to put "realistic" numbers following `reana.cern.ch` typical deployment, so that users that happen to see these examples won't think it's the REANA defaults.

docs/openapi.json
```diff
@@ -384,6 +384,18 @@
...
+                },
+                "kubernetes_memory_limit": {
+                  "title": "Default memory limit for Kubernetes jobs",
+                  "value": "2Gi"
```
tiborsimko, line 393: 3Gi

docs/openapi.json
```diff
@@ -384,6 +384,18 @@
...
+                },
+                "maximum_workspace_retention_period": {
+                  "title": "Maximum retention period in days for workspace files",
+                  "value": "3"
```
tiborsimko, line 397: Again, 3 days would look too harsh here. Please put 3650 (=10 years) or even 36500 (=100 years) to express a really long time.

reana_server/rest/info.py
```diff
@@ -91,7 +123,19 @@ def info(user, **kwargs):  # noqa
                         "htcondorcern",
                         "slurmcern"
                     ]
-              }
+                },
+                "kubernetes_memory_limit": {
+                    "title": "Default memory limit for Kubernetes jobs",
+                    "value": "2Gi"
```
tiborsimko, line 129: 3Gi

reana_server/rest/info.py
```diff
@@ -91,7 +123,19 @@ def info(user, **kwargs):  # noqa
...
+                },
+                "kubernetes_max_memory_limit": {
+                    "title": "Maximum allowed memory limit for Kubernetes jobs",
+                    "value": "4Gi"
```
tiborsimko, line 133: 10Gi

reana_server/rest/info.py
```diff
@@ -91,7 +123,19 @@ def info(user, **kwargs):  # noqa
...
+                },
+                "maximum_workspace_retention_period": {
+                    "title": "Maximum retention period in days for workspace files",
+                    "value": "3"
```
tiborsimko, line 137: 3650 or 36500

reana_server/rest/info.py
```diff
@@ -136,3 +192,8 @@ class InfoSchema(Schema):
     )
     default_workspace = fields.Dict(keys=fields.Str(), values=fields.Str())
     compute_backends = fields.Dict(keys=fields.Str(), values=fields.List(fields.Str()))
+    kubernetes_memory_limit = fields.Dict(keys=fields.Str(), values=fields.Str())
+    kubernetes_max_memory_limit = fields.Dict(keys=fields.Str(), values=fields.Str())
+    maximum_workspace_retention_period = fields.Dict(
+        keys=fields.Str(), values=fields.Str()
+    )
```
mdonadoni, line 199: I think this schema is wrong for `workspaces_available` and `compute_backends` given that the values are both strings (for `title`) and lists (for `value`). What do you think about this new schema?

```python
class ListStringInfoValue(Schema):
    """Schema for a value represented by a list of strings."""
    title = fields.String()
    value = fields.List(fields.String())


class StringInfoValue(Schema):
    """Schema for a value represented by a string."""
    title = fields.String()
    value = fields.String(allow_none=True)


class InfoSchema(Schema):
    """Marshmallow schema for ``info`` endpoint."""

    workspaces_available = fields.Nested(ListStringInfoValue)
    default_workspace = fields.Nested(StringInfoValue)
    compute_backends = fields.Nested(ListStringInfoValue)
    kubernetes_memory_limit = fields.Nested(StringInfoValue)
    kubernetes_max_memory_limit = fields.Nested(StringInfoValue)
    maximum_workspace_retention_period = fields.Nested(StringInfoValue)
```

I also found out that we can use this schema to automatically generate the OpenAPI specification!
In particular:

1. Add the Marshmallow plugin in `scripts/generate_openapi_spec.py`

    ```diff
    -        plugins=("apispec.ext.flask",),
    +        plugins=("apispec.ext.flask", "apispec.ext.marshmallow"),
    ```

2. Use `InfoSchema` in the OpenAPI specification

    ```yaml
    responses:
      200:
        description: >-
          Request succeeded. The response contains general info about the cluster.
        schema: InfoSchema
        examples:
          ...
    ```

Should we try this approach?

VMois: You are right about schema. One thing from your example is that `default_workspace` is not nullable, but if this doesn't have a problem, we can use the schemas above.

I tried generating OpenAPI spec using schema. It looks pretty nice.

mdonadoni: > You are right about schema. One thing from your example is that `default_workspace` is not nullable, but if this doesn't have a problem, we can use the schemas above.

In this case we can use two different schemas, one with `allow_none=False` and the other with `allow_none=True`, if you prefer to be more precise!

reana_server/rest/info.py
```diff
@@ -45,34 +52,7 @@ def info(user, **kwargs):  # noqa
...
-              compute_backends:
-                type: object
-                properties:
-                  title:
-                    type: string
-                  value:
-                    type: array
-                    items:
-                      type: string
+          schema: InfoSchema
```
tiborsimko, line 55: Is this change OK with respect to the OpenAPI specification and client generation and such? I think we use explicit YAML everywhere for the other endpoints?

docs/openapi.json
```diff
@@ -380,10 +380,30 @@
...
+                  "title": "Default memory limit for Kubernetes jobs",
+                  "value": "3Gi"
+                },
+                "maximum_kubernetes_jobs_timeout": {
+                  "title": "Maximum timeout for Kubernetes jobs",
+                  "value": "1209600"
+                },
+                "maximum_workspace_retention_period": {
+                  "title": "Maximum retention period in days for workspace files",
+                  "value": "3650"
```
tiborsimko, line 405: BTW the default REANA installation leads to:

```console
$ reana-client info | grep None
Maximum allowed memory limit for Kubernetes jobs: None
Maximum retention period in days for workspace files: None
```

None may be perhaps leading to misinterpretations, but dunno whether we can do something better. Perhaps hide null values in the client? Just thinking out loud...

```console
$ reana-client info --json | jq
{
  "kubernetes_max_memory_limit": {
    "title": "Maximum allowed memory limit for Kubernetes jobs",
    "value": null
  },
  "maximum_kubernetes_jobs_timeout": {
    "title": "Maximum timeout for Kubernetes jobs",
    "value": "1209600"
  },
  "maximum_workspace_retention_period": {
    "title": "Maximum retention period in days for workspace files",
    "value": null
  }
}

```

VMois: I can detect if it is `None` and replace it with a string like `Unlimited maximum retention period` or `Keep forever`.

Example:

```json
 "maximum_workspace_retention_period": {
    "title": "Maximum retention period in days for workspace files",
    "value": "No limit. Keep forever",
  }
```

WDYT?

reana_server/rest/info.py
tiborsimko: If we generate the same OpenAPI, and Bravado is happy, then it should be good then! 

P.S. We might want to harmonise this later elsewhere for the other endpoints, just for consistency's sake, but this will take time.

docs/openapi.json
tiborsimko: Do you mean to keep `null` in JSON and replace `None` with a string such as "Not set" on the client-side only for the textual output? Or do you mean to massage the output already on the server-side so that a string such as "Not set" is returned by the server REST API side already? 

(BTW if we want to better expose default values to the user, instead of just saying "not set" for N endpoints and leaving the user to wonder how to detect default value, we could indeed generate nicer phrases such as "Keep forever" on the server-side, although they won't be very configurable then.  Alternatively, we could transform these special strings such as "forever" to special values such as 9999999999999 to represent infinity; in this way no unconfigurable strings would be necessary, but it would would be a bigger change that would go against "optionality" 
 of values and types a bit.)

VMois: I meant to change it on the server side. But it might as well work on the client side. Maybe, even better, cause if the client receives `null/None`, you can act on it knowing no limit exists and later display it to the user like `No limit`.

VMois: So we will keep JSON output with `null/None` and user response with a user-friendly message.

```console
> reana-client info
List of supported compute backends: kubernetes
Default timeout for Kubernetes jobs: 604800
Default memory limit for Kubernetes jobs: 4Gi
Default workspace: /var/reana
Maximum allowed memory limit for Kubernetes jobs: no limit
Maximum timeout for Kubernetes jobs: 1209600
Maximum retention period in days for workspace files: no limit
List of available workspaces: /var/reana
```

And if `--json` passed, it will return this:

```json
{"compute_backends": {"title": "List of supported compute backends", "value": ["kubernetes"]}, "default_kubernetes_jobs_timeout": {"title": "Default timeout for Kubernetes jobs", "value": "604800"}, "default_kubernetes_memory_limit": {"title": "Default memory limit for Kubernetes jobs", "value": "4Gi"}, "default_workspace": {"title": "Default workspace", "value": "/var/reana"}, "kubernetes_max_memory_limit": {"title": "Maximum allowed memory limit for Kubernetes jobs", "value": null}, "maximum_kubernetes_jobs_timeout": {"title": "Maximum timeout for Kubernetes jobs", "value": "1209600"}, "maximum_workspace_retention_period": {"title": "Maximum retention period in days for workspace files", "value": null}, "workspaces_available": {"title": "List of available workspaces", "value": ["/var/reana"]}}
```

VMois: @mdonadoni mentioned that it might be nice to group values together for better output ([link](https://github.com/reanahub/reana-server/pull/540#pullrequestreview-1069850777)). I can create an issue to improve this and, in addition, figure out this `null/None` and user-friendly message.

tiborsimko: WRT grouping, that we can leave for later. It should be enough if the values are sorted alphabetically and if similar symbols and near each other.  This is why I usually prefer common prefix rather than common suffix, e.g. `kubernetes_memory_limit_max` and `..._min`  vs ` max_kubernetes_memory_limit` and `min_...`.  In this way, the grouping is occurring naturally thanks to alphabetical ordering. (But it may not be practical to change the naming of all these symbols now if they are already exposed elsewhere...)

PR #543, opened 2022-08-24
openapi: add missing param schema do set_workflow_status

BrunoRosendo:
closes https://github.com/reanahub/reana-client-go/issues/93


tiborsimko (approved):
Works well :+1: See a minor behaviour observation for older clients in https://github.com/reanahub/reana-commons/pull/365

Issue #526, opened 2022-07-19
reana-admin: rules can become inactive while executing `retention-rules-apply`


mdonadoni:
When executing the `retention-rules-apply` command, workflows can be restarted even while retention rules are being applied.

This can become an issue: if a rule is fetched from the DB before it is disabled by the workflow's restart, the rule will be applied even though its status in the database has already become `inactive`.

We should decide whether this behaviour is a problem or not and, in case it is, we should discuss how to synchronize workflow restarts and the application of retention rules.

VMois:
Unless there is an easy fix for the MVP release, I would prefer not to bother much with this issue as the possibility of experiencing it is low. Why?

1. **Small window**. The user needs to restart the workflow exactly when the cron job is running. But even in this case, it is only *submission*, the workflow is not running yet (needs to get scheduled, etc.). The workflow might start after rules are applied, and nothing wrong happens for the user because we do not delete inputs. In addition, we do not have many workflows to process so the window can be tiny (< 30 seconds).

*side note:* maybe checking in a scheduler if it is okay to start a workflow might fix this problem.

2. **It happens only on a restart**. If a workflow is newly created, this issue doesn't exist. I don't know how often people restart their workflows, but if this number is low, the user has an even smaller probability of experiencing the issue.

PR #551, opened 2022-10-06
rest: disallow workflow restart if retention rules are pending

mdonadoni:
Closes https://github.com/reanahub/reana-server/issues/526


audrium (approved):
Works great!

```
reana ❯ rc restart -w workflow
==> ERROR: Cannot start workflow workflow:
The workflow cannot be restarted because some retention rules are currently being applied to the workspace. Please retry later.
```

PR #567, opened 2022-12-01
rest: validate inputs on workflow creation

audrium:
closes https://github.com/reanahub/reana-client/issues/626

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -32,7 +32,11 @@ from webargs.flaskparser import use_kwargs
 from reana_server.api_client import current_rwc_api_client
 from reana_server.config import REANA_HOSTNAME
 from reana_server.decorators import check_quota, signin_required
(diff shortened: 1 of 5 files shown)


reana_server/rest/workflows.py
```diff
@@ -531,6 +535,8 @@ def create_workflow(user):  # noqa
         workspace_root_path = reana_spec_file.get("workspace", {}).get("root_path")
         validate_workspace_path(reana_spec_file)
 
+        validate_inputs(reana_spec_file)
```
tiborsimko, line 538: We should probably prettify the error messages, because currently the user gets:

```
Invalid input paths: 'code' is a prefix of 'code/helloworld.py'
```

This does not look very user-friendly, e.g. the path is valid, not invalid.

We should say something like:

```
Duplicate input paths ''code' and 'code/helloworld.py' found. Please deduplicate inputs first.
```

And the "invalid path" message should be reserved for those user declarations that don't point to any actually existing file.

PR #568, opened 2022-12-05
scripts: add openapi checker

mdonadoni:
Closes reanahub/reana#680


mdonadoni:
As this was mostly a prototype:
- What do you think about it?
- Are the returned error/warnings useful?
- Are there more checks we should implement?
- Regarding the checks that are already implemented, are they working well enough or do you think that there can be too many false positives/negatives?
- Should we change approach (e.g. do not use AST but something else)?

Finally, where should we put this script? In every repo like `generate_openapi_spec.py` , as part of `reana-dev`, somewhere else?

PR #574, opened 2023-02-02
admin: improve disk usage summary in status report

### LINKED ISSUE #573: cronjob: better disk reporting terminology in system status reports
**Current behaviour**

Currently, the nightly system status report uses language such as:

```
STORAGE
user_directory_size: 3.1T
shared_volume_health: 3.1T/4.5T (41%)
```

This terminology may be misleading: is 3.1 total size? is 3.1 out of 4.5 supposed to be 41%? Also, is it necessary to repeat 3.1T in two lines?

**Expected behaviour**

Let's modify the reporting verbiage to be more conform to the terminology used for example in the `df` command: 

```
Filesystem      Size  Used Avail Use% Mounted on
ceph-fuse       7.6T  3.1T  4.5T  41% /var/reana
```

This would give something like:

```
STORAGE
user_directory_size: 7.6T
shared_volume_health: 3.1T/7.6T (41%)
```

Or, simply, we could only report shared_volume_health line (note modified denominator to include total volume size), because the latter line already includes all the information from the former line.

mdonadoni:
This commit improves how the disk usage summary is formatted in the
system status report. It also removes the size of the users' directory
from the report, as this information is redundant but it takes a lot of
time to be collected.

Closes #573

### DIFF
--- a/reana_server/status.py
+++ b/reana_server/status.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2020, 2021, 2022 CERN.
+# Copyright (C) 2020, 2021, 2022, 2023 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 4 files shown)


reana_server/status.py
```diff
@@ -173,23 +173,15 @@ def users_directory_size(self):
...
-            f"({values[use_percentage_index]})"
-        )
+        used_size, total_size, used_percentage = output[1].split()
+
+        return f"{used_size}/{total_size} ({used_percentage})"
 
     def get_status(self):
         """Get status summary for REANA storage."""
         return {
-            "user_directory_size": self.users_directory_size(),
```
mdonadoni, line 192: Should I also delete the `users_directory_size()` function? As far as I can tell, it is not used anywhere after these changes.

audrium (approved):
Nicely done, works as expected

edit: maybe worth adding a note to the changelog?

reana_server/status.py
audrium: I think we should remove it together with `_get_path()`, since as you mentioned it's not being used

PR #577, opened 2023-02-08
reana-admin: add interactive-session-cleanup command

audrium:
Add command to close interactive sessions running longer than a specified number of days.

closes https://github.com/reanahub/reana-workflow-controller/issues/498

### DIFF
--- a/reana_server/reana_admin/cli.py
+++ b/reana_server/reana_admin/cli.py
@@ -17,12 +17,19 @@ from typing import List, Optional
 
 import click
 import tablib
+import requests
 from flask.cli import with_appcontext
 from invenio_accounts.utils import register_user
-from reana_commons.config import REANAConfig, REANA_RESOURCE_HEALTH_COLORS
+from kubernetes.client.rest import ApiException
+from reana_commons.config import (
+    REANAConfig,
+    REANA_RESOURCE_HEALTH_COLORS,
+    REANA_RUNTIME_KUBERNETES_NAMESPACE,
+)
 from reana_commons.email import send_email
 from reana_commons.errors import REANAEmailNotificationError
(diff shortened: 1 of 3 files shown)


reana_server/reana_admin/cli.py
```diff
@@ -945,3 +955,94 @@ def check_workflows(
...
+            click.secho(
+                f"Couldn't parse user details from '{pod_name}' session metadata: {e}",
+                fg="red",
+                err=True,
+            )
+            continue
+        host = REANA_HOSTNAME or "172.17.0.1:30443"
+        try:
+            session_status = requests.get(
+                f"https://{host}/{workflow_id}/api/status?token={token}", verify=False
```
mdonadoni, line 1009: I don't really have a solution to this, but I don't like that we are making a lot of assumptions on how interactive sessions are deployed and configured (e.g. the token is passed as a command line argument, the status is accessible at `<workflow id>/api/status`...). This is somewhat made worse by the fact that some places of r-w-controller are ready to support multiple different session types, while here (and in other places of REANA) we assume that all the sessions are jupyter notebooks.

What is your opinion on having a new endpoint in r-w-controller to get the status of an interactive session? Given that we manage the deployment of the interactive sessions in r-w-controller, I suspect it might be easier to get the status of the session there, as all the configuration details should be known (e.g. port used by the service etc.)

mdonadoni: This would probably require some important changes to r-w-controller, though

mdonadoni (approved):

reana_server/reana_admin/cli.py
audrium: Created an issue to address this https://github.com/reanahub/reana-workflow-controller/issues/504

Issue #513, opened 2022-07-06
login: support for generic OIDC Keycloak


tiborsimko:
Investigate supporting ESCAPE OIDC https://iam-escape.cloud.cnaf.infn.it/

This could be done by creating a new "contrib" client in the `invenio-oauthclient` package, or using generalised Keycloak client similarly to described in https://inveniordm.docs.cern.ch/customize/authentication/#keycloak

We will then need then to expose login configuration options to REANA cluster administrators during Helm installation, and amend `reana-ui` package to offer appropriate login link (or more than one login link!).

- [x] Verify that this works with the currently used invenio version if not an upgrade is necessary
- [x] First, implement a hardcoded SSO login for IAM ESCAPE as a test
- [x] Invenio OAuth client might require changes in its source code as realm is not used for IAM ESCAPE

This issue is related to: https://github.com/reanahub/reana-server/issues/564

goseind:
To implement this, `reana-server`, `reana`, and `reana-ui` need to be enhanced in order to support another login method, this means:

- [ ] Adding helm values to pass on information on the third-party IAM service as environment variables to server and UI ref. https://github.com/reanahub/reana/pull/692
- [ ] Import the `keycloak` contrib from Invenio into the server config and add this as an option ref. https://github.com/reanahub/reana-server/pull/578
- [ ] Additionally as mentioned above some IAM services do not have a realm structure and thus the `keycloak` contrib should be adjusted to also work without realms ref. https://github.com/inveniosoftware/invenio-oauthclient/issues/293
- [ ] Add an additional button to UI sign-in page ref. https://github.com/reanahub/reana-ui/pull/311

PR #583, opened 2023-03-17
setup: pin `invenio-i18n` and `invenio-theme` versions

mdonadoni:
This commit downgrades `invenio-i18n` and `invenio-theme`, which were
updated to v2 in 0.9.1a1 (see #581). This solves issues with the
upcoming support to Keycloak SSO (see #513).


audrium (approved):
LGTM!

Issue #489, opened 2022-05-06
admin: standardize `--admin-access-token` usage


mdonadoni:
Some of the `reana-admin` commands require the admin access token to be provided by the user (e.g. [token-grant](https://github.com/reanahub/reana-server/blob/a42b1efeb722c19181d6fe016c0e308cbafa134c/reana_server/reana_admin.py#L215)), but not all of them. We should decide whether `reana-admin` should ask for the admin access token, and then make sure each command does the same thing.

PR #595, opened 2023-06-02
admin: standardize --admin-access-token usage

audrium:
closes https://github.com/reanahub/reana-server/issues/489

### DIFF
--- a/reana_server/decorators.py
+++ b/reana_server/decorators.py
@@ -10,45 +10,17 @@
 
 import functools
 import logging
-import os
 import traceback
-import sys
 
-import click
 from flask import jsonify, request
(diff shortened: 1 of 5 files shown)


reana_server/decorators.py
mdonadoni, line 1: We should consider moving `admin_access_token_options` to the `reana_admin` directory, as it is only used for the admin CLI tool.

mdonadoni (commented):
Works well!

Could you please also update the Helm Chart to provide the access token to the cronjobs, and the docs to update the command usage examples?

giuseppe-steduto (approved):
LGTM and works as expected!

PR #598, opened 2023-06-05
gitlab: use `url_for` to build API URLs

### LINKED ISSUE #593: gitlab: repositories have repeated webhooks
Reported when REANA is deployed on Rancher with HAProxy.
REANA is accessible over HTTPS through HAProxy. However, HAProxy connects to REANA over HTTP.
REANA is connected to a custom instance of GitLab different from https://gitlab.cern.ch .

GitLab projects are always reported as "disabled" in the Profile page of REANA-UI (radio button is off), even when the integration with that repository has been enabled in the past. Trying again to enable a repository results in multiple webhooks being created (this can be checked in Settings -> Webhooks of the repository). This also results in the workflow not being executed on REANA when a new commit is pushed to the repository.

Probably related to how we construct URLs in `gitlab.py`, as we construct them in different ways:

https://github.com/reanahub/reana-server/blob/fe1cbed2098afbd48f7ad8a562e5efbc9808899a/reana_server/utils.py#L516

https://github.com/reanahub/reana-server/blob/fe1cbed2098afbd48f7ad8a562e5efbc9808899a/reana_server/rest/gitlab.py#L385

This might be related to #590

mdonadoni:
Use `url_for` to build API URLs. Before, URLs were built in different
ways when creating and retrieving GitLab's webhooks, which sometimes
resulted in having different URLs for the same `create_workflow`
endpoint. This caused already created webhooks not to be recognized by
REANA, as a wrong URL was used to identify them.

Closes #593

### DIFF
--- a/reana_server/rest/gitlab.py
+++ b/reana_server/rest/gitlab.py
@@ -30,7 +30,6 @@ from reana_server.config import (
     REANA_GITLAB_OAUTH_APP_ID,
     REANA_GITLAB_OAUTH_APP_SECRET,
     REANA_GITLAB_URL,
-    REANA_HOSTNAME,
 )
 from reana_server.decorators import signin_required
 from reana_server.utils import (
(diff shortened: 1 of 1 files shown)


mdonadoni:
Tested on DEV, working as expected

tiborsimko (approved):

PR #599, opened 2023-06-06
rest: add max inactivity period for interactive sessions to info endpoint

giuseppe-steduto:
Adds the `maximum_interactive_session_inactivity_period` value to the ones returned by `/api/info`.

Addresses reanahub/reana-client#657

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -455,6 +455,18 @@
                   },
                   "type": "object"
                 },
+                "maximum_interactive_session_inactivity_period": {
+                  "properties": {
+                    "title": {
+                      "type": "string"
+                    },
+                    "value": {
+                      "type": "string",
+                      "x-nullable": true
+                    }
+                  },
+                  "type": "object"
+                },
                 "maximum_kubernetes_jobs_timeout": {
                   "properties": {
                     "title": {
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -207,6 +207,7 @@ PROXYFIX_CONFIG = {"x_proto": 1}
 APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {}
 APP_HEALTH_BLUEPRINT_ENABLED = False
 
+
 # Rate limiting configuration using invenio-app
 # ===========================
 
@@ -328,11 +329,9 @@ REANA_GITLAB_OAUTH_APP_SECRET = os.getenv("REANA_GITLAB_OAUTH_APP_SECRET", "CHAN
 REANA_GITLAB_HOST = os.getenv("REANA_GITLAB_HOST", None)
 REANA_GITLAB_URL = "https://{}".format((REANA_GITLAB_HOST or "CHANGE ME"))
 
-
 # Email configuration
 # ===================
 ADMIN_EMAIL = os.getenv("REANA_EMAIL_SENDER", "CHANGE_ME")
-
 
 # Workflow scheduler
 # ==================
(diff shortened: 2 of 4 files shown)


docs/openapi.json
```diff
@@ -455,6 +455,17 @@
                   },
                   "type": "object"
                 },
+                "maximum_interactive_session_inactivity_period": {
```
audrium, line 458: Since you modified the OpenAPI specification you have to call the `generate_openapi_spec.py` script to publish these changes to `reana-commons` (and create another PR there)

reana_server/rest/info.py
```diff
@@ -108,6 +108,7 @@ def info(user, **kwargs):  # noqa
               }
     """
     try:
+        ui_config = REANAConfig.load("ui")
```
audrium, line 111: Have you considered injecting inactivity period value through [Helm templates](https://github.com/reanahub/reana/blob/master/helm/reana/templates/reana-server.yaml#L87) similarly as it is done for `REANA_KUBERNETES_JOBS_MAX_USER_TIMEOUT_LIMIT`?. Now it's a bit strange to get it from the UI configmap

audrium (commented):
It works nicely! I left couple of small comments how it could be potentially improved.
Can you please modify the PR/commit messages to use `closes` instead of `addresses` since the PR is not properly linked to the issue in the kanban board. I had to dig a bit to find it

docs/openapi.json
```diff
@@ -455,6 +455,17 @@
...
+                  "properties": {
+                    "title": {
+                      "type": "string"
+                    },
+                    "value": {
+                      "type": "string"
```
audrium, line 464: Since this field could have a `None` value it should include `"x-nullable": true` flag similarly as for `kubernetes_max_memory_limit` field

Please not that after this change `reana-client-go` swagger specs will have to be regenerated

reana_server/config.py
```diff
@@ -389,6 +388,19 @@ def _get_rate_limit(env_variable: str, default: str) -> str:
 DEFAULT_WORKSPACE_RETENTION_RULE = "**/*"
 """Workspace retention rule which will be applied to all the workflows by default."""
 
+# Interactive sessions configuration
+# ==================
+_reana_session_max_inactivity_period_env = os.getenv(
+    "REANA_SESSION_MAX_INACTIVITY_PERIOD", "forever"
+)
+if _reana_session_max_inactivity_period_env == "forever":
+    REANA_SESSION_MAX_INACTIVITY_PERIOD: Optional[str] = None
```
audrium, line 397: Small suggestion: perhaps it would be better to keep a long and descriptive name and use `REANA_INTERACTIVE_SESSION` prefix for this value so that it wouldn't be confused to other type of sessions reana might have? WDYT?

reana_server/rest/info.py
```diff
@@ -180,3 +185,4 @@ class InfoSchema(Schema):
     maximum_workspace_retention_period = fields.Nested(StringNullableInfoValue)
     default_kubernetes_jobs_timeout = fields.Nested(StringInfoValue)
     maximum_kubernetes_jobs_timeout = fields.Nested(StringInfoValue)
+    maximum_interactive_session_inactivity_period = fields.Nested(StringInfoValue)
```
audrium, line 188: It should be `StringNullableInfoValue` because it might containt `None` value

audrium (commented):
Looks really nice now. The only two missing things now are to 
1) update the `r-commons` PR with the missing `x-nullable` option 
2) and to update the `reana-client-go` PR accordingly.

mdonadoni:
Works well, but this needs to be rebased!

mdonadoni (approved):

PR #604, opened 2023-07-13
docker: podman-friendly image locations

tiborsimko:
Adds fully qualified canonical locations of container images, making the container technology setup podman-friendly.

Closes reanahub/reana#729.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -2,9 +2,15 @@ codecov:
   require_ci_to_pass: yes
 
 coverage:
-  precision: 2
-  round: down
-  range: "80...100"
+  status:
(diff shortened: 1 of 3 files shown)


mdonadoni (approved):
Should we change also the docker image in `run-tests.sh`? This applies also to the other repositories
https://github.com/reanahub/reana-server/blob/c45927005c5a00942fd5c893759d17b0bbe3006a/run-tests.sh#L46
https://github.com/reanahub/reana-server/blob/c45927005c5a00942fd5c893759d17b0bbe3006a/run-tests.sh#L92-L98

mdonadoni (approved):

PR #608, opened 2023-07-20
global: update to apispec v3

mdonadoni:
Update to apispec v3, as the version used before is not compatible with
PyYAML v6. This commit also removes the marshmallow plugin for apispec,
as the generated OpenAPI spec is different and causes issues with
reana-client. This is due to the fact that the OpenAPI specifications of
marshmallow schemas are now put inside the global `definitions` object,
instead of being inlined in the endpoint's specification as before.

Closes reanahub/reana-commons#399

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1,11 +1,9 @@
 {
-  "definitions": {},
   "info": {
     "description": "Submit workflows to be run on REANA Cloud",
     "title": "REANA Server",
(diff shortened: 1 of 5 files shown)


mdonadoni:
CI should pass after releasing new alpha version of reana-commons and pytest-reana

giuseppe-steduto (approved):

PR #611, opened 2023-08-08
launch: include validation warnings in returned message

giuseppe-steduto:
In case some warnings are issued when launching a workflow, the `launch` endpoint will now include them in the returned `message`. 
Closes reanahub/reana-client#660.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Changes
 Version 0.9.1 (UNRELEASED)
 --------------------------
 
+- Changes ``launch`` endpoint also include the warnings of the validation of the workflow specification.
 - Changes OpenAPI specification with respect to return the maximum inactivity time before automatic closure of interactive sessions in ``info`` endpoint.
 - Adds the timestamp of when the workflow was stopped (``run_stopped_at``) to the workflow list and the workflow status endpoints.
 - Adds the content of the ``REANA_GITLAB_HOST`` environment variable to the list of GitLab instances from which it is possible to launch a workflow.
--- a/reana_server/rest/launch.py
+++ b/reana_server/rest/launch.py
@@ -104,6 +104,10 @@ def launch(user, url, name="", parameters="{}", specification=None):
             Request succeeded. Information of the workflow launched.
           schema:
             type: object
+            required:
+              - workflow_id
+              - workflow_name
+              - message
             properties:
               workflow_id:
                 type: string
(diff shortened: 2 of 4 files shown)


reana_server/rest/launch.py
```diff
@@ -218,10 +218,18 @@ def launch(user, url, name="", parameters="{}", specification=None):
...
+                "warnings were found during validation. "
+                f"{' '.join([x.message for x in validation_warnings])}"
+            )
+        else:
+            message = "The workflow has been successfully submitted."
         response_data = {
             "workflow_id": workflow.id_,
             "workflow_name": workflow.name,
-            "message": "The workflow has been successfully submitted.",
+            "message": message,
```
mdonadoni, line 232: Maybe we can serialise the warnings in a new `validation_warnings` (name can be changed) field of the reply, so that we can keep it structured? Otherwise it is impossible client/frontend-side to know whether there were warnings when validating the specification without parsing the message.

Something like
```json
{
  "...": "...",
  "validation_warnings": {
    "additional_properties": ["x", "y", "z"]
  }
}
```

We can discuss this in IRL, too.

CHANGES.rst
```diff
@@ -4,6 +4,7 @@ Changes
 Version 0.9.1 (UNRELEASED)
 --------------------------
 
+- Changes OpenAPI specification to include the optional ``validation_warnings`` key in the response of the ``launch`` endpoint.
```
mdonadoni, line 7: Given that not only the OpenAPI spec has changed, but also what the endpoint returns, I would say something like `Changes ``launch`` endpoint to also include the warnings of the validation of the workflow specification`. What do you think?

reana_server/rest/launch.py
```diff
@@ -104,13 +104,22 @@ def launch(user, url, name="", parameters="{}", specification=None):
...
                 type: string
               workflow_name:
                 type: string
               message:
                 type: string
+              validation_warnings:
+                description: >-
+                    Dictionary of validation warnings, if any. Each
+                    key is a property that was not correctly validated.
+                type: object
```
mdonadoni, line 122: What do you think about specifying this more, so that it is easily usable also in r-client-go? With the current specification, `validation_warnings` is a `interface{}` (note that currently we do not use the launch endpoint in the clients, but I think it's still a good idea to have precise OpenAPI specs)

reana_server/validation.py
```diff
@@ -107,7 +107,7 @@ def validate_inputs(reana_yaml: Dict) -> None:
             )
 
 
-def validate_workflow(reana_yaml: Dict, input_parameters: Dict) -> None:
+def validate_workflow(reana_yaml: Dict, input_parameters: Dict) -> List[Exception]:
```
mdonadoni, line 110: I think this should be `Dict[...]` and not `List[Exception]`

mdonadoni (approved):

PR #619, opened 2023-08-24
not send emails notifications disabled

### LINKED ISSUE #618: emails: sender email is used as receiver
Currently, the `REANA_EMAIL_SENDER` email is saved in `ADMIN_EMAIL`:

https://github.com/reanahub/reana-server/blob/b72bc727adaa87e371d982b5e9f20f9361ecbc12/reana_server/config.py#L334

However, `ADMIN_EMAIL` is then used as recipient of some emails:
https://github.com/reanahub/reana-server/blob/b72bc727adaa87e371d982b5e9f20f9361ecbc12/reana_server/rest/users.py#L324

We should use instead the `notifications.email_config.receiver` Helm value

giuseppe-steduto:
- cli: warn about disabled notifications when sending the status report
  Closes reanahub/reana#736
- users: send token request email to receiver address
  Closes #618


reana_server/config.py
```diff
@@ -332,6 +332,12 @@ def _get_rate_limit(env_variable: str, default: str) -> str:
 # Email configuration
 # ===================
 ADMIN_EMAIL = os.getenv("REANA_EMAIL_SENDER", "CHANGE_ME")
+REANA_NOTIFICATIONS_ENABLED = bool(
+    strtobool(os.getenv("REANA_NOTIFICATIONS_ENALBED", "False"))
+)
+REANA_NOTIFICATIONS_EMAIL_RECEIVER = os.getenv(
+    "REANA_NOTIFICATIONS_EMAIL_RECEIVER", "CHANGE_ME"
+)
```
mdonadoni, line 340: What do you think about putting these variables in reana-commons in `email.py`?

Also, what do you think about using `REANA_EMAIL_RECEIVER` instead of `REANA_NOTIFICATIONS_EMAIL_RECEIVER`, to be consistent with `REANA_EMAIL_SENDER`?

Finally, let's also remove `ADMIN_EMAIL` and let's use `REANA_EMAIL_SENDER` instead?

reana_server/reana_admin/cli.py
```diff
@@ -419,7 +424,13 @@ def _format_statuses(type_, statuses):
 
         if email:
             send_email(email, f"{hostname} system status report", status_report_body)
-            click.echo(f"Status report successfully sent by email to {email}.")
+            if REANA_NOTIFICATIONS_ENABLED:
```
mdonadoni, line 427: If we decide to throw an exception instead of just writing to the logs, then we do not need this check

mdonadoni (approved):

mdonadoni:
Could you please bump the lower version of reana-commons in `setup.py` and `requirements.txt`?

PR #628, opened 2023-09-16
reana-admin: fix setting storage quota for users after global default…

### LINKED ISSUE #625: reana-admin: fix setting storage quota for users after global default updates
**Current behaviour**

Consider A REANA deployment where there are no user quotas for CPU or Disk usage. (This is default behaviour.)

Consider that some users were consuming a lot of disk space and so individual quotas were introduced by the admins to these heavy users.

Now the REANA administrators would like to introduce a new global disk quota limit for all users.

We offer a function for this, `reana-admin quota-set-default-limits`.

However it seems to work well for setting quotas for newly created users, but updating quotas of those past users that did not have any previously set limits does not seem to work. 

**Expected behaviour**

If a user already had a quota limit before the admin change, the old individual value of the user should be preserved.

If a user did not have any quota limit, then the new default quota limit should be set.

If a new user is created in the system, then the user should pick up the new default quota limit.

**How to reproduce**

As the admin, create two local users:

```
$ kubectl exec -i -t deployment/reana-server -- flask reana-admin user-create -e aaa@example.org --user-access-token aaa --admin-access-token $REANA_ACCESS_TOKEN
$ kubectl exec -i -t deployment/reana-server -- flask reana-admin user-create -e bbb@example.org --user-access-token bbb --admin-access-token $REANA_ACCESS_TOKEN
```

As the user aaa, launch some workflows.

As the admin, update user quotas:

```
$ kubectl create job --from cronjob/reana-resource-quota-update force-quota-update2
```

As the user aaa, check disk usage from the user point of view:

```
$ reana-client quota-show --resource disk
184320 used
```

As the admin, set some limit to the first aaa user:

```
$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -e aaa@example.org -r disk -l 200000 --admin-access-token $REANA_ACCESS_TOKEN
```

As the user aaa, check whether it worked:

```
$ reana-client quota-show --resource disk
184320 out of 200000 used (92%)
```

As the admin, apply global quota to all users:

```
$ git diff | cat
diff --git a/helm/reana/values.yaml b/helm/reana/values.yaml
index 6cccfed..f5da0e8 100644
--- a/helm/reana/values.yaml
+++ b/helm/reana/values.yaml
@@ -161,6 +161,7 @@ traefik:
 # Quota
 quota:
   enabled: true
+  default_disk_limit: 100000
   periodic_update_policy: "{{ .Values.quota.disk_update }}"
   workflow_termination_update_policy: "{{ .Values.quota.termination_update_policy }}"
   # backward compatibility

$ helm upgrade reana helm/reana -f helm/configurations/values-dev.yaml
```

As the admin, check that limits are well visible in the infrastructure pods:

```
$ kubectl delete pod reana-server-64d9c9d9f9-d9694
$ kubectl delete pod reana-workflow-controller-5574fd9db8-gx6hl
$ kubectl exec -i -t reana-server-64d9c9d9f9-flvdp -c rest-api -- /bin/bash -c 'env | grep DISK'
REANA_DEFAULT_QUOTA_DISK_LIMIT=100000
```

As the user aaa and bbb, check the statutes:

```
$ reana-client quota-show --resource disk
184320 out of 200000 used (92%)

$ reana-client quota-show --resource disk -h
180 KiB used
```

Whoops! The limit is not applied to the user bbb.

However this is OK, because it behaves just as our docs say:

    If the default quota values are changed, they will only be applied to newly
    created users. Users that already exist will not be affected by this change.

Let's test it. An admin, let's create a new user ccc:

```
$ kubectl exec -i -t deployment/reana-server -- flask reana-admin user-create -e ccc@example.org --user-access-token ccc --admin-access-token $REANA_ACCESS_TOKEN
```

As the user ccc, check the quota:

```
$ reana-client quota-show --resource disk
0 out of 100000 used (0%)
```

It works nicely indeed.

Now REANA offers updating quotas of users:

```
$ kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set-default-limits --help
Defaulted container "rest-api" out of: rest-api, scheduler
Usage: flask reana-admin quota-set-default-limits [OPTIONS]

  Set default quota limits to users that do not have any.

Options:
  --admin-access-token TEXT  The access token of an administrator.  [required]
  --help                     Show this message and exit.
```

The tasks related to this issue are:

1. Run the above command and check whether the quota limit is well introduced to the user bbb.

2. If it is not, check the code responsible for this functionality in r-server.

Hint: The culprit might be in the query that should find all the users without the desired resource limit, but it does not in reality:

```python
users_without_quota_limits = User.query.filter(~User.resources.any()).all()
```

3. Test scenarios when both CPU and Disk limit are set, or when only the CPU default limit is set, prior or after the admin intervention.

DaanRosendal:
**Consideration**: the code assumes a default quota limit of zero. This assumption can lead to unexpected behaviour for admins when they change their default quota limit from a previously defined value to a new one. This is because any previously set default limits will be treated as custom limits and thus won't be automatically updated to match the new default limit.

I tested several scenarios:

**Single User Scenarios**
| Scenario                      | CPU Limit | Disk Limit | Outcome                    |
| ----------------------------- | --------- | ---------- | --------------------------- |
| 1 | 0         | Custom     | ✅ CPU updated to default      |
| 2 | Custom    | 0          | ✅ Disk updated to default     |
| 3      | 0         | 0          | ✅ Disk and CPU updated to default     |

**Multiple Users Scenarios**
| Scenario                                      | Outcome                  |
| --------------------------------------------- | ------------------------- |
| 1. Multiple Users (No Custom Quota Limits, all limits = 0)       | ✅ All updated to default   |
| 2. Multiple Users (All default Quota Limits)       | ✅ None updated             |
| 3. Multiple Users (All custom Quota Limits)       | ✅ None updated             |
| 4. CPU: 0, Disk: Custom (All Users)               | ✅ CPU updated for all users |
| 5. CPU: Custom, Disk: 0 (All Users)              | ✅ Disk updated for all users|
| 6. User 1: CPU: 0 for CPU and disk, User 2: Custom, User 3: Default| ✅ User 1 updated to default|

Closes #625

### DIFF
--- a/reana_server/reana_admin/cli.py
+++ b/reana_server/reana_admin/cli.py
@@ -16,20 +16,20 @@ import traceback
 from typing import List, Optional
 
 import click
-import tablib
 import requests
+import tablib
 from flask.cli import with_appcontext
 from invenio_accounts.utils import register_user
 from kubernetes.client.rest import ApiException
 from reana_commons.config import (
-    REANAConfig,
     REANA_RESOURCE_HEALTH_COLORS,
     REANA_RUNTIME_KUBERNETES_NAMESPACE,
+    REANAConfig,
 )
-from reana_commons.email import send_email, REANA_EMAIL_SENDER
+from reana_commons.email import REANA_EMAIL_SENDER, send_email
 from reana_commons.errors import REANAEmailNotificationError
-from reana_commons.utils import click_table_printer
 from reana_commons.k8s.api_client import current_k8s_corev1_api_client
+from reana_commons.utils import click_table_printer
 from reana_db.config import DEFAULT_QUOTA_LIMITS
 from reana_db.database import Session
 from reana_db.models import (
@@ -46,6 +46,7 @@ from reana_db.models import (
 )
 from reana_db.utils import update_workspace_retention_rules
 
+from reana_server.api_client import current_rwc_api_client
 from reana_server.config import ADMIN_USER_ID, REANA_HOSTNAME
 from reana_server.reana_admin.check_workflows import check_workspaces
 from reana_server.reana_admin.options import (
@@ -55,8 +56,8 @@ from reana_server.reana_admin.options import (
 )
 from reana_server.reana_admin.retention_rule_deleter import RetentionRuleDeleter
 from reana_server.status import STATUS_OBJECT_TYPES
-from reana_server.api_client import current_rwc_api_client
 from reana_server.utils import (
+    JinjaEnv,
     _create_user,
     _export_users,
     _get_user_by_criteria,
@@ -65,7 +66,6 @@ from reana_server.utils import (
     _validate_email,
     _validate_password,
     create_user_workspace,
-    JinjaEnv,
 )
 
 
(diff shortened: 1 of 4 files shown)


reana_server/reana_admin/cli.py
```diff
@@ -638,24 +638,46 @@ def set_quota_limit(
...
+            ).first()
+
+            if user_resource and user_resource.quota_limit == 0:
+                # If no limit exists, set the default limit
+                ctx.invoke(
+                    set_quota_limit,
+                    emails=[user.email],
+                    resource_name=resource.name,
+                    resource_type=resource.type_.name,
+                    limit=DEFAULT_QUOTA_LIMITS.get(resource.type_.name),
```
tiborsimko, line 678: Works nicely, but perhaps we should check whether the setting is actually needed here, because otherwise some unnecessary operations would be called:

```
Quota limit 0 for 'cpu (processing time)' successfully set to users ['aaa@example.org'].                                       │
Quota limit 0 for 'cpu (processing time)' successfully set to users ['bbb@example.org'].
```

This populates the table with zeros which may not be necessary.  (Even though later upgrade to a real value would well.)

reana_server/reana_admin/cli.py
```diff
@@ -901,10 +923,10 @@ def check_workflows(
 ) -> None:
     """Check consistency of selected workflow run statuses between database, message queue and Kubernetes."""
     from .check_workflows import (
-        check_workflows,
+        InfoCollectionError,
         check_interactive_sessions,
+        check_workflows,
```
tiborsimko, line 928: We can sync in the team about using common import sorting technique across editors.

tests/test_cli.py
```diff
@@ -968,3 +970,33 @@ def test_check_workspaces(
         if workflow:
             assert len(result.errors) == 2
             assert any(workflow.workspace_path in str(error) for error in result.errors)
+
+
+def test_quota_set_default_limits_with_user_with_custom_limits(default_user, session):
```
tiborsimko, line 975: You can squash the test commit together with the feature commit. It's nice if features are coming with tests in the same atomic unit. Helps to make sure things work already from the start.

(And while squashing, you may want to restrict the commit log message line length.)

DaanRosendal: Noted and done!

tiborsimko (approved):

PR #634, opened 2023-10-24
ci: publish docker image after new release

mdonadoni:
Closes reanahub/reana#752

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,3 +153,30 @@ jobs:
 
       - name: Build Docker image
         run: ./run-tests.sh --check-docker-build
+
+  release-docker:
+    runs-on: ubuntu-20.04
+    if: >
+      vars.RELEASE_DOCKER == 'true' &&
+      github.event_name == 'push' &&
+      startsWith(github.ref, 'refs/tags/')
+    needs:
+      - docs-sphinx
+      - lint-black
+      - lint-check-manifest
+      - lint-dockerfile
+      - lint-flake8
+      - lint-pydocstyle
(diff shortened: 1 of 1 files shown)


.github/workflows/release-docker.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  release-docker:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Release docker image
+        uses: reanahub/reana-github-actions/release-docker@v1
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          organisation: reanahub
```
tiborsimko, line 23: Two minor musings, we can fully forget them:

- Perhaps "organisation" could also be a secret so that the setup could be all secretified for a possible theoretical replication elsewhere.

- I guess we shell be using access tokens, so perhaps we could even name the PASSWORD in a ACCESS_TOKEN style. (Even though the tokens work in place of passwords, so the naming is already OK in this way; just thinking out loud about possible "discouraging" of the use of passwords here via a visible name.)

mdonadoni: I have added `vars.DOCKER_ORGANISATION` and ` vars.DOCKER_REGISTRY`. I am also using the boolean `vars.RELEASE_DOCKER` so that this job does not run on our own forks

tiborsimko (approved):
Cosmetics: let's capitalise the "name" as in "Release Docker image" for consistency with the other rules we have, and let's order the "needs" alphabetically.

PR #638, opened 2023-11-01
rest: add include command to get_workflow_status

giuseppe-steduto:
- installation: bump shared modules
- rest: add include_command parameter to get_workflow_status
  Closes reanahub/reana-workflow-controller#486.


reana_server/rest/workflows.py
```diff
@@ -341,6 +345,9 @@ def get_workflows(user, **kwargs):  # noqa
         search = request.args.get("search")
         sort = request.args.get("sort", "desc")
         status = request.args.getlist("status")
+        include_last_command = json.loads(
+            request.args.get("include_last_command", "false").lower()
+        )
```
mdonadoni, line 350: What about validating and parsing parameters with `use_args`, as done in the other endpoint?

reana_server/rest/workflows.py
```diff
@@ -349,6 +356,7 @@ def get_workflows(user, **kwargs):  # noqa
             sort=sort,
             status=status or None,
             verbose=bool(verbose),
+            include_last_command=include_last_command or verbose,
```
mdonadoni, line 359: I don't think this is needed, as it's part of `kwargs`. See for example how `include_workspace_size` is not passed explicitly. Am I right?

reana_server/rest/workflows.py
```diff
@@ -1113,9 +1126,14 @@ def get_workflow_status(workflow_id_or_name, user):  # noqa
     try:
         if not workflow_id_or_name:
             raise ValueError("workflow_id_or_name is not supplied")
+        include_last_command = json.loads(
+            request.args.get("include_last_command", "false").lower()
+        )
```
mdonadoni, line 1131: What about validating and parsing parameters with `use_args`, as done in the other endpoint?

mdonadoni:
No need to change API after performance improvements of https://github.com/reanahub/reana-db/pull/213

PR #652, opened 2023-11-30
release: 0.9.2

mdonadoni:
TODO:
- [x] Update shared modules
- [x] Update requirements.txt
- [ ] Change release date in CHANGES.rst and Dockerfile to actual release day

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -2,7 +2,7 @@
   "info": {
     "description": "Submit workflows to be run on REANA Cloud",
     "title": "REANA Server",
-    "version": "0.9.1"
+    "version": "0.9.2"
   },
   "paths": {
     "/account/settings/linkedaccounts/": {},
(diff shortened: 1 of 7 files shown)


docs/openapi.json
```diff
@@ -2,7 +2,7 @@
   "info": {
     "description": "Submit workflows to be run on REANA Cloud",
     "title": "REANA Server",
-    "version": "0.9.1"
```
tiborsimko, line 5: BTW a tiny independent change to MANIFEST.in that we could put into this release concerns:

```
include scripts/*
```

that could be better expressed as:

```
recursive-include scripts *.py
recursive-include scripts *.sh
```

For the consistency with other commands here and with what we do in other components.

See also
https://github.com/reanahub/reana-workflow-engine-snakemake/pull/72#discussion_r1412113033

tiborsimko (approved):

PR #656, opened 2023-12-12
build(docker): non-editable submodules in "latest" mode

tiborsimko:
When running development cluster in "latest" mode, install shared modules in non-editable mode. The editable mode should be reserved only for development clusters in "debug" mode with the live code reload.


mdonadoni (approved):
Looks good!

This PR needs to be rebased. The commit message also needs to be updated to switch to conventional commits.

PR #669, opened 2024-01-23
ci(commitlint): check for the presence of concrete PR number

tiborsimko:
- ci(commitlint): check for the presence of concrete PR number

    Enrich commitlint checker in order to check the precise PR number in the
    commit log headline.

- ci(shellcheck): fix exit code propagation

    Return properly the exit code status of shellcheck command when looping
    through shell script files.


mdonadoni (approved):
LGTM!

PR #674, opened 2024-02-28
build(python): update deps 0.9.3

mdonadoni:
- build(python): bump shared modules (#674)
- build(python): bump all dependencies (#674)


mdonadoni:
On second thought, should these commits be `chore` and not `build`, to avoid having them in the changelog? @tiborsimko

tiborsimko (approved):

PR #681, opened 2024-03-19
feat(gitlab): add filtering and pagination (#681)

### LINKED ISSUE #518: gitlab: add support for filtering and pagination of projects
After #509, the profile page in the web interface will show all the GitLab projects the user has access to, even those contained in groups/subgroups. This means that we will show many more projects in the UI, and that sometimes we might reach the limit of 100 projects that can be retrieved in a single request to GitLab APIs.

We should:
 - add support for pagination, so that the user can see all the projects he has access to (instead of only the first 100)
 - add support for filtering the projects, so that the user can find a given project quickly

We also need to change the interface of reana-ui accordingly.

mdonadoni:
- feat(gitlab): add search query parameter to filter projects (#681)
- feat(gitlab): add support for paginated project list (#681)
    
    BREAKING CHANGE: The REST API endpoint `gitlab_projects` now includes
    pagination details.

Closes #518

### DIFF
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2018, 2019, 2020, 2021, 2022, 2023 CERN.
+# Copyright (C) 2018, 2019, 2020, 2021, 2022, 2023, 2024 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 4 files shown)


reana_server/utils.py
```diff
@@ -509,23 +509,30 @@ def _get_gitlab_hook_id(project_id, gitlab_token):
     :param project_id: Project id on GitLab.
     :param gitlab_token: GitLab token.
     """
-    reana_hook_id = None
+    # FIXME: handle pagination of results
```
tiborsimko, line 512: Do we need the FIXME here?

tiborsimko (approved):
Works nicely :+1: 

Some cosmetic comments:

- Left a comment about a (forgotten?) FIXME in the code.

- The base commit could perhaps say "add" instead of "accept" (?), giving something like:

```
 feat(gitlab): add search query parameter to filter projects (#681)
```

reana_server/utils.py
mdonadoni: This is fixed by #684 , but I have removed it anyway

mdonadoni:
Updated the commit title, and also the BREAKING CHANGE notice to read as follows:

BREAKING CHANGE: The REST API endpoint `gitlab_projects` now includes pagination details.

PR #683, opened 2024-03-26
build(python): drop support for Python 3.6 and 3.7 (#683)

mdonadoni:
BREAKING CHANGE: drop support for Python 3.6 and 3.7

Closes reanahub/reana#784

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -139,6 +139,7 @@ setup(
         ],
     },
     include_package_data=True,
+    python_requires=">=3.8",
     extras_require=extras_require,
     install_requires=install_requires,
     setup_requires=setup_requires,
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -140,6 +140,7 @@
     },
     include_package_data=True,
     extras_require=extras_require,
+    python_requires=">=3.8",
```
mdonadoni, line 143: Like other components, what to do with `python_requires` and `classifiers`?

mdonadoni: See https://github.com/reanahub/reana-db/pull/235#discussion_r1618779094

> I have kept only Python 3.8 in `setup.py`'s classifiers and in the CI for non-shared components, and Python 3.8-3.12 for the shared ones (reana-db, reana-commons, reana-client, pytest-reana)

PR #684, opened 2024-03-28
fix(gitlab): handle pagination of GitLab webhooks (#684)

### LINKED ISSUE #682: gitlab: handle paginated results when fetching webhooks
When fetching the list of GitLab projects, the list of webhooks of each repository is also fetched to see whether REANA's webhook exists or not. However, results coming from GitLab are paginated, so it might happen that webhooks are not found if they are not part of the first page.

mdonadoni:
Depends on #681 

Closes #682

### DIFF
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -17,7 +17,7 @@ import pathlib
 import secrets
 import sys
 import shutil
-from typing import Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union, Generator
 from uuid import UUID, uuid4
 
 import click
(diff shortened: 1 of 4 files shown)


reana_server/utils.py
```diff
@@ -511,27 +524,22 @@ def _get_gitlab_hook_id(project_id, gitlab_token):
     """
     gitlab_hooks_url = (
         REANA_GITLAB_URL
-        + "/api/v4/projects/{0}/hooks?access_token={1}".format(project_id, gitlab_token)
+        + "/api/v4/projects/{0}/hooks?per_page=100&access_token={1}".format(
```
tiborsimko, line 527: Perhaps we could introduce a new configuration variable in `config.py` with `REANA_GITLAB_PER_PAGE = 100` so that the value wouldn't be hard-coded?

tiborsimko (approved):

reana_server/utils.py
mdonadoni: I have introduced `REANA_GITLAB_MAX_PER_PAGE`

PR #689, opened 2024-06-25
fix(start): validate endpoint parameters (#689)

mdonadoni:
Closes reanahub/reana-client#718

PR #691, opened 2024-07-01
fix(set_workflow_status): publish workflows to submission queue (#691)

### LINKED ISSUE #690: set_workflow_status: starting workflow should queue it
At the moment, calling `set_workflow_status` to start a workflow will result in the workflow being executed immediately, as r-server calls directly r-w-controller:

https://github.com/reanahub/reana-server/blob/3f716f7369230ab2a864b6bfcae44a952ca5bf90/reana_server/rest/workflows.py#L1510-L1519
Note that we have a second `start_workflow` endpoint which correctly queues the workflow:

https://github.com/reanahub/reana-server/blob/3f716f7369230ab2a864b6bfcae44a952ca5bf90/reana_server/rest/workflows.py#L1136

mdonadoni:
When starting a new workflow, publish the workflow to the submission
queue instead of executing the workflow immediately by calling
`set_workflow_status` in workflow-controller.

Depends on #689 
Closes #690

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -1131,6 +1131,75 @@ def get_workflow_status(workflow_id_or_name, user):  # noqa
         return jsonify({"message": str(e)}), 500
 
 
+def _start_workflow(workflow_id_or_name, user, **parameters):
+    """Start given workflow by publishing it to the submission queue.
+
+    This function is used by both the `set_workflow_status` and `start_workflow`.
+    """
+    operational_options = parameters.get("operational_options", {})
+    input_parameters = parameters.get("input_parameters", {})
+    restart = parameters.get("restart", False)
+    reana_specification = parameters.get("reana_specification")
+
+    try:
+        if not workflow_id_or_name:
+            raise ValueError("workflow_id_or_name is not supplied")
+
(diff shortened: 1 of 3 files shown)


reana_server/rest/workflows.py
```diff
@@ -1374,6 +1413,10 @@ def set_workflow_status(workflow_id_or_name, user):  # noqa
           description: Required. New workflow status.
           required: true
           type: string
+          enum:
+            - start
```
tiborsimko, line 1417: Hmm, this may look a bit inconsistent, it would be preferable to use either "start/stop/delete" (verbs) or "started/stopped/deleted" (nouns) everywhere ... but I guess this may not be for now.

(... for the API 1.0 stabilisation we could also clean mixing params/options for execution with all-runs/workspace for deletion...)

mdonadoni: Yeah, this comes from RWC, so not much we can do as of now: https://github.com/reanahub/reana-workflow-controller/blob/f1764ed7101fa9d30516b8687b7e5ef4ae8781fe/reana_workflow_controller/rest/workflows_status.py#L34-L37

tiborsimko (approved):

PR #692, opened 2024-07-02
build(docker): upgrade to ubuntu 24.04 and Python 3.12

mdonadoni:
Closes reanahub/reana#808


reana_server/factory.py
```diff
@@ -8,72 +8,4 @@
...
-    app.register_blueprint(launch.blueprint, url_prefix="/api")
-
-    @app.teardown_appcontext
-    def shutdown_session(response_or_exc):
-        """Close session on app teardown."""
-        current_app.session.remove()
-        return response_or_exc
-
-    return app
+from invenio_app.factory import create_app
```
mdonadoni, line 11: TODO: find out whether using invenio's `create_app` has unwanted side-effects, but as far as I can tell this is only used in tests

requirements.txt
```diff
@@ -1,220 +1,223 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.12
 # by the following command:
 #
-#    pip-compile --annotation-style=line --output-file=requirements.txt setup.py
+#    pip-compile --annotation-style=line --output-file=requirements.txt ./modules/reana-db/setup.py setup.py
```
mdonadoni, line 5: TODO: regenerate `requirements.txt` with the usual command after https://github.com/reanahub/reana-db/pull/234 has been merged and reana-db has been released

setup.py
```diff
@@ -59,9 +59,10 @@
     "Werkzeug<2.1",
```
mdonadoni, line 59: TODO: Flask and werkzeug (and maybe other packages) should also be updated

setup.py
```diff
@@ -59,9 +59,10 @@
...
     "flask-celeryext<0.5.0",
     "gitpython>=3.1",
-    "marshmallow>2.13.0,<=2.20.1",
+    "marshmallow>2.13.0,<3",
     "reana-commons[kubernetes,yadage,snakemake,cwl]>=0.95.0a2,<0.96.0",
-    "reana-db>=0.95.0a2,<0.96.0",
+    # commenting out so that I can test newer dependencies
+    # "reana-db>=0.95.0a2,<0.96.0",
```
mdonadoni, line 65: TODO: uncomment after https://github.com/reanahub/reana-db/pull/234 has been merged and reana-db has been released

reana_server/factory.py
mdonadoni: This change breaks the generation of the OpenAPI specification file.

Also let's check which invenio endpoints are exposed, even more so after adding `invenio-access` to the dependencies of REANA. Let's also double check that `invenio-access` is really needed

mdonadoni: REANA is technically a Flask extension (and not a Flask app), as the main Flask app is created by Invenio. This is how `ext.py`/`factory.py` worked before this PR set:

- When running in debug mode, `invenio run ...` is invoked. This calls `invenio_app.factory.create_app`.
- When running in production mode, `uwsgi` is used, and the module configured is `invenio_app.wsgi:application`. This calls `invenio_app.factory.create_app`.
- When running the tests, `reana_server.factory.create_app` is called. This a manually created Flask app with _some_ of Invenio modules loaded.
- When running `generate_openapi_spec.py`, the app is created with `reana_server.factory.create_app`.

mdonadoni: Keeping the previous create_app for tests and for the `generate_openapi_spec.py` script. Added some comments to clarify the difference between REANA's create_app and Invenio's create_app.

mdonadoni: I forgot the case of `flas reana-admin` commands. To also make this work, I have made the following changes:
- The app in `app.py` is created with `invenio_app.factory.create_app`. This is the app that Flask autodetects, so we probably want this to be the same as the production app
- The factory method `create_app` in `factory.py` has been renamed to `create_minimal_app`, and it is now used in the tests and in `generate_openapi_spec.py`

mdonadoni: I have also checked the endpoints exposed by Invenio before/after the upgrade:
- For `/api` nothing changes
- For `/oauth` nothing changes
- There are additional endpoints coming from `invenio-access`, but these are not exposed as the ingress only forwards `/api` and `/oauth` requrests to reana-server ([source](https://github.com/reanahub/reana/blob/8762855d33f859215233439496fc1934532482c4/helm/reana/templates/ingress.yaml#L41-L66))

mdonadoni:
TODO: check that everything builds and works on ARM64 (e.g. Apple Silicon)

tiborsimko (approved):

Issue #693, opened 2024-07-11
retention_rules: error when applying rule to deleted workflow


mdonadoni:
When a workflow is deleted, its workspace directory is deleted, too. This causes an error when applying retention rules, such as:
```
Applying rule <rule_id> to workflow <workflow_id>: 'path/to/delete/*' will be deleted from '/var/reana/users/<user_id>/workflows/<workflow_id>'
Error while applying rule <rule_id>: [Errno 2] No such file or directory: '/var/reana/users/<user_id>/workflows/<workflow_id>'
Setting the status of rule <rule_id> to `active`
```
We should handle this case in the script that applies retention rules, but we should also probably set the status of retention rules for deleted workflows to `inactive`.

PR #697, opened 2024-08-07
fix(reana-admin): skip retention rules with deleted workspace (#697)

mdonadoni:
Closes reanahub/reana-server#693


tiborsimko (approved):
Works nicely :+1: 

Commit news suggestion:

```
fix(reana-admin): skip retention rules for deleted workspaces (#697) 
```

PR #704, opened 2024-09-20
feat(dask): dask check resources

### LINKED ISSUE #703: feat(dask): inform users when they request too few resources
Currently, when users request resources which are fewer then one single worker blueprint, then no worker can be instantiated in the cluster and the workflow gets stuck. We should detect this before the workflow is scheduled and return an informative error message to the user.

Alputer:
closes #703

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -3,6 +3,7 @@
 The list of contributors in alphabetical order:
 
 - [Adelina Lintuluoto](https://orcid.org/0000-0002-0726-1452)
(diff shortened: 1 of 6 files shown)


Alputer:
This pull request became unnecessary after we changed the `reana.yaml` schema to ask for the `number of workers` for Dask clusters directly. If it is not provided, default value is used which is nonzero. Therefore, the case where users request too few resources so that `no worker can be created with the given amount` is no longer possible. Therefore closing the pr.

PR #706, opened 2024-10-24
chore(master): update reana-commons to 0.95.0a5

Alputer:


tiborsimko (approved):
Works nicely :+1: 

We shall need to address workflow complexity calculations. I have created a separate issue for this: https://github.com/reanahub/reana-server/issues/708

PR #738, opened 2025-07-16
fix(rest): fix non-JSON errors when opening interactive sessions

### LINKED ISSUE #738: open: print better error message when trying to open a session on non-existing workflows
### Current behaviour

When a user tries to open a Jupyter notebook session on a non-existing workflow, the error message is:

```console
$ reana-client open -w nonexistent
==> ERROR: Interactive session could not be opened:
The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.
```

This can be misleading as if there was something wrong with the cluster rather than the client arguments.

### Expected behaviour

It will be more user friendly to say that the workflow does not exist, and therefore we cannot open a notebook session on it. Similarly as we do for other commands such as `status`:

```console
$ reana-client status -w nonexistent
==> ERROR: Cannot retrieve the status of a workflow nonexistent:
REANA_WORKON is set to nonexistent, but that workflow does not exist. Please set your REANA_WORKON environment variable appropriately.
```

michaelbuchar:
Fixes https://github.com/reanahub/reana-client/issues/738

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -2631,7 +2631,17 @@ def open_interactive_session(
         return jsonify(response), http_response.status_code
     except HTTPError as e:
         logging.error(traceback.format_exc())
-        return jsonify(e.response.json()), e.response.status_code
+        # Try to parse JSON, but gracefully handle empty/non-JSON responses
+        try:
+            error_payload = e.response.json()
+            return jsonify(error_payload), e.response.status_code
+        except ValueError:
+            return (
+                jsonify(
+                    {"message": (f"Workflow '{workflow_id_or_name}' does not exist.")}
+                ),
+                404,
+            )
     except KeyError as e:
         logging.error(traceback.format_exc())
         return jsonify({"message": str(e)}), 400


reana_server/rest/workflows.py
```diff
@@ -2630,7 +2630,20 @@ def open_interactive_session(
...
-        return jsonify(e.response.json()), e.response.status_code
+        # Try to parse JSON, but gracefully handle empty/non-JSON responses
+        try:
+            error_payload = e.response.json()
+            return jsonify(error_payload), e.response.status_code
+        except ValueError:
+            return (
+                jsonify(
+                    {
+                        "message": f"REANA_WORKON is set to {workflow_id_or_name}, but that workflow does not exist. "
```
CameronMcClymont: Also the error message isn't always accurate because if your `REANA_WORKON` env variable is set to 'nothing' and you run `reana-client open -w something`, it will say `REANA_WORKON is set to something` which isn't true.

CameronMcClymont (changes requested):

CameronMcClymont (dismissed):
🦭

tiborsimko (approved):

PR #739, opened 2025-07-16
feat(validation): add workflow image vetting

### LINKED ISSUE #728: validate: check workflow specification for unathorised container images
In some deployment scenarios we may want to allow running only those workflows which use vetted authorised container images.

(This is similar to how we allow to run only certain vetted Jupyter notebook images by users, see `interactive_sessions.environments.jupyter.recommended` Helm value.)

To address these deployment scenarios, it would be useful to:

- allow REANA cluster administrators to specify which images can be used by users in their runtime workflows;
- enrich `reana-client info` to report to users which images are authorised;
- amend `reana-client validate` to report any troubles early in case a user tries to use an unauthorised image;
- make strong check on the REANA server side not to accept user workflows if the workflow uses an unauthorised image for some of its steps.

The implementation could be similar to how we handle Jupyter images, i.e. via a Helm value. However, the list of authorised images may grow, so we may have to be ready to handle potentially large number of allowed items in the vetted container image list.

CameronMcClymont:
Extend `validate_workflow` to check that all images used in workflows are authorized. Such authorized images are specified via the `vetted_container_images` Helm value. Also add details of container vetting to the `info` API endpoint.

Closes #728

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -466,6 +466,11 @@ REANA_INTERACTIVE_SESSIONS_ENVIRONMENTS_CUSTOM_ALLOWED = (
 )
 """Whether users can set custom interactive session images or not."""
 
+REANA_VETTED_CONTAINER_IMAGES = json.loads(
+    os.getenv("REANA_VETTED_CONTAINER_IMAGES", "{}")
+)
+"""Container images that users are allowed to use in their workflows."""
+
 # Kubernetes jobs timeout
 # ==================
 REANA_KUBERNETES_JOBS_TIMEOUT_LIMIT = os.getenv("REANA_KUBERNETES_JOBS_TIMEOUT_LIMIT")
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -616,6 +616,14 @@ REANA_INTERACTIVE_SESSIONS_ENVIRONMENTS_CUSTOM_ALLOWED = (
 )
 """Whether users can set custom interactive session images or not."""
 
+REANA_VETTED_CONTAINER_IMAGES = json.loads(
+    os.getenv(
+        "REANA_VETTED_CONTAINER_IMAGES",
+        '{"enabled": false, "allowlist": []}',
+    )
+)
+"""Container images that users are allowed to use in their workflows."""
+
 # Kubernetes jobs timeout
 # ==================
 REANA_KUBERNETES_JOBS_TIMEOUT_LIMIT = os.getenv("REANA_KUBERNETES_JOBS_TIMEOUT_LIMIT")
--- a/reana_server/rest/info.py
+++ b/reana_server/rest/info.py
@@ -27,6 +27,7 @@ from reana_server.config import (
     REANA_INTERACTIVE_SESSION_MAX_INACTIVITY_PERIOD,
     REANA_INTERACTIVE_SESSIONS_ENVIRONMENTS,
     REANA_INTERACTIVE_SESSIONS_ENVIRONMENTS_CUSTOM_ALLOWED,
+    REANA_VETTED_CONTAINER_IMAGES,
     DASK_ENABLED,
     DASK_AUTOSCALER_ENABLED,
     REANA_DASK_CLUSTER_DEFAULT_NUMBER_OF_WORKERS,
@@ -145,6 +146,22 @@ def info(user, **kwargs):  # noqa
                     x-nullable: true
                 type: object
               workspaces_available:
+                properties:
+                  title:
+                    type: string
+                  value:
+                    items:
+                      type: string
+                    type: array
+                type: object
+              vetted_container_images_enabled:
+                properties:
+                  title:
+                    type: string
+                  value:
+                    type: boolean
+                type: object
+              vetted_container_images_allowlist:
                 properties:
                   title:
                     type: string
@@ -318,6 +335,14 @@ def info(user, **kwargs):  # noqa
                     "title": "Whether users are allowed to spawn custom interactive session images",
                     "value": "False"
                 },
+                "vetted_container_images_enabled": {
+                    "title": "Whether container image vetting is enabled",
+                    "value": "False"
+                },
+                "vetted_container_images_allowlist": {
+                    "title": "List of container images allowed in user workflows",
+                    "value": []
+                },
                 "supported_workflow_engines": {
                     "title": "List of supported workflow engines",
                     "value": [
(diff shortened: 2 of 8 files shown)


reana_server/validation.py
```diff
@@ -119,6 +120,23 @@ def validate_inputs(reana_yaml: Dict) -> None:
             )
 
 
+def validate_images(reana_yaml: Dict) -> None:
```
michaelbuchar, line 125: I think this only scans workflow.specification.steps. Yadage uses nested stages and CWL uses requirements[].dockerPull. The client already contains extraction logic for [Yadage](https://github.com/reanahub/reana-client/blob/381a8c8fa18e1ffc2d85dd3a3795fde974d572b4/reana_client/validation/environments.py#L459-L506) and [CWL](https://github.com/reanahub/reana-client/blob/381a8c8fa18e1ffc2d85dd3a3795fde974d572b4/reana_client/validation/environments.py#L509-L528). Maybe let's move the shared extraction into reana-commons or implement equivalent server extractors?

reana_server/validation.py
```diff
@@ -119,6 +120,23 @@ def validate_inputs(reana_yaml: Dict) -> None:
...
+    if not REANA_VETTED_CONTAINER_IMAGES["enabled"]:
+        return
+
+    allowed_images = REANA_VETTED_CONTAINER_IMAGES["allowlist"]
+    steps = reana_yaml["workflow"].get("specification", {}).get("steps", [])
+
+    for step in steps:
+        image = step.get("environment", None)
+        if image and image not in allowed_images:
+            raise REANAValidationError(f"Image not allowed: {image}")
```
michaelbuchar: The server ignores falsey environments here. Snakemake jobs without an explicit image use REANA_DEFAULT_SNAKEMAKE_ENV_IMAGE [client fallback](https://github.com/reanahub/reana-client/blob/381a8c8fa18e1ffc2d85dd3a3795fde974d572b4/reana_client/validation/environments.py#L531-L547). Maybe let's also vet the resolved default image?

reana_server/config.py
```diff
@@ -564,6 +564,11 @@ def _get_rate_limit(env_variable: str, default: str) -> str:
 )
 """Whether users can set custom interactive session images or not."""
 
+REANA_VETTED_CONTAINER_IMAGES = json.loads(
+    os.getenv("REANA_VETTED_CONTAINER_IMAGES", "{}")
+)
+"""Container images that users are allowed to use in their workflows."""
```
michaelbuchar, line 625: The environment variable defaults to {} in [config.py](https://github.com/reanahub/reana-server/blob/c5e1c16053eca4a773fb6095041d79412df55548/reana_server/config.py#L567-L570), but validation and [info](https://github.com/reanahub/reana-server/blob/c5e1c16053eca4a773fb6095041d79412df55548/reana_server/rest/info.py#L595-L602) index both keys. This probably breaks standalone deployments and version skewed rollouts. Maybe let's default to {"enabled": false, "allowlist": []}?

reana_server/rest/info.py
```diff
@@ -405,6 +422,14 @@ def info(user, **kwargs):  # noqa
                     "title": "Whether users are allowed to spawn custom interactive session images",
                     "value": "False"
                 },
+                "vetted_container_images_enabled": {
+                    "title": "Whether container image vetting is enabled",
+                    "value": "False"
```
michaelbuchar: The schema declares a boolean, but the example uses "False" as a string in [reana-server](https://github.com/reanahub/reana-server/blob/c5e1c16053eca4a773fb6095041d79412df55548/reana_server/rest/info.py#L425-L428) and the generated [reana-commons snapshot](https://github.com/reanahub/reana-commons/blob/e4e40a872a0eb7e1ed00f89e0c72212d75826c5b/reana_commons/openapi_specifications/reana_server.json#L552-L559). Use JSON false and regenerate the specification.

CameronMcClymont: Thanks, fixed!

reana_server/config.py
CameronMcClymont: Oh ye nice catch, updated the default now.

reana_server/validation.py
CameronMcClymont: We don't need to vet the default image because both the list of vetted images and the default snakemake env image are set by cluster admins (i.e. `REANA_DEFAULT_SNAKEMAKE_ENV_IMAGE` is trustworthy so doesn't need to be vetted).

reana_server/validation.py
CameronMcClymont: Yes good idea, moved image extraction logic to reana-commons

reana_server/rest/info.py
```diff
@@ -583,6 +608,14 @@ def info(user, **kwargs):  # noqa
                     ]
                 ],
             ),
+            vetted_container_images_enabled=dict(
```
tiborsimko, line 611: 🤖 **`reana-client-go info` drops vetted image settings (MEDIUM)**

The Python client seems fine here, but `reana-client-go info` will not show these new settings yet. Its generated `InfoOKBody` model does not include `vetted_container_images_enabled` or `vetted_container_images_allowlist`, and the human-readable output in `cmd/info.go` is hardcoded field by field. As a result, both plain output and likely `--json` output discard these fields even though the server OpenAPI now exposes them.

Could we please update `reana-client-go` as part of this change as well? The fix should be to regenerate the Go client from the updated `reana-server/docs/openapi.json`, add the two fields to `cmd/info.go`, and update the `info` command fixtures/tests.

CameronMcClymont: Fixed in a new reana-client-go PR: https://github.com/reanahub/reana-client-go/pull/182

tiborsimko (commented):
**🤖 GitLab webhook path publishes workflows without image vetting (HIGH)**

`reana_server/rest/workflows.py:540-613` short-circuits to `publish_workflow_submission` when `git_data` is present, bypassing the start endpoint where `validate_workflow()` → `validate_images()` normally runs. With vetting enabled, an unvetted image referenced in a GitLab-tracked `reana.yaml` will still be queued and executed.

Adding `validate_images()` only at create-time isn't enough on its own: GitLab `reana.yaml` files commonly use `workflow.file:`, so the initial `yaml.safe_load()` carries no `workflow.specification` and the early check no-ops. The specification is materialised later by `_load_and_save_yadage_spec` / `load_reana_spec` at lines 600-610. The vetting check therefore has to run *after* that load.

Suggested fix — add the import and two `validate_images()` calls:

```python
from reana_server.validation import (
    validate_images,
    validate_inputs,
    validate_workflow,
    validate_workspace_path,
    validate_dask_limits,
)
```

```python
        validate_inputs(reana_spec_file)

        validate_images(reana_spec_file)   # catches inline-spec workflows

        validate_dask_limits(reana_spec_file)
```

```python
            elif workflow.type_ in ["cwl", "snakemake"]:
                reana_yaml_path = os.path.join(workflow.workspace_path, "reana.yaml")
                workflow.reana_specification = load_reana_spec(
                    reana_yaml_path, workflow.workspace_path
                )
                Session.commit()

            validate_images(workflow.reana_specification)   # closes the GitLab bypass

            parameters = request.json
            publish_workflow_submission(workflow, user.id_, parameters)
```

Worth adding a `tests/test_views.py` case that posts a `git_data` create with a disallowed image and asserts the webhook returns 400 before `publish_workflow_submission` is reached.

PR #744, opened 2025-08-01
feat(auth): implement PoC for Just-In-Time user provisioning via IdP

tomondre:
ℹ️ Contains more changes than neccessary as this PR depends on https://github.com/reanahub/reana-server/pull/741

This PR adds support for Just-In-Time (JIT) user provisioning in REANA when a user with a valid token is not yet present in the system.

If a user is not found in the REANA database based on their identity provider ID (`sub` claim), the system will:

* Use the access token to fetch user profile information from the configured `/userinfo` endpoint of the external IdP.
* Automatically create a new user record or update an existing one:
  * When a user returned by the IdP matches an existing account (same email), update that account’s idp_id.
  * When no account matches the email, create a new user and set its idp_id accordingly.

The following configuration options are introduced:

* `REANA_OAUTH_USERINFO_URL`: URL of the IdP's UserInfo endpoint.
* `REANA_OAUTH_JWK_URL`: Already present; used for JWT validation.

This makes REANA easier to integrate with external identity providers like Keycloak by removing the need for pre-registered users in the REANA database.


tiborsimko:
Closing; the move to OIDC and JWT was completed recently in reanahub/reana#977 and its related PRs.

PR #748, opened 2025-09-25
feat(rest): add quota management endpoint

### LINKED ISSUE #747: rest: add optional quota management endpoints
### Current behaviour

The CPU and Disk quota for users can be set through command-line by REANA administrators:

```
kubectl exec -i -t deployment/reana-server -c rest-api -- flask reana-admin quota-set -e john.doe@example.org -r disk -l 429496729600 --admin-access-token $REANA_ADMIN_ACCESS_TOKEN
```

This is perfectly fine, however for certain deployment scenarios, one would like to achieve this over REST API without having direct Kubernetes cluster access, so that first support line could manage user quotas without having any cluster privileges.

### Desired behaviour

It would be therefore good to introduce new REST API endpoints that would allow to consult and set the quota for users remotely.

This new REST API should be accessible only after providing certain secret which is not necessarily the same as REANA_ADMIN_ACCESS_TOKEN. This will allow to support scenarios where the frontend support line could be granted rights to manage user quotas without having access to much else.

This secret could be set in a Helm value. And, if the secret would be empty (default value), this functionality would be disabled (and so quota management would be only available from the command line, as is the case now).

The calling of the new REST API endpoints could be simply via `curl` commands. (We don't need to enrich reana-client.) The admin documentation under `docs.reana.io` should then be updated accordingly.

CameronMcClymont:
To use the endpoints, the secret `REANA_ADMIN_QUOTA_MANAGER` secret set in your Helm `values.yaml` under `components.reana_server.environment` must be included in the `X-Quota-Management-Secret` header of each request.

### `GET` user quota info

Other query params:
```
- name: user_id
  description: Get the quota limit by user ID (mutually exclusive with `email` and `user_access_token`)
- name: email
  description: Get the quota limit by user email (mutually exclusive with `user_id` and `user_access_token`)
- name: user_access_token
  description: Get the quota limit by user access token (mutually exclusive with `user_id` and `email`)
- name: resource_type
  description: The type of resource
```

### `POST` user quota limit

Body format:

```
user_id:
  type: string
  description: ID of the target user (mutually exclusive with `email`)
email:
  type: string
  description: Email of the target user (mutually exclusive with `user_id`)
resource_type:
  type: string
  description: Resource type to set
limit:
  type: integer
  description: Raw quota limit to set
```

### Examples

Fetch CPU quota info for user with email 'test@test.com':

```
GET /quota?email=test@test.com&resource_type=cpu

X-Quota-Management-Secret your_secret
```

Set disk quota limit for user with ID 'abcde':

```
POST /quota

X-Quota-Management-Secret your_secret

{
    "user_id": "abcde",
    "limit": 12345678,
    "resource_type": "disk"
}
```

Closes reanahub/reana-server#747

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1144,6 +1144,316 @@
         "summary": "Ping the server (healthcheck)"
       }
     },
+    "/api/quota": {
+      "get": {
+        "description": "This endpoint gets resource quota limits for a given user.",
+        "operationId": "get_quota_usage",
+        "parameters": [
+          {
+            "description": "REANA user quota management secret",
+            "in": "header",
+            "name": "X-Quota-Management-Secret",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "description": "Get the quota limit by user ID (mutually exclusive with `email` and `user_access_token`)",
+            "in": "query",
+            "name": "user_id",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "description": "Get the quota limit by user email (mutually exclusive with `user_id` and `user_access_token`)",
+            "in": "query",
+            "name": "email",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "description": "Get the quota limit by user access token (mutually exclusive with `user_id` and `email`)",
+            "in": "query",
+            "name": "user_access_token",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "description": "The type of resource",
+            "in": "query",
+            "name": "resource_type",
+            "required": true,
+            "type": "string"
+          }
+        ],
+        "produces": [
+          "application/json"
+        ],
+        "responses": {
+          "200": {
+            "description": "Request succeeded. Raw resource quota limit is returned.",
+            "examples": {
+              "application/json": {
+                "limit": 1000000000,
+                "message": "OK",
+                "usage": 500000000
+              }
+            },
+            "schema": {
+              "properties": {
+                "limit": {
+                  "type": "number"
+                },
+                "message": {
+                  "type": "string"
+                },
+                "usage": {
+                  "type": "number"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "400": {
+            "description": "Request failed. The incoming data specification seems malformed.",
+            "examples": {
+              "application/json": {
+                "message": "No user specified."
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "401": {
+            "description": "Request failed. Unauthorized.",
+            "examples": {
+              "application/json": {
+                "message": "Unauthorized"
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "403": {
+            "description": "Request failed. Forbidden.",
+            "examples": {
+              "application/json": {
+                "message": "Quota management endpoint is not configured."
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "404": {
+            "description": "Request failed. User not found.",
+            "examples": {
+              "application/json": {
+                "message": "User not found."
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "500": {
+            "description": "Request failed. Internal server error.",
+            "examples": {
+              "application/json": {
+                "message": "Internal server error."
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          }
+        },
+        "summary": "Get resource quota limits."
+      },
+      "post": {
+        "consumes": [
+          "application/json"
+        ],
+        "description": "This endpoint sets resource quota limits for a given user.",
+        "operationId": "set_quota_limit",
+        "parameters": [
+          {
+            "description": "REANA user quota management secret",
+            "in": "header",
+            "name": "X-Quota-Management-Secret",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "description": "Data required to set quota limits (exactly one of `user_id` or `email` must be provided).",
+            "in": "body",
+            "name": "data",
+            "required": true,
+            "schema": {
+              "properties": {
+                "email": {
+                  "description": "Email of the target user (mutually exclusive with `user_id`)",
+                  "type": "string"
+                },
+                "limit": {
+                  "description": "Raw quota limit to set",
+                  "type": "integer"
+                },
+                "resource_type": {
+                  "description": "Resource type to set",
+                  "type": "string"
+                },
+                "user_id": {
+                  "description": "ID of the target user (mutually exclusive with `email`)",
+                  "type": "string"
+                }
+              },
+              "required": [
+                "resource_type",
+                "limit"
+              ],
+              "type": "object"
+            }
+          }
+        ],
+        "produces": [
+          "application/json"
+        ],
+        "responses": {
+          "200": {
+            "description": "Resource quotas successfully set.",
+            "examples": {
+              "application/json": {
+                "limit": 1000000000,
+                "message": "OK",
+                "usage": 500000000
+              }
+            },
+            "schema": {
+              "properties": {
+                "limit": {
+                  "type": "number"
+                },
+                "message": {
+                  "type": "string"
+                },
+                "usage": {
+                  "type": "number"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "400": {
+            "description": "Request failed. The incoming data specification seems malformed.",
+            "examples": {
+              "application/json": {
+                "message": "Invalid request."
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "401": {
+            "description": "Request failed. Unauthorized.",
+            "examples": {
+              "application/json": {
+                "message": "Unauthorized"
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "403": {
+            "description": "Request failed. Forbidden.",
+            "examples": {
+              "application/json": {
+                "message": "Quota management endpoint is not configured."
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "404": {
+            "description": "Request failed. User not found.",
+            "examples": {
+              "application/json": {
+                "message": "User not found."
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "500": {
+            "description": "Request failed. Internal server error.",
+            "examples": {
+              "application/json": {
+                "message": "Quota could not be set: {error}"
+              }
+            },
+            "schema": {
+              "properties": {
+                "message": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          }
+        },
+        "summary": "Set resource quota limits."
+      }
+    },
     "/api/secrets": {
       "get": {
         "description": "Get user secrets.",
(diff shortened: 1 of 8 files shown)


CameronMcClymont:
I haven't seen these `modules/reana-db` and `modules/reana-commons` issues in the CI before. Any ideas how to fix?

CameronMcClymont:
> > I haven't seen these `modules/reana-db` and `modules/reana-commons` issues in the CI before. Any ideas how to fix?
> 
> You added these two packages as a part of your PR, I would suggest removing them so they don't interfere with the docker build and manifest.

Ah weird, not sure how those got committed. Fixed now. Doesn't it make sense for these to be in the `.gitignore` in all the REANA repos?

reana_server/rest/quota.py
```diff
@@ -0,0 +1,314 @@
...
+              message:
+                type: string
+          examples:
+            application/json:
+              {
+                "message": "Quota limit {limit} for '{resource_type} ({resource_name})' successfully set to user(s) {user_ids}."
+              }
+        500:
+          description: >-
+            Request failed. Internal server error.
```
michaelbuchar: If I provide a user_access_token but everything else is correct, I am getting a 500, I feel like I should get a 200.

Statement Krror

reana_server/rest/quota.py
```diff
@@ -0,0 +1,314 @@
...
+
+    limit, usage, user_found = _get_quota(
+        resource_type, user_id, email, user_access_token
+    )
+    if not user_found:
+        return jsonify(message="User not found."), 404
+
+    if not limit:
+        return jsonify(limit=-1, usage=usage, message="Resource limit is not set."), 200
+
```
michaelbuchar: If I set the limit to 0, this incorrectly reports “not set”. Maybe let's use is None?

reana_server/rest/quota.py
```diff
@@ -0,0 +1,314 @@
...
+) -> tuple[int | None, int | None, bool]:
+    users = _get_users(user_id, email, user_access_token) or None
+    user = users[0] if users else None
+
+    if not user:
+        return None, None, False
+
+    quota_usage = user.get_quota_usage()
+    limit: int = quota_usage.get(resource_type).get("limit", {}).get("raw")
+    usage: int = quota_usage.get(resource_type).get("usage", {}).get("raw")
```
michaelbuchar: If resource_type isn’t present, quota_usage.get(resource_type) is None and .get(...) might crash. Maybe we can use safe defaults?

reana_server/utils.py
```diff
@@ -639,6 +637,119 @@ def _validate_email(ctx, param, value):
...
+
+    if user_ids is not None and len(user_ids) == 0:
+        return "ERROR: `user_ids` must not be empty.", 500, True
+
+    if (not resource_type and not resource_name) or (resource_type and resource_name):
+        return (
+            "ERROR: Exactly one of `resource_type` or `resource_name` must be provided.",
+            500,
+            True,
+        )
```
michaelbuchar: I'd return a 400 for all of these

reana_server/rest/quota.py
```diff
@@ -0,0 +1,314 @@
...
+    resource_type = fields.Str(required=True)
+    limit = fields.Int(required=True)
+
+
+def _check_quota_management_secret() -> tuple[Response, int] | None:
+    if not REANA_ADMIN_QUOTA_MANAGER:
+        return jsonify(message="Quota functionality is not enabled"), 404
+
+    # Check if secret is provided and matches the one in the config
+    secret = request.headers.get("Authorization")
```
michaelbuchar: You’re reading admin secret from the standard Authorization header. That’s fine, but it will collide with common Bearer schemes if you later add them. Consider a custom header like X-Admin-Token (and document it), or parse Authorization: Bearer <secret>.

michaelbuchar (changes requested):

docs/openapi.json
```diff
@@ -1032,6 +1032,209 @@
...
+            "in": "query",
+            "name": "user_id",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "description": "Get the quota limit by user email (mutually exclusive with `user_id` and `user_access_token`)",
+            "in": "query",
+            "name": "email",
+            "required": false,
```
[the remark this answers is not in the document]
michaelbuchar: Sounds good to me!

reana_server/rest/quota.py
CameronMcClymont: Thanks for spotting, fixed

reana_server/rest/quota.py
```diff
@@ -0,0 +1,314 @@
...
+            ),
+            400,
+        )
+
+    limit, usage, user_found = _get_quota(
+        resource_type, user_id, email, user_access_token
+    )
+    if not user_found:
+        return jsonify(message="User not found."), 404
+
```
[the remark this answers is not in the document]
CameronMcClymont: 404 can also mean that the requested resource was not found, not just that the endpoint wasn't found. So I think we keep it as 404

reana_server/rest/quota.py
CameronMcClymont: The `user_access_token` parameter is only available with `GET` so is ignored here. And in this case the request is invalid because you are using an email for `user_id`, but I have improved the validation and error messages around this.

michaelbuchar (dismissed):

reana_server/utils.py
```diff
@@ -639,6 +637,129 @@ def _validate_email(ctx, param, value):
     return value
 
 
+def _set_quota_limit(
```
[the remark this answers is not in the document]
michaelbuchar: Let's create a new PR set for this and other EEN dependencies?

reana_server/reana_admin/cli.py
```diff
@@ -559,76 +559,17 @@ def set_quota_limit(
...
-                error_msg += f"Available resources are: {', '.join(resources)}."
-            if error_msg:
-                click.secho(
-                    error_msg,
-                    fg="red",
-                    err=True,
-                )
-                sys.exit(1)
+    msg, status_code, fatal = _set_quota_limit(
+        resource_type, resource_name, limit, emails=emails
```
tiborsimko: ### AI review

**Bug: Wrong argument order in `_set_quota_limit` call.**

The positional arguments here don't match the function signature `_set_quota_limit(limit, resource_type=None, resource_name=None, emails=None, user_ids=None)`. As written, `limit` receives the `resource_type` string, `resource_type` receives `resource_name`, and `resource_name` receives `limit`.

Suggested fix — use keyword arguments:

```python
msg, status_code, fatal = _set_quota_limit(
    limit, resource_type=resource_type, resource_name=resource_name, emails=emails
)
```

reana_server/rest/quota.py
```diff
@@ -0,0 +1,421 @@
...
+    limit = fields.Int(required=True)
+
+
+def _check_quota_management_secret() -> tuple[Response, int] | None:
+    if not REANA_QUOTA_MANAGEMENT_SECRET:
+        return jsonify(message="Quota functionality is not enabled"), 403
+
+    # Check if secret is provided and matches the one in the config
+    secret = request.headers.get("X-Quota-Management-Secret")
+    if secret != REANA_QUOTA_MANAGEMENT_SECRET:
```
tiborsimko: ### AI review

**Security: Secret comparison should be constant-time.**

Using `!=` for secret comparison is vulnerable to timing attacks. Consider using `secrets.compare_digest()` instead:

```python
import secrets

secret = request.headers.get("X-Quota-Management-Secret", "")
if not secrets.compare_digest(secret, REANA_QUOTA_MANAGEMENT_SECRET):
    return jsonify(message="Unauthorized"), 401
```

reana_server/rest/quota.py
```diff
@@ -0,0 +1,421 @@
...
+          in: header
+          description: REANA user quota management secret
+          required: true
+          type: string
+        - name: data
+          in: body
+          description: Data required to set quota limits (exactly one of `user_id` or `email` must be provided).
+          required: true
+          schema:
+            type: object
```
tiborsimko, line 283: ### AI review

**OpenAPI spec: `required` fields missing from POST body schema.**

The marshmallow `SetQuotaLimitBodySchema` marks `resource_type` and `limit` as `required=True`, but the OpenAPI schema here doesn't declare a `required` array. Swagger UI and generated clients will treat all fields as optional.

Suggested fix — add `required` to the schema:

```yaml
          schema:
            type: object
            required:
              - resource_type
              - limit
            properties:
              ...
```

(The same fix should be applied to the corresponding spec in `docs/openapi.json` and `reana-commons`.)

reana_server/utils.py
```diff
@@ -769,6 +771,129 @@ def _validate_email(ctx, param, value):
...
+        elif resource_type in ResourceType._member_names_:
+            available_resources = Resource.query.filter_by(type_=resource_type).all()
+            if available_resources and len(available_resources) > 1:
+                return (
+                    f"ERROR: There are more than one `{resource_type}` resource. Please provide resource name to specify the exact resource.",
+                    400,
+                    True,
+                )
+            else:
+                resource = available_resources[0]
```
tiborsimko, line 851: ### AI review

**Bug: `IndexError` if no `Resource` rows exist for a valid `resource_type`.**

If `resource_type` is a valid `ResourceType` member but no matching `Resource` rows exist in the DB, `available_resources` is `[]`. The `if available_resources and len(...) > 1` condition is `False`, so the `else` branch runs `available_resources[0]` which raises `IndexError`. This was pre-existing in the CLI code, but is now reachable via the new REST API where it would surface as an unhandled 500.

Suggested fix:

```python
elif resource_type in ResourceType._member_names_:
    available_resources = Resource.query.filter_by(type_=resource_type).all()
    if len(available_resources) > 1:
        return (
            f"ERROR: There are more than one `{resource_type}` resource. ...",
            400,
            True,
        )
    elif len(available_resources) == 1:
        resource = available_resources[0]
    # else: resource stays None, handled by the `if not resource:` block below
```

tiborsimko:
### AI review

**No unit tests for new REST endpoints and utility function.**

The new `reana_server/rest/quota.py` (GET and POST endpoints) and the `_set_quota_limit` utility in `reana_server/utils.py` have no corresponding tests. Codecov reports 105 lines missing coverage: `quota.py` at 22.78% patch coverage, `utils.py` at 6.97% patch coverage.

Given the amount of validation logic (mutually exclusive parameters, resource type checks, secret verification, error handling), unit tests would help catch issues like the argument-order bug noted in the inline comments.

reana_server/reana_admin/cli.py
michaelbuchar: Thanks, fixed!

reana_server/rest/quota.py
michaelbuchar: Fixed!

reana_server/rest/quota.py
```diff
@@ -0,0 +1,421 @@
...
+
+    user_id = fields.Str()
+    email = fields.Str()
+    resource_type = fields.Str(required=True)
+    limit = fields.Int(required=True)
+
+
+def _check_quota_management_secret() -> tuple[Response, int] | None:
+    if not REANA_QUOTA_MANAGEMENT_SECRET:
+        return jsonify(message="Quota functionality is not enabled"), 403
```
tiborsimko: ### AI review

**Misleading error message: "Quota functionality is not enabled".**

The REANA quota system itself is not disabled — only the new REST management endpoints are not configured (because `REANA_QUOTA_MANAGEMENT_SECRET` is empty). The current message could confuse callers into thinking quotas don't exist on this cluster.

Suggested fix — use a more precise message, e.g.:

```python
return jsonify(message="Quota management endpoint is not configured."), 403
```

The same message appears in the OpenAPI docstrings in this file (lines 178 and 353) and in `docs/openapi.json` (lines 1252 and 1405). All occurrences should be updated consistently.

michaelbuchar:
> ### AI review
> **No unit tests for new REST endpoints and utility function.**
> 
> The new `reana_server/rest/quota.py` (GET and POST endpoints) and the `_set_quota_limit` utility in `reana_server/utils.py` have no corresponding tests. Codecov reports 105 lines missing coverage: `quota.py` at 22.78% patch coverage, `utils.py` at 6.97% patch coverage.
> 
> Given the amount of validation logic (mutually exclusive parameters, resource type checks, secret verification, error handling), unit tests would help catch issues like the argument-order bug noted in the inline comments.

Integration tests performed manually

reana_server/utils.py
michaelbuchar: Fixed

reana_server/rest/quota.py
michaelbuchar: Fixed

tiborsimko:
### AI review

**Commit headline suggestion:**

```
feat(rest): add quota management endpoint (#748)
```

(Singular "endpoint" — it's one `/api/quota` route with GET and POST methods.)

reana_server/rest/quota.py
michaelbuchar: Fixed!

tiborsimko (approved):

PR #762, opened 2026-02-12
feat(config): allow automatic access token issuance

### LINKED ISSUE #925: helm: automatic issuance of access tokens
### Current status

After registration, users need to ask for command-line access token.

We have started a project where this could be automatised by means of `reana-client login` command, see https://github.com/reanahub/reana-client/issues/740 This requires to move to JWT tokens for all endpoint handling, and to eventually remove internal user database and move to Keycloak, which will take some time.

### Desired behaviour

It would be useful to introduce a new Helm value that would alter the behaviour of the `reana-server` component in such a way that when a new user logs in for the first time, the access token would be created automatically.

This is desired for some REANA deployments and would come handy before`reana-client login` is implemented.

michaelbuchar:
Closes https://github.com/reanahub/reana/issues/925

### DIFF
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -10,25 +10,22 @@
 import base64
 import csv
 import io
-import json
 import logging
 import os
 import pathlib
 import secrets
 import sys
 import shutil
-from typing import Any, Dict, List, Optional, Union, Generator
+from typing import Dict, List, Optional, Tuple, Union
 from uuid import UUID, uuid4
 
 import click
-import requests
 import yaml
 from flask import url_for
 from invenio_oauthclient.errors import OAuthClientUnAuthorized
 from jinja2 import Environment, PackageLoader, select_autoescape
 from marshmallow.exceptions import ValidationError
 from marshmallow.validate import Email
-from urllib import parse as urlparse
 
 from reana_commons.config import REANAConfig, REANA_WORKFLOW_UMASK, SHARED_VOLUME_PATH
 from reana_commons.email import send_email, REANA_EMAIL_SENDER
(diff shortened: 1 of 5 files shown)


reana_server/utils.py
```diff
@@ -439,12 +495,43 @@ def _create_and_associate_reana_user(email, fullname, username):
...
 
 
+def _auto_issue_access_token(user: User, *, requested_via: str) -> None:
+    """Issue an access token automatically when policy is auto and user lacks one."""
+    if ACCESS_TOKEN_ISSUANCE_POLICY != "auto":
+        return
+    # Do not re-issue if token exists
+    if user.access_token:
+        return
+    # Do not auto-issue for revoked users
```
tiborsimko: Currently, when users get the access token revoked, the login still works, but leads to:

Image

I wonder whether we shouldn't allow the login to proceed and to issue a new token.

Motivation: getting token revoked could happen because it leaked, but it does not mean that the user is banned. Hence a new logic could automatically issue a new token, whilst keeping the old one revoked. This is how things usually work, e.g. see GitLab where a user can have many tokens with different scopes, some can be revoked, some active... And our automated token issuance could kick in and behave in the same way, if the mode is on "auto".

BTW, the above web page screenshot is the same if the token issuance mode is on "manual", and the user has no way to click anywhere to ask for a new token. So revoking a token means to effectively "ban" a user. It would be good to distinguish the two, and offer a way to "request token" there, just as we do this for new users with the "manual" token issuance policy.

reana_server/utils.py
```diff
@@ -439,12 +495,43 @@ def _create_and_associate_reana_user(email, fullname, username):
     return user
 
 
+def _auto_issue_access_token(user: User, *, requested_via: str) -> None:
```
tiborsimko: Tested a case of a former-user-created-under-manual-policy-conditions, and the branch update works nicely 👍

tiborsimko (approved):

PR #766, opened 2026-04-20
feat(config): add security headers and restrict CORS origins

CameronMcClymont:
Add more security headers to Invenio's APP_DEFAULT_SECURE_HEADERS. Restrict CORS to REANA_HOSTNAME and disable wildcard origins. Add COOP and CORS headers in the REANA flask extension.

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -307,9 +307,13 @@ SECURITY_MSG_PASSWORD_INVALID_LENGTH = failed_signin_msg
 # CORS
 # ====
 REST_ENABLE_CORS = True
-# change this only while developing
-CORS_SEND_WILDCARD = True
+# See: https://flask-cors.readthedocs.io/en/latest/configuration.html
+# Echo the request's origin back only if it matches CORS_ORIGINS.
+# If True, it would send Access-Control-Allow-Origin: *
+# unconditionally, and let any website read API responses in the browser
+CORS_SEND_WILDCARD = False
 CORS_SUPPORTS_CREDENTIALS = False
+CORS_ORIGINS = [REANA_URL]
 
 # Flask configuration
 # ===================
@@ -348,7 +352,26 @@ if REANA_HOSTNAME:
 # ======================
 PROXYFIX_CONFIG = json.loads(os.getenv("PROXYFIX_CONFIG", '{"x_proto": 1}'))
 
-APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {}
+# See:
+# - Invenio docs: https://invenio-app.readthedocs.io/en/latest/configuration.html
+# - flask-talisman docs: https://github.com/GoogleCloudPlatform/flask-talisman
+APP_DEFAULT_SECURE_HEADERS["frame_options"] = "DENY"
+APP_DEFAULT_SECURE_HEADERS["strict_transport_security"] = True
+APP_DEFAULT_SECURE_HEADERS["strict_transport_security_max_age"] = 31536000
+APP_DEFAULT_SECURE_HEADERS["strict_transport_security_include_subdomains"] = True
+APP_DEFAULT_SECURE_HEADERS["referrer_policy"] = "strict-origin-when-cross-origin"
+# NOTE: keep in sync with reana-ui nginx/reana-ui.conf content security policy headers
+APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {
+    "default-src": "'self'",
+    "script-src": "'self'",
+    "style-src": ["'self'", "'unsafe-inline'"],
+    "img-src": ["'self'", "data:"],
+    "font-src": "'self'",
+    "connect-src": "'self'",
+    "frame-ancestors": "'none'",
+    "object-src": "'none'",
+    "base-uri": "'self'",
+}
 APP_DEFAULT_SECURE_HEADERS.update(
     json.loads(os.getenv("APP_DEFAULT_SECURE_HEADERS", "{}"))
 )
--- a/reana_server/ext.py
+++ b/reana_server/ext.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2019, 2020, 2021, 2022, 2024 CERN.
+# Copyright (C) 2019, 2020, 2021, 2022, 2024, 2026 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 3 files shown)


reana_server/config.py
```diff
@@ -311,7 +315,24 @@ def _(x):
...
+APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {
+    "default-src": "'self'",
+    "script-src": "'self'",
+    "style-src": ["'self'", "'unsafe-inline'"],
+    "img-src": ["'self'", "data:"],
+    "font-src": "'self'",
+    "connect-src": "'self'",
+    "frame-ancestors": "'none'",
+    "object-src": "'none'",
+    "base-uri": "'self'",
```
tiborsimko, line 373: ### AI review

**Consider adding `Referrer-Policy` and `Permissions-Policy` headers.**

`Referrer-Policy` prevents leaking URL paths in the `Referer` header to external sites. `Permissions-Policy` disables browser features REANA doesn't need (camera, microphone, geolocation).

Suggested additions:

```python
APP_DEFAULT_SECURE_HEADERS["referrer_policy"] = "strict-origin-when-cross-origin"
APP_DEFAULT_SECURE_HEADERS["permissions_policy"] = "camera=(), microphone=(), geolocation=()"
```

Also, please verify that `X-Content-Type-Options: nosniff` is already set by Invenio/flask-talisman by default — the `reana-ui` nginx config adds it explicitly, but this PR doesn't set it.

reana_server/config.py
```diff
@@ -311,7 +315,24 @@ def _(x):
 # ======================
 PROXYFIX_CONFIG = json.loads(os.getenv("PROXYFIX_CONFIG", '{"x_proto": 1}'))
```
tiborsimko, line 354: ### AI review

**Note: This CSP policy is duplicated in `reana-ui` nginx config.**

The same Content-Security-Policy is defined in `reana-ui`'s `nginx/reana-ui.conf` (PR reanahub/reana-ui#495). If either is updated, the other must be kept in sync. Consider adding a comment here pointing to the counterpart, e.g.:

```python
# NOTE: keep in sync with reana-ui nginx/reana-ui.conf security headers
```

reana_server/ext.py
```diff
@@ -98,6 +98,12 @@ def init_app(self, app):
         account_info_received.connect(_create_and_associate_oauth_user)
         user_registered.connect(_create_and_associate_local_user)
 
+        @app.after_request
+        def add_cross_origin_headers(response):
+            response.headers["Cross-Origin-Opener-Policy"] = "same-origin"
+            response.headers["Cross-Origin-Resource-Policy"] = "same-origin"
+            return response
```
tiborsimko, line 151: ### AI review

**Consider adding `Cache-Control: no-store` to API responses.**

REANA API responses can contain sensitive data (user emails, access tokens, workflow details, quota information). Without explicit cache control, intermediate proxies or browser caches could store and leak this data between users.

Suggested addition to this `after_request` handler:

```python
@app.after_request
def add_cross_origin_headers(response):
    response.headers["Cross-Origin-Opener-Policy"] = "same-origin"
    response.headers["Cross-Origin-Resource-Policy"] = "same-origin"
    response.headers["Cache-Control"] = "no-store, no-cache"
    return response
```

This could also go into a follow-up hardening PR if preferred.

tiborsimko:
### AI review

**Commit message suggestion:**

```
feat(config): restrict CORS and add HTTP security headers (#766)

Restrict CORS origins to the REANA hostname instead of
using a wildcard, and add HTTP response security headers
including Content-Security-Policy, HSTS, X-Frame-Options,
Cross-Origin-Opener-Policy, and Cross-Origin-Resource-Policy.
This hardens the server against clickjacking, MIME sniffing,
and cross-origin data leaks.
```

reana_server/ext.py
CameronMcClymont: Good suggestion, addressed and pushing soon

reana_server/config.py
CameronMcClymont: Comments added and pushing soon

reana_server/config.py
CameronMcClymont: For the content type options, Talisman sets it by default. Referrer-Policy and Permissions-Policy added here and in reana-ui.

reana_server/config.py
```diff
@@ -280,9 +280,13 @@ def _(x):
...
 REST_ENABLE_CORS = True
-# change this only while developing
-CORS_SEND_WILDCARD = True
+# See: https://flask-cors.readthedocs.io/en/latest/configuration.html
+# Echo the request's origin back only if it matches CORS_ORIGINS.
+# If True, it would send Access-Control-Allow-Origin: *
+# unconditionally, and let any website read API responses in the browser
+CORS_SEND_WILDCARD = False
 CORS_SUPPORTS_CREDENTIALS = False
+CORS_ORIGINS = [f"https://{REANA_HOSTNAME}"]
```
michaelbuchar: ### AI-aided review

CORS_ORIGINS is computed as https://{REANA_HOSTNAME}, but the server’s own canonical public URL is derived from both hostname and port via [REANA_URL](https://github.com/reanahub/reana-server/blob/c8b740348ef3555bac0a8215f784057a89ea1103/reana_server/config.py#L47). On the default config, that public origin is https://localhost:30443, not https://localhost. With this PR, browser requests from any non-443 deployment will stop matching the allowlist and lose Access-Control-Allow-Origin. I would derive CORS_ORIGINS from REANA_URL or compose_reana_url(REANA_HOSTNAME, REANA_HOSTPORT) instead of rebuilding it from hostname only.

reana_server/ext.py
```diff
@@ -98,6 +98,18 @@ def init_app(self, app):
         account_info_received.connect(_create_and_associate_oauth_user)
         user_registered.connect(_create_and_associate_local_user)
 
+        @app.after_request
```
michaelbuchar, line 141: ### AI-aided review

The new global after_request header hook is untested, and Codecov is already flagging the added ext.py lines as uncovered. Given the port-sensitive CORS change above and the unconditional header mutation, this should have at least one request-level test that sends an Origin header and asserts the returned Access-Control-Allow-Origin, plus one asserting the added security headers on a normal API response. Maybe we can add a quick test for this?

reana_server/config.py
CameronMcClymont: Ah nice catch, changed to `[REANA_URL]` now.

michaelbuchar (dismissed):
Looks good to me!

tiborsimko (approved):

PR #767, opened 2026-04-21
fix(dask): ensure workflows have a Dask service row

### LINKED ISSUE #493: service-logs: Dask workflow shows no service logs and empty component selector
### Current behaviour
When running a Dask workflow on a REANA deployment with Dask enabled, the workflow finishes successfully, but the Service logs tab is empty. The page also currently renders a component dropdown even though there are no components to choose from.

Image

### Expected behaviour

For a Dask workflow, one of these should happen:

1. the Service logs tab should show the available Dask service logs or
2. if there are genuinely no service log components to show, the Component selector should be hidden/removed and the empty state should explain that no service components were created for this workflow.

### How to reproduce:

I used a local dev REANA deployment with Dask enabled and the Dask operator installed.

I ran the Dask demo workflow from:

```console
/src/reana-demo-dask-coffea
````

using:

```console
source reana-venv/bin/activate
reana-client run -w dask-test
```

The workflow completed successfully.

michaelbuchar:
Restarted Dask workflows can be created without a linked dask-service-* database row, which leaves the operator with nowhere to attach service logs. Fixed this by creating the Dask service row when cloning restarted workflows and by ensuring it exists again just before workflow submission.

Closes reanahub/reana-ui#493

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -40,6 +40,7 @@ from reana_server.utils import (
     _get_reana_yaml_from_gitlab,
     _load_and_save_yadage_spec,
     clone_workflow,
+    ensure_dask_service,
     get_quota_excess_message,
     get_workspace_retention_rules,
     is_uuid_v4,
@@ -1220,6 +1221,10 @@ def _start_workflow(workflow_id_or_name, user, **parameters):
         validate_workflow(
             workflow.reana_specification, input_parameters=input_parameters
         )
+
+        # Backfill the Dask service row for legacy workflows created before this fix
+        if ensure_dask_service(workflow):
+            Session.object_session(workflow).commit()
 
         # when starting the workflow, the scheduler will call RWC's `set_workflow_status`
         # with the given `parameters`
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -35,13 +35,16 @@ from reana_commons.errors import (
     REANAValidationError,
     REANAEmailNotificationError,
 )
-from reana_commons.utils import get_quota_resource_usage
+from reana_commons.utils import get_dask_component_name, get_quota_resource_usage
 from reana_commons.yadage import yadage_load_from_workspace
 from reana_db.database import Session
 from reana_db.models import (
     ResourceType,
     ResourceUnit,
     RunStatus,
+    Service,
+    ServiceStatus,
+    ServiceType,
     User,
     UserResource,
     UserToken,
@@ -68,6 +71,7 @@ from reana_server.complexity import (
 from reana_server.config import (
     ADMIN_USER_ID,
     REANA_HOSTNAME,
+    REANA_URL,
     REANA_SSO_EOSC_REQUIRED_ENTITLEMENT,
     REANA_USER_EMAIL_CONFIRMATION,
     REANA_WORKFLOW_SCHEDULING_POLICY,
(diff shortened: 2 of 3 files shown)


reana_server/utils.py
```diff
@@ -698,6 +701,36 @@ def get_workspace_retention_rules(
...
+    )
+
+
+def ensure_dask_service(workflow: Workflow) -> bool:
+    """Ensure that a Dask workflow has a matching service DB row."""
+    if not workflow_uses_dask(workflow.reana_specification):
+        return False
+
+    service_name = get_dask_component_name(workflow.id_, "database_model_service")
+    if workflow.services.filter_by(name=service_name).one_or_none():
```
tiborsimko: This bit needs update as per recent DB changes.

```diff
diff --git a/reana_server/utils.py b/reana_server/utils.py
index f56a8e2..ce56b70 100644
--- a/reana_server/utils.py
+++ b/reana_server/utils.py
@@ -730,7 +730,7 @@ def ensure_dask_service(workflow: Workflow) -> bool:
         return False

     service_name = get_dask_component_name(workflow.id_, "database_model_service")
-    if workflow.services.filter_by(name=service_name).one_or_none():
+    if any(s.name == service_name for s in workflow.services):
         return False

     workflow.services.append(build_dask_service(workflow))
```

Before this fix, I was getting:

```console
$ reana-client start -w dask-coffea-serial-kubernetes

  ==> ERROR: Cannot start workflow dask-coffea-serial-kubernetes:
  'InstrumentedList' object has no attribute 'filter_by'
```

After this fix, everything is OK.

reana_server/utils.py
```diff
@@ -704,6 +701,36 @@ def get_workspace_retention_rules(
...
+    return bool(
+        reana_specification.get("workflow", {}).get("resources", {}).get("dask")
+    )
+
+
+def build_dask_service(workflow: Workflow) -> Service:
+    """Build the Dask service DB row for a workflow."""
+    return Service(
+        name=get_dask_component_name(workflow.id_, "database_model_service"),
+        uri=f"https://{REANA_HOSTNAME}/{workflow.id_}/dashboard/status",
```
tiborsimko: 🤖 **Inconsistent URI between Dask service creation paths**

Here the URI is built as `f"https://{REANA_HOSTNAME}/..."`, but RWC's existing creation path at `reana_workflow_controller/rest/workflows.py:652` uses `f"{REANA_URL}/..."` where `REANA_URL = compose_reana_url(REANA_HOSTNAME, REANA_HOSTPORT)` (omits the port only for 443). For non-default ports the two paths produce different `service.uri` values for otherwise-identical workflows.

Suggested fix — share the URL composition:

```diff
-        uri=f"https://{REANA_HOSTNAME}/{workflow.id_}/dashboard/status",
+        uri=f"{REANA_URL}/{workflow.id_}/dashboard/status",
```

This needs `REANA_URL` to exist in `reana_server/config.py` analogous to RWC's helper, or `compose_reana_url` to be moved into `reana-commons` for sharing.

reana_server/rest/workflows.py
```diff
@@ -1221,6 +1217,9 @@ def _start_workflow(workflow_id_or_name, user, **parameters):
             workflow.reana_specification, input_parameters=input_parameters
         )
 
+        if ensure_dask_service(workflow):
```
tiborsimko, line 1226: 🤖 **Safety net likely redundant + minor commit-ordering note**

With `clone_workflow` in this PR now creating the Service row for restarted workflows, and RWC's create endpoint already creating it for new workflows, both creation paths initialise the row reliably. The only thing this safety net catches is legacy workflows created before this fix is deployed.

Side note on ordering: the commit happens *before* `publish_workflow_submission`. If publish raises, the Service+WorkflowService rows persist for a workflow that never queues — harmless on retry (the workflow stays `created`; `ensure_dask_service` finds the existing row and skips), but worth being aware of.

Two options:

1. Drop the safety net for simplicity:

   ```diff
   -        if ensure_dask_service(workflow):
   -            Session.object_session(workflow).commit()
   -
           # when starting the workflow, the scheduler will call RWC's `set_workflow_status`
   ```

2. Keep it, but add a one-line comment explaining the legacy-compat reason so future readers don't wonder.

reana_server/utils.py

reana_server/rest/workflows.py
michaelbuchar: Fixed, added a comment

reana_server/utils.py
tiborsimko: Thanks for the revisions! The URI now matches RWC for the default deployment, and keeping the `_start_workflow` safety net
  with the legacy-compat comment is exactly the right call.

  One small observation: server's `compose_reana_url` does `if str(hostport) == "443":`, while RWC's does `if hostport ==
  443:`. Since `os.getenv` returns `str`, RWC's check never actually fires — for a port-443 deployment RWC would emit
  `https://host:443` while server (correctly) emits `https://host`, so the two paths still disagree in that edge case. The bug
  is on the RWC side, not here. 

Would it be worth a follow-up PR (A) to fix this in RWC? Or perhaps (B) to lift `compose_reana_url` into `reana-commons` and fix RWC's
   type comparison in one go?

michaelbuchar: I would go with A as a small follow-up PR in reana-workflow-controller. The bug is on the r-w-c side and since compose_reana_url currently only exists in two places moving it to reana-commons feels like more work than necessary for this fix.

michaelbuchar: I created a new PR: https://github.com/reanahub/reana-workflow-controller/pull/680

tiborsimko (approved):

PR #768, opened 2026-04-22
feat(rest): add quota period endpoints 

michaelbuchar:
Closes https://github.com/reanahub/reana-server/issues/774

### DIFF
--- a/reana_server/reana_admin/cli.py
+++ b/reana_server/reana_admin/cli.py
@@ -19,6 +19,7 @@ from typing import List, Optional
 import click
 import requests
 import tablib
+from click.core import ParameterSource
 from flask.cli import with_appcontext
 from invenio_accounts.utils import register_user
 from kubernetes.client.rest import ApiException
@@ -62,7 +63,9 @@ from reana_server.utils import (
     _export_users,
     _get_user_by_criteria,
     _get_users,
+    _UNSET,
     _import_users,
+    _set_quota_period,
     _validate_email,
     _validate_password,
     create_user_workspace,
@@ -74,6 +77,12 @@ from reana_server.utils import (
 @click.group()
 def reana_admin():
     """REANA administration commands."""
+
+
+def _unset_if_option_omitted(ctx, param, value):
+    if ctx.get_parameter_source(param.name) == ParameterSource.DEFAULT:
+        return _UNSET
+    return value
 
 
 @reana_admin.command("create-admin-user")
@@ -567,6 +576,71 @@ def set_quota_limit(
     )
     click.secho(
         msg,
+        fg="green" if status_code == 200 else "red",
+        err=False if status_code == 200 else True,
+    )
+
+    if fatal:
+        sys.exit(1)
+
+
+@reana_admin.command(
+    "quota-set-period",
+    help="Set periodic quota fields for a given user.",
+)
+@click.option("--id", "user_id", help="The id of the user.")
+@click.option("-e", "--email", help="The email of the user.")
+@click.option(
+    "--resource",
+    "resource_type",
+    required=True,
+    type=click.Choice(["cpu"]),
+    help="Specify quota resource. Only cpu is currently supported.",
+)
+@click.option(
+    "--quota-period-months",
+    type=int,
+    default=None,
+    callback=_unset_if_option_omitted,
+    help="Length of quota period in months.",
+)
+@click.option(
+    "--quota-period-start-at",
+    type=click.DateTime(formats=["%Y-%m-%d", "%Y-%m-%dT%H:%M:%S"]),
+    default=None,
+    callback=_unset_if_option_omitted,
+    help="Current active quota period start datetime.",
+)
+@admin_access_token_option
+@click.pass_context
+def set_quota_period(
+    ctx,
+    user_id,
+    email,
+    resource_type,
+    quota_period_months,
+    quota_period_start_at,
+    admin_access_token,
+):
+    """Set periodic quota fields for one user."""
+    if int(bool(user_id)) + int(bool(email)) != 1:
+        click.secho(
+            "ERROR: Exactly one of `user_id` or `email` must be provided.",
+            fg="red",
+            err=True,
+        )
+        sys.exit(1)
+
+    msg, status_code, fatal = _set_quota_period(
+        resource_type=resource_type,
+        quota_period_months=quota_period_months,
+        quota_period_start_at=quota_period_start_at,
+        user_id=user_id,
+        email=email,
+    )
+
+    click.secho(
+        msg or "Periodic quota fields updated successfully.",
         fg="green" if status_code == 200 else "red",
         err=False if status_code == 200 else True,
     )
--- a/reana_server/rest/quota.py
+++ b/reana_server/rest/quota.py
@@ -12,11 +12,18 @@ import secrets
 from typing import Optional
 
 from flask import Blueprint, jsonify, request, Response
-from marshmallow import fields, Schema
-from reana_db.models import ResourceType
+from marshmallow import fields, Schema, ValidationError
+from reana_db.database import Session
+from reana_db.models import ResourceType, UserResource
+from reana_db.utils import get_default_quota_resource
 
 from reana_server.config import REANA_QUOTA_MANAGEMENT_SECRET
-from reana_server.utils import _set_quota_limit, _get_users
+from reana_server.utils import (
+    _get_users,
+    _set_quota_limit,
+    _set_quota_period,
+    serialize_utc_datetime,
+)
 
 blueprint = Blueprint("quota", __name__)
 
@@ -30,6 +37,16 @@ class SetQuotaLimitBodySchema(Schema):
     limit = fields.Int(required=True)
 
 
+class PatchQuotaBodySchema(Schema):
+    """Schema for patch_quota endpoint body."""
+
+    user_id = fields.Str()
+    email = fields.Str()
+    resource_type = fields.Str(required=True)
+    quota_period_months = fields.Int(allow_none=True, strict=True)
+    quota_period_start_at = fields.DateTime(allow_none=True)
+
+
 def _check_quota_management_secret() -> tuple[Response, int] | None:
     if not REANA_QUOTA_MANAGEMENT_SECRET:
         return jsonify(message="Quota management endpoint is not configured."), 403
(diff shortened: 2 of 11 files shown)


michaelbuchar:
Two checks are failing since reana-db changes from this PR: https://github.com/reanahub/reana-db/pull/255 need to be merged in first.

tiborsimko (commented):
Inline review of the quota period endpoints. Punch list grouped by severity (BLOCKER / HIGH / MEDIUM / LOW / NIT). Happy to discuss any of these.

reana_server/reana_admin/cli.py
```diff
@@ -575,6 +584,118 @@ def set_quota_limit(
...
+    help="Set periodic quota fields for a given user.",
+)
+@click.option("--id", "user_id", help="The id of the user.")
+@click.option("-e", "--email", help="The email of the user.")
+@click.option(
+    "--resource",
+    "resource_type",
+    required=True,
+    type=click.Choice(["cpu", "disk"]),
+    help="Specify quota resource. Currently only cpu is supported.",
```
tiborsimko: 🤖 **`quota-set-period` advertises `disk` but rejects it at runtime (BLOCKER)**

The `--resource` option is declared with `click.Choice(["cpu", "disk"])` while the help text says "Currently only cpu is supported." If a user follows the choices and passes `--resource disk`, Click happily accepts it, the call reaches `_set_quota_period`, and only then does it return *"Periodic quota is currently supported only for CPU."* That's a CLI UX trap and an inconsistent contract.

Suggested fix — restrict the choice to what is actually supported and update the help text accordingly:

```python
@click.option(
    "--resource",
    "resource_type",
    required=True,
    type=click.Choice(["cpu"]),
    help="Specify quota resource. Only `cpu` is currently supported.",
)
```

When disk support lands, widen the `click.Choice` and lift the runtime restriction in `_set_quota_period` together.

reana_server/rest/quota.py
```diff
@@ -422,4 +492,128 @@ def set_quota_limit():  # noqa
...
+          description: Invalid request.
+        401:
+          description: Unauthorized.
+        403:
+          description: Quota functionality is not enabled.
+        404:
+          description: User not found.
+        500:
+          description: Internal server error.
+    """
```
tiborsimko, line 571: 🤖 **`PATCH /api/quota` has no REST-level test coverage (HIGH)**

The new endpoint has non-trivial logic that doesn't appear in any test:

- `X-Quota-Management-Secret` header check (`_check_quota_management_secret`)
- Marshmallow body validation via `PatchQuotaBodySchema`
- The mutual-exclusion check on `user_id` / `email` (line 565)
- The "at least one of `quota_period_months` or `quota_period_start_at`" check (line 579)
- Wiring the response from `_set_quota_period` + `_get_quota` + `_get_quota_period`

The new tests in `tests/test_utils.py` cover the helper `_set_quota_period`, and `tests/test_cli.py` covers one CLI rejection path, but the endpoint itself is exercised nowhere.

Suggested fix — add `tests/test_quota.py` (or extend an existing test file) with at least:

- happy path PATCH returning the updated quota payload (200)
- missing/invalid `X-Quota-Management-Secret` (403)
- body that violates `int(bool(user_id)) + int(bool(email)) != 1` (400)
- empty period payload `{user_id, resource_type}` with neither period field (400)
- non-existent user (404)

reana_server/rest/quota.py
```diff
@@ -422,4 +492,128 @@ def set_quota_limit():  # noqa
...
+def patch_quota():  # noqa
+    r"""Endpoint to patch quota period fields.
+
+    ---
+    patch:
+      summary: Patch periodic quota fields.
+      description: >-
+        This endpoint sets periodic quota accounting fields for a given user.
+      operationId: patch_quota
+      produces:
```
tiborsimko, line 522: 🤖 **PATCH endpoint OpenAPI spec is missing `consumes: application/json` (MEDIUM)**

Sibling endpoints in this file (`POST /quota` at line 338-339, `GET /quota` similar) declare `consumes: application/json` on the operation. The new PATCH operation only declares `produces`. Generated OpenAPI clients and our `docs/openapi.json` will under-specify the request content-type as a result.

Suggested fix — add `consumes` between `operationId` and `produces`:

```yaml
      operationId: patch_quota
      consumes:
        - application/json
      produces:
        - application/json
```

reana_server/rest/quota.py
```diff
@@ -422,4 +492,128 @@ def set_quota_limit():  # noqa
...
+                type: string
+              email:
+                type: string
+              resource_type:
+                type: string
+              quota_period_months:
+                type: integer
+              quota_period_start_at:
+                type: string
+                format: date-time
```
tiborsimko, line 555: 🤖 **PATCH body schema lacks property descriptions and a `required` array (MEDIUM)**

The sibling `POST /quota` body schema (around line 350-380 of this file) gives every property a `description` and an explicit `required: ["resource_type", ...]` list. The new PATCH body schema declares only types — leaving generated clients and the rendered API docs unable to tell which fields are required and what they mean. (`PatchQuotaBodySchema` *does* mark `resource_type` as required at the marshmallow level, so the spec is also out of sync with runtime behaviour.)

Suggested fix — mirror the POST endpoint's documentation style:

```yaml
          schema:
            type: object
            required:
              - resource_type
            properties:
              user_id:
                type: string
                description: ID of the target user (mutually exclusive with `email`).
              email:
                type: string
                description: Email of the target user (mutually exclusive with `user_id`).
              resource_type:
                type: string
                description: Resource whose periodic quota is being updated. Currently only `cpu` is supported.
              quota_period_months:
                type: integer
                minimum: 1
                description: Length of the accounting window in months. Pass `null` to disable periodic accounting.
              quota_period_start_at:
                type: string
                format: date-time
                description: Timestamp of the active window start. Requires `quota_period_months` to be set.
```

reana_server/utils.py
```diff
@@ -940,6 +944,80 @@ def _set_quota_limit(
...
+    if quota_period_start_at is _UNSET and not user_resource.quota_period_start_at:
+        quota_period_start_at = get_current_quota_period_start_at_from_anchor(
+            anchor_at=user.created,
+            quota_period_months=effective_quota_period_months,
+        )
+
+    if quota_period_start_at is not _UNSET:
+        user_resource.quota_period_start_at = quota_period_start_at
+
+    Session.commit()
```
tiborsimko: 🤖 **`_set_quota_period` doesn't wrap `Session.commit()` in try/except (LOW)**

The sibling helper `_set_quota_limit` (line 818) wraps the body in `try/except Exception`, returns a `(message, 500, False)` tuple on failure, and logs the traceback. `_set_quota_period` does not — if the commit fails (constraint violation, lost DB connection, integrity error), the exception will propagate out of the helper and the endpoint will return Flask's generic 500 HTML page instead of a JSON error envelope. Inconsistent with the rest of the file.

Suggested fix — adopt the same pattern as `_set_quota_limit`:

```python
try:
    ...
    Session.commit()
    return None, 200, False
except Exception as e:
    logging.debug(traceback.format_exc())
    logging.debug(str(e))
    return f"Error setting quota period: \n{e}", 500, False
```

reana_server/utils.py
```diff
@@ -940,6 +944,80 @@ def _set_quota_limit(
...
+    if quota_period_start_at is not _UNSET and effective_quota_period_months is None:
+        return (
+            "Cannot set `quota_period_start_at` for a user without periodic "
+            "quota cadence. Set `quota_period_months` first.",
+            400,
+            True,
+        )
+
+    if quota_period_months is not _UNSET:
+        user_resource.quota_period_months = quota_period_months
```
tiborsimko: 🤖 **Disabling periodic quotas leaves a stale period start (MEDIUM)**

When a caller sets `quota_period_months` to `None`, the cadence is disabled but `quota_period_start_at` is left unchanged unless the caller also sends `quota_period_start_at: null`. This leaves stale period metadata in the database. If the user is later re-enabled with only `quota_period_months`, the old start can be reused, which is surprising for an operation that looked like it disabled periodic accounting.

Suggested fix — clear the start automatically when disabling the cadence, or make the API require both fields for disable operations and document that contract. The simpler behaviour would be:

```python
if quota_period_months is not _UNSET:
    user_resource.quota_period_months = quota_period_months
    if quota_period_months is None:
        user_resource.quota_period_start_at = None
```

Please add a test that `quota_period_months: null` clears both periodic fields.

reana_server/rest/quota.py
```diff
@@ -422,4 +492,128 @@ def set_quota_limit():  # noqa
...
+            type: object
+            properties:
+              user_id:
+                type: string
+              email:
+                type: string
+              resource_type:
+                type: string
+              quota_period_months:
+                type: integer
```
tiborsimko, line 549: 🤖 **Nullable period fields are not documented as nullable (MEDIUM)**

`PatchQuotaBodySchema` accepts `null` for both `quota_period_months` and `quota_period_start_at`, and the administration docs describe using `null` to disable periodic accounting. The OpenAPI docstring only declares these as `integer` and `string`, so the generated `docs/openapi.json` will not tell clients that `null` is valid.

Suggested fix — mark both PATCH body fields as nullable using the Swagger 2 convention already used elsewhere in the generated spec:

```yaml
              quota_period_months:
                type: integer
                x-nullable: true
              quota_period_start_at:
                type: string
                format: date-time
                x-nullable: true
```

Please mirror the same nullable metadata in `reana-commons` so the shared OpenAPI specification stays aligned.

reana_server/rest/quota.py
michaelbuchar: Integration tests have been done locally; if you agree, I would move this to a separate PR, what do you think?

reana_server/reana_admin/cli.py
michaelbuchar: Fixed

reana_server/rest/quota.py
michaelbuchar: Fixed

reana_server/rest/quota.py
michaelbuchar: Fixed

reana_server/utils.py

reana_server/utils.py

reana_server/rest/quota.py
michaelbuchar: Fixed

reana_server/rest/quota.py
```diff
@@ -13,14 +13,21 @@
...
+from reana_db.utils import get_default_quota_resource
 
 from reana_server.config import REANA_QUOTA_MANAGEMENT_SECRET
-from reana_server.utils import _set_quota_limit, _get_users
+from reana_server.utils import _set_quota_limit, _set_quota_period, _get_users
 
 blueprint = Blueprint("quota", __name__)
 
 
+def _serialize_utc_datetime(dt):
```
tiborsimko: 🤖 **`_serialize_utc_datetime` is duplicated byte-for-byte in `rest/users.py` (LOW)**

The same three-line helper now lives in both `reana_server/rest/quota.py:26-28` and `reana_server/rest/users.py:33-35`. Both files already import from `reana_server.utils`, which is the natural home for a shared serialisation helper — `_serialize_user_quota` itself just moved there in spirit. Keeping two copies means the next correctness fix (e.g. handling timezone-aware datetimes, or switching to a stricter RFC 3339 with microseconds normalisation) needs to land in two places.

Suggested fix — move the helper to `reana_server/utils.py` once and import from both REST modules:

```python
# reana_server/utils.py
def serialize_utc_datetime(dt):
    """Serialize a naive UTC datetime using an RFC 3339 UTC designator."""
    return dt.isoformat() + "Z" if dt else None
```

```python
# reana_server/rest/quota.py and reana_server/rest/users.py
from reana_server.utils import serialize_utc_datetime
```

Dropping the leading underscore is optional but consistent with the public-ish naming the rest of `reana_server/utils.py` uses (e.g. `_set_quota_limit` is private because it returns a tuple-status convention, while a pure formatting helper doesn't need that visibility hint).

tiborsimko (commented):
Inline review of the latest update. One nit on the new datetime helper.

reana_server/rest/quota.py
michaelbuchar: Fixed, thanks!

tiborsimko (approved):

PR #769, opened 2026-04-22
fix(rest): sync workspace listing error docs

### LINKED ISSUE #644: workspace: ensure smooth performance of file listings for workspaces containing 250k files
## Current behaviour

When a workspace contains many produced output files, the `reana-workflow-controller` pod starts to use too much CPU and too much memory. It could happen that the whole system "freezes". This was observed for a workflow that was producing over 250k files, and when the researcher clicked on the "Workspace" tab of the REANA web interface.

(When this happened, the REANA web interface was slow to serve other users as well. The solution was to delete the workflow controller pod that was "stuck" in calculating the workspace listing response.)

## Expected behaviour

We should improve the file serving behaviour for workspaces containing large number of files so that the workflow controller would not "freeze". This could be achieved in several manners:

1. Simply refuse to serve the full file content if there are more than N files in the workspace, and serve only N files. (N should be configurable by the REANA cluster admins, because some deployments may have worker nodes with a lot of memory, some with smaller memory. For example, N could be set to 10k.)
2. Allow specifying filtering, such as `reana-client ls -w myanalysis --filter name=mydirectory`, both from CLI and from Web UI, and restrict serving N files _after_ the filtering. If we do this, then in theory the researcher could repeat downloading all files by issuing many requests, each with desired filtering.
3. Improve pagination so that only parts of the workspace files are returned, i.e. process at most 10 files using some pre-defined sorting order (e.g. creation timestamps), and then move onto next N, so that all the files do not have to be processed in advance. This could have an advantage in case the workspace is modified afterwards manually, since the pagination could be "lost".

Either of these solutions would prevent "freezing" of the workflow controller component.

## Notes

This issue is related to #641

michaelbuchar:
Sync the workspace listing filter description and limit-exceeded 400 example in the REST OpenAPI docs with reana-workflow-controller so the documented response matches the current backend behavior.

Closes reanahub/reana-workflow-controller#644

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -5267,7 +5267,7 @@
             "type": "integer"
           },
           {
-            "description": "Filter workflow workspace files.",
+            "description": "Filter workflow workspace files by file name, size, or modification date.",
             "in": "query",
             "name": "search",
             "required": false,
@@ -5315,10 +5315,10 @@
             }
           },
           "400": {
-            "description": "Request failed. The incoming payload seems malformed.",
+            "description": "Request failed. The request parameters are invalid or the filtered result set exceeds the configured display limit.",
             "examples": {
               "application/json": {
-                "message": "Field 'size': Must be at least 1."
+                "message": "Too many files to display (e.g. limit=100000). Please use more specific filters to narrow the results. Available filters: file name, size, or last-modified."
               }
             },
             "schema": {
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -2098,7 +2098,7 @@ def get_files(workflow_id_or_name, user, **kwargs):  # noqa
           type: integer
         - name: search
           in: query
-          description: Filter workflow workspace files.
+          description: Filter workflow workspace files by file name, size, or modification date.
           required: false
           type: string
       responses:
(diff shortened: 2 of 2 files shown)


docs/openapi.json
```diff
@@ -5315,10 +5315,10 @@
             }
           },
           "400": {
-            "description": "Request failed. The incoming payload seems malformed.",
+            "description": "Request failed. The request parameters are invalid or the filtered result set exceeds the configured display limit.",
             "examples": {
               "application/json": {
-                "message": "Field 'size': Must be at least 1."
+                "message": "Too many files to display (limit=100000). Please use more specific filters to narrow the results. Available filters: file name, size, or last-modified."
               }
```
tiborsimko, line 5322: The 400 example hardcodes `limit=100000`, but this value is configurable via `WORKSPACE_DISPLAY_FILE_LIMIT`. If an admin changes it, the documented example won't match the actual response. Consider prefixing with "e.g." to make it clear the number is illustrative:

```json
"message": "Too many files to display (e.g. limit=100000). Please use more specific filters..."
```

tiborsimko: (or just remove the parenthesis in the message above)

michaelbuchar: Fixed by adding e.g.

tiborsimko (approved):

PR #772, opened 2026-04-28
refactor(admin): replace dynamic subquery with explicit join

tiborsimko:
-    refactor(admin): replace deprecated datetime.utcnow() 

    `datetime.utcnow()` is deprecated since Python 3.12 in favour of
    the timezone-aware `datetime.now(datetime.UTC)`.

    Drop the `.replace(tzinfo=None)` that stripped the timezone from
    `last_activity` to make it match the naive `utcnow()` result.
    Both `last_activity` (parsed with `%z`) and the new
    `datetime.now(datetime.UTC)` are now timezone-aware UTC values,
    so the subtraction yields the same `timedelta` as before.

-    refactor(admin): replace dynamic subquery with explicit join 

    `reana-admin retention-rules-apply` selected candidate retention
    rules either from `workflow.retention_rules` (a query object,
    under the old `lazy="dynamic"` setting on
    `Workflow.retention_rules`) or by joining onto
    `user.workflows.subquery()` (a subquery built from the dynamic
    `User.workflows` query).

    After dropping `lazy="dynamic"` in reana-db so these
    relationships return regular lists, the previous code no longer
    works: lists have neither a `.subquery()` method nor the
    follow-up `.filter(...)` chain that the rest of the function
    relies on.

    Replace both branches with explicit queries on
    `WorkspaceRetentionRule`: filter by `workflow_id` for the
    single-workflow case, and join `Workflow` filtered by
    `owner_id` for the per-user case. Result is the same set of
    rows, but expressed through standard `Session.query(...)`
    calls.

### DIFF
--- a/reana_server/reana_admin/cli.py
+++ b/reana_server/reana_admin/cli.py
@@ -764,10 +764,10 @@ def retention_rules_apply(
 
     candidate_rules = Session.query(WorkspaceRetentionRule)
     if workflow:
-        candidate_rules = workflow.retention_rules
+        candidate_rules = candidate_rules.filter_by(workflow_id=workflow.id_)
     elif user:
-        candidate_rules = Session.query(WorkspaceRetentionRule).join(
-            user.workflows.subquery()
+        candidate_rules = candidate_rules.join(Workflow).filter(
+            Workflow.owner_id == user.id_
         )
 
     click.echo("Setting the status of all the rules that will be applied to `pending`")
--- a/reana_server/reana_admin/cli.py
+++ b/reana_server/reana_admin/cli.py
@@ -1049,8 +1049,8 @@ def interactive_session_cleanup(
 
         last_activity = datetime.datetime.strptime(
             session_status["last_activity"], "%Y-%m-%dT%H:%M:%S.%f%z"
-        ).replace(tzinfo=None)
-        duration = datetime.datetime.utcnow() - last_activity
+        )
+        duration = datetime.datetime.now(datetime.UTC) - last_activity
         if duration.days >= days:
             if dry_run:
                 click.echo(


reana_server/reana_admin/cli.py
```diff
@@ -1049,8 +1049,8 @@ def interactive_session_cleanup(
 
         last_activity = datetime.datetime.strptime(
             session_status["last_activity"], "%Y-%m-%dT%H:%M:%S.%f%z"
-        ).replace(tzinfo=None)
-        duration = datetime.datetime.utcnow() - last_activity
+        )
+        duration = datetime.datetime.now(datetime.UTC) - last_activity
```
michaelbuchar, line 1053: interactive_session_cleanup() now uses datetime.datetime.now(datetime.UTC) but setup.py still declares python_requires=">=3.8". The datetime.UTC alias was only added in Python 3.11 so on supported 3.8–3.10 interpreters this admin command will fail with AttributeError as soon as the cleanup path is exercised. Maybe let's use datetime.timezone.utc instead if the package still intends to support Python 3.8+?

tiborsimko: Yes, I know. This affects most of our components:

```console
$ rg python_requires r*/setup.py
zsh: no matches found: r*/setu.py
project/reana/src via 🐍 v3.12.11 (reana) $ rg python_requires r*/setup.py
reana-server/setup.py:132:    python_requires=">=3.8",
reana-db/setup.py:72:    python_requires=">=3.8",
reana-workflow-engine-serial/setup.py:76:    python_requires=">=3.8",
reana-workflow-engine-cwl/setup.py:68:    python_requires=">=3.8",
reana-commons/setup.py:104:    python_requires=">=3.8",
reana-client/setup.py:80:    python_requires=">=3.8",
reana-job-controller/setup.py:97:    python_requires=">=3.8",
reana-workflow-engine-snakemake/setup.py:77:    python_requires=">=3.8",
reana/setup.py:75:    python_requires=">=3.8",
reana-workflow-controller/setup.py:92:    python_requires=">=3.8",
reana-workflow-engine-yadage/setup.py:86:    python_requires=">=3.8",
```

But it's not a show-stopper here, since `reana-server` is being released as a Docker container image only, and we pin Python version in `Dockerfile`. So it will never run on older Pythons, so to speak.

Let's bump `python_requires` later in all components...

michaelbuchar (approved):

PR #777, opened 2026-05-18
feat(rest): add token revocation management endpoint

michaelbuchar:
Closes #776

### DIFF
--- a/reana_server/rest/users.py
+++ b/reana_server/rest/users.py
@@ -9,10 +9,12 @@
 """Reana-Server User Endpoints."""
 
 import logging
+import secrets
 import traceback
 
 from bravado.exception import HTTPError
-from flask import Blueprint, jsonify
+from flask import Blueprint, Response, jsonify, request
+from marshmallow import Schema, ValidationError, fields
 from reana_db.database import Session
 from reana_db.models import AuditLogAction, ResourceType, User, UserWorkflow, Workflow
 from reana_commons.config import (
@@ -23,11 +25,32 @@ from reana_commons.email import send_email, REANA_EMAIL_RECEIVER
 from reana_commons.errors import REANAEmailNotificationError
 
 from reana_server import __version__
-from reana_server.config import REANA_HOSTNAME
+from reana_server.config import (
+    REANA_HOSTNAME,
+    REANA_TOKEN_MANAGEMENT_SECRET,
+)
 from reana_server.decorators import signin_required
-from reana_server.utils import JinjaEnv, serialize_utc_datetime
+from reana_server.utils import (
+    JinjaEnv,
+    _get_admin_user_or_raise,
+    _get_user_by_criteria,
+    revoke_access_token_of_user,
+    serialize_utc_datetime,
+)
 
 blueprint = Blueprint("users", __name__)
+
+
+class DeleteTokenBodySchema(Schema):
+    """Schema for delete_token endpoint body."""
+
+    user_id = fields.Str()
+    email = fields.Str()
+
+
+def _management_response(status_code: int, message: str, **payload):
+    """Build a management endpoint response including HTTP status in the body."""
+    return jsonify(message=message, status=status_code, **payload), status_code
 
 
 def _serialize_user_quota(user):
@@ -51,6 +74,17 @@ def _serialize_user_quota(user):
             ),
         }
     return quota
+
+
+def _check_token_management_secret() -> tuple[Response, int] | None:
+    if not REANA_TOKEN_MANAGEMENT_SECRET:
+        return _management_response(403, "Token management endpoint is not configured.")
+
+    secret = request.headers.get("X-Token-Management-Secret", "")
+    if not secrets.compare_digest(secret, REANA_TOKEN_MANAGEMENT_SECRET):
+        return _management_response(401, "Unauthorized")
+
+    return None
 
 
 @blueprint.route("/you", methods=["GET"])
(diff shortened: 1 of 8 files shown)


reana_server/rest/users.py
```diff
@@ -391,6 +425,177 @@ def request_token(user):
...
+          description: Unauthorized.
+          schema:
+            type: object
+            properties:
+              status:
+                type: integer
+              message:
+                type: string
+        403:
+          description: Token management functionality is not enabled.
```
tiborsimko: 🤖 **403 wording in the OpenAPI docstring doesn't match what the endpoint actually returns**

The docstring says the 403 means *"Token management functionality is not enabled."*, but the code at `_check_token_management_secret` (line 81) returns the body *"Token management endpoint is not configured."* — so a client reading the spec sees one phrase and gets back another. Easiest to align the docstring with the actual response body:

```diff
         403:
-          description: Token management functionality is not enabled.
+          description: Token management endpoint is not configured.
           schema:
```

Then regenerate `docs/openapi.json` and copy it across to the bundled spec in reana-commons so all three places agree.

reana_server/utils.py
```diff
@@ -312,10 +312,73 @@ def grant_access_token_to_user(
             # Token was granted successfully, attach canonical message so callers can still display it
             setattr(e, "log_msg", log_msg)
             raise
+        except Exception as e:
```
tiborsimko, line 315: 🤖 **`grant_access_token_to_user` gets a new behaviour change that isn't covered by this PR's tests**

This PR is about revocation, but the same try/except shape was also added to `grant_access_token_to_user` here for symmetry. The effect on grant is real: any non-`REANAEmailNotificationError` exception raised during email send (e.g. a `FileNotFoundError` from `REANAConfig.load("ui")` if the UI config isn't mounted) used to propagate as itself; now it gets wrapped into a synthetic `REANAEmailNotificationError`. Callers that previously saw the underlying error class will now see the wrapped one — including the `token-grant` CLI command and anywhere `grant_access_token_to_user` is used.

For revocation that change is covered by `test_revoke_token_reports_success_when_email_preparation_fails` in this PR. For grant there's no analogous test, so the new behaviour ships silently.

Two options:

1. **Keep the symmetry** and add a sibling test, e.g. `test_grant_token_reports_success_when_email_preparation_fails` modelled directly on the revoke version — patches `REANAConfig.load` to raise `FileNotFoundError`, asserts the token was still granted and the CLI prints both the success line and the "something went wrong while sending email" line.
2. **Tighten scope** and revert the `except Exception` block in `grant_access_token_to_user` (lines 315–318), leaving only the existing `except REANAEmailNotificationError`. The revoke helper still gets the broader catch, which is what the PR is actually about.

I'd lean towards option 1 — the symmetry is genuinely nicer and the test is ~15 lines — but either is fine, it's mostly about not leaving the change untested.

reana_server/utils.py
```diff
@@ -312,10 +312,73 @@ def grant_access_token_to_user(
...
+    if not active_token:
+        raise ValueError(
+            f"User {user.id_} ({user.email}) does not have an active access token."
+        )
+
+    revoked_token = active_token.token
+    active_token.status = UserTokenStatus.revoked
+    Session.commit()
+
+    log_msg = f"User token {revoked_token} ({user.email}) was successfully revoked."
```
tiborsimko: 🤖 **Mirror `include_token_in_log` from the grant helper while we're here**

This line builds a `log_msg` that includes the raw revoked token value, and a few lines down that same string ends up in the audit log via `revoked_by.log_action(AuditLogAction.revoke_token, {..., "reana_admin": log_msg})`. So the plaintext token gets persisted in `AuditLog.details` for every revocation.

That behaviour is pre-existing — master's CLI does the exact same thing, and I wouldn't push back on it on those grounds alone. But two things make this PR the right moment to tighten it:

1. The grant counterpart already has an `include_token_in_log: bool = True` knob (line 265), and one existing caller passes `include_token_in_log=False` (line 608). The codebase already shows awareness that token-in-audit-log is a thing one might want to opt out of — but the revoke side never grew the equivalent.
2. The PR adds a brand-new REST entry point to this code path. Audit logs are more likely to be scraped/exported centrally than CLI output, so the disclosure surface grows a bit even though the underlying behaviour doesn't.

Concrete suggestion — mirror the grant signature:

```diff
 def revoke_access_token_of_user(
     user: User,
     *,
     revoked_by: Optional[User] = None,
     send_notification_email: bool = True,
+    include_token_in_log: bool = True,
     requested_via: str = "reana_server",
 ) -> Tuple[str, str]:
```

Then build `log_msg` conditionally (matching the grant helper's pattern at line 281–283):

```python
log_msg = f"User token for {user.email} was successfully revoked."
if include_token_in_log:
    log_msg += f"\n\nToken: {revoked_token}"
```

Pass `include_token_in_log=False` from the REST caller in `users.py` (the CLI can keep the default `True` so admins running `flask reana-admin token-revoke` still see the token in their terminal, matching today's behaviour).

If you'd rather defer the cleanup, that's also fine — happy to open a separate follow-up issue instead. But since the helper is being introduced fresh in this PR, slotting the flag in now is a small change with no migration cost.

reana_server/rest/users.py
```diff
@@ -391,6 +425,177 @@ def request_token(user):
...
+    user_id = json_body.get("user_id")
+    email = json_body.get("email")
+    if int(bool(user_id)) + int(bool(email)) != 1:
+        return _management_response(
+            400, "Exactly one of `user_id` or `email` must be provided."
+        )
+
+    user = _get_user_by_criteria(user_id, email)
+    if not user:
+        return _management_response(404, "User not found.")
```
tiborsimko: 🤖 **404 vs 409 gives the secret holder a user-enumeration oracle**

The endpoint returns 404 when the user doesn't exist and 409 when they exist but have no active token. That distinction lets a caller probe arbitrary emails or UUIDs and learn which ones map to real REANA users, even when there's nothing to revoke.

Worth taking seriously here because of the deployment model: per the docs PR, this endpoint exists specifically so token management can be done *"remotely without Kubernetes cluster access"*. The secret holder is intentionally a less-privileged principal than a cluster admin — they have HTTPS and a secret, but no `kubectl`, no database access, and no other window into the user table. The REST endpoint is their only channel, so any signal it emits is the only signal they can get. The 404/409 split turns that channel into a user-enumeration oracle that they wouldn't otherwise have.

Easiest fix is to collapse the two outcomes into one indistinguishable response. Two reasonable shapes:

```python
# Option A: always 200, with a status field for genuine callers who care
return _management_response(
    200,
    "No active token to revoke.",
    reana_token={"status": "no_active_token"},
)

# Option B: keep 404, with a generic message that doesn't distinguish the two cases
return _management_response(404, "No active token to revoke for the given user.")
```

I'd lean towards (B) — it stays closer to standard REST semantics (the addressed resource — an active token — isn't there) and doesn't require any client-side branching. Either one closes the enumeration channel.

While you're here, the matching 409 branch a few lines down (`return _management_response(409, str(e))` from `revoke_access_token_of_user`'s `ValueError`) needs the same treatment — currently its message includes both the UUID and the email (`f"User {user.id_} ({user.email}) does not have an active access token."`), which leaks the full identity back to the caller.

reana_server/rest/users.py
```diff
@@ -391,6 +425,177 @@ def request_token(user):
...
+
+    admin = Session.query(User).filter_by(id_=ADMIN_USER_ID).one_or_none()
```
tiborsimko: 🤖 **Silent audit-log skip if `ADMIN_USER_ID` is misconfigured**

`one_or_none()` returns `None` if there's no row matching `ADMIN_USER_ID` — e.g. if the env var is unset, points to a UUID that doesn't exist in the database, or was rotated. That `None` then gets passed straight into the helper as `revoked_by=admin`, and `revoke_access_token_of_user` does:

```python
if revoked_by:
    revoked_by.log_action(AuditLogAction.revoke_token, {...})
```

So the audit log entry is silently skipped, the token gets revoked anyway, and the operator gets back a `200 Access token revoked.` response with no indication that the audit trail wasn't written. That's the worst-case shape of an audit-log failure: the action happens, but the log of who did it doesn't.

Realistic likelihood: probably low — `ADMIN_USER_ID` is set at deploy time and rarely changes. But this endpoint is the very place where a missing audit trail would matter most (a remote actor with the management secret revoking tokens).

A couple of low-cost options:

1. **Refuse the request when no admin can be resolved**, since the endpoint can't fulfil its audit-trail contract:
   ```python
   admin = Session.query(User).filter_by(id_=ADMIN_USER_ID).one_or_none()
   if not admin:
       logging.error("ADMIN_USER_ID does not resolve to a known user; refusing to revoke.")
       return _management_response(500, "Server misconfiguration.")
   ```
2. **Log loudly when it happens** but still let the revocation through:
   ```python
   admin = Session.query(User).filter_by(id_=ADMIN_USER_ID).one_or_none()
   if not admin:
       logging.error(
           "ADMIN_USER_ID %s does not resolve to a known user; "
           "token revocation will proceed without an audit log entry.",
           ADMIN_USER_ID,
       )
   ```

(1) is stricter and probably the right default for a security-relevant endpoint. (2) preserves availability at the cost of leaving the gap silent in the response — but at least operators see the warning in the pod logs. Same consideration applies to the CLI `token-revoke` command, which has the identical `one_or_none()` pattern.

reana_server/rest/users.py
```diff
@@ -391,6 +425,177 @@ def request_token(user):
...
+        )
+
+    try:
+        json_body = DeleteTokenBodySchema().load(json_body)
+    except ValidationError as e:
+        return _management_response(400, f"Invalid request. Errors: {e.messages}")
+
+    user_id = json_body.get("user_id")
+    email = json_body.get("email")
+    if int(bool(user_id)) + int(bool(email)) != 1:
```
tiborsimko: 🤖 **Readability nit on the "exactly one of" check**

`int(bool(user_id)) + int(bool(email)) != 1` reads a bit obliquely — the reader has to mentally evaluate the arithmetic to see that you're enforcing "exactly one truthy". The error message right below already tells the reader what the intent is, so the check should ideally match it as directly as possible.

Two cleaner equivalents:

```python
# Option A: boolean XOR
if bool(user_id) == bool(email):
    return _management_response(
        400, "Exactly one of `user_id` or `email` must be provided."
    )

# Option B: spell out the two failure cases
if not user_id and not email:
    return _management_response(400, "Either `user_id` or `email` must be provided.")
if user_id and email:
    return _management_response(
        400, "`user_id` and `email` are mutually exclusive."
    )
```

(B) is more lines but gives the caller a more actionable error — they learn whether they sent zero identifiers or two. (A) is a one-line swap that preserves current behaviour. Either is fine; flagging mostly for the arithmetic-on-booleans pattern.

reana_server/rest/users.py
michaelbuchar: Fixed

reana_server/utils.py
michaelbuchar: Fixed

reana_server/utils.py

reana_server/rest/users.py
michaelbuchar: Fixed

reana_server/rest/users.py
michaelbuchar: Fixed

reana_server/rest/users.py
michaelbuchar: Fixed using option A

tiborsimko (approved):

PR #779, opened 2026-05-19
feat(rest): surface info on minimum Kubernetes job user ID

### LINKED ISSUE #951: helm: allow cluster admins to configure the minimum accepted `kubernetes_uid` value
## Current behaviour

User-supplied `kubernetes_uid` workflow hint values are accepted as long as they are `>= 100`. This is hard-coded in REANA ecosystem, e.g. in the `reana-job-controller` component. A REANA deployment may want to strengthen the policy to allow only higher user IDs, similar to OpenShift.

  ## Expected behaviour

The minimum accepted `kubernetes_uid` value should be configurable per REANA deployment via the Helm value files, for example by means of an environment variable such as  `REANA_KUBERNETES_JOBS_MIN_USER_UID`.

tiborsimko:
Expose the cluster-configured minimum accepted runtime container user ID (`REANA_KUBERNETES_JOBS_MIN_USER_UID`, default 100) on the `/info` endpoint so that users and administrators can verify the configured value before submitting workflows with `kubernetes_uid` hints.

Closes reanahub/reana#951

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -349,6 +349,23 @@ def _get_rate_limit(env_variable: str, default: str) -> str:
         return default
 
 
+def _get_int_env_variable(env_variable: str, default: int) -> int:
+    """Return an integer environment variable value or fall back to default."""
+    env_value = os.getenv(env_variable)
+    if env_value is None:
+        return default
+    try:
+        return int(env_value)
+    except ValueError:
+        logging.warning(
+            "Invalid %s=%r; falling back to %s.",
+            env_variable,
+            env_value,
+            default,
+        )
+        return default
+
+
 # Note: users that are connecting via reana-client will be treated as guests by the Invenio framework
 RATELIMIT_GUEST_USER = _get_rate_limit("REANA_RATELIMIT_GUEST_USER", "20 per second")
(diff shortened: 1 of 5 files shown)


reana_server/config.py
```diff
@@ -582,6 +582,16 @@ def _get_rate_limit(env_variable: str, default: str) -> str:
 https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup.
 """
 
+REANA_KUBERNETES_JOBS_MIN_USER_UID = int(
+    os.getenv("REANA_KUBERNETES_JOBS_MIN_USER_UID", 100)
```
michaelbuchar: ### AI-aided review

The new REANA_KUBERNETES_JOBS_MIN_USER_UID = int(os.getenv(...)) parse will make reana-server fail at import/startup time if the env var is present but empty or non-numeric. That may be acceptable if you want strict fail-fast config validation, but if this is meant to be robust across ad hoc or externally managed deployments, I’d parse more defensively or fall back to 100 with a clearer warning.

tiborsimko: Thanks, addressed.

michaelbuchar (approved):

Issue #516, opened 2022-07-07
launcher: verify existence of granted tokens for non-CERN users


tiborsimko:
Currently, a user can launch a workflow after successful login even if they don't have any previously granted access token for CLI use.

This is OK for OIDC CERN users, which can be recognised as `account_info.user.email` being of the form `*@cern.ch`, or, perhaps better due to existence of primary/secondary accounts which may be non-CERN addresses, which can be recognised by  `account_info.external_id` not containing character `@`. (It usually contains CERN login name, such as `johndoe`.)

However, for non-CERN users, such as those possibly coming from OIDC eduGAIN or from social login mechanisms (if allowed), it would be good to require an explicit demand for permission to use REANA first; just like we do for CLI tokens.  These users can be recognised as `account_info.external_id` being of the form `1234@google`, `5678@github`, `1357@unige2468`, etc.  That is, we can check the presence of `@` there.

The launcher can have a look at the user identity, and if it is deemed non-CERN, and if the account does not have an already-approved token, it should basically redirect to "welcome" page where we explain people that they must first ask for a token etc:

![welcome](https://user-images.githubusercontent.com/517546/177774701-e0991efa-3218-4a51-a19d-47bb44e72b46.png)

Note: after login, `account_info.external_id` is basically being stored in `__reana.user_` table in the `username` column;  so the launcher functionality can simply consult that column to judge about CERN-ness of logged-in users.

PR #780, opened 2026-05-26
feat(scheduler): cap concurrent workflows per compute backend

CameronMcClymont:
Workflows whose steps run on external backends (HTCondor, Slurm) consume fewer Kubernetes resources, but the scheduler does not take this into account. Even fully external workflows are still subject to the concurrency limit.

This PR introduces more generous concurrency limits for external workflows rather than subjecting them to a coarse global limit.

Closes https://github.com/reanahub/reana-workflow-controller/issues/663

### DIFF
--- a/reana_server/complexity.py
+++ b/reana_server/complexity.py
@@ -26,9 +26,10 @@ def validate_job_memory_limits(complexity: List[Tuple[int, float]]) -> None:
     :param complexity: workflow complexity list which consists of number of initial jobs and the memory in bytes they require. (e.g. [(8, 1073741824), (5, 2147483648)])
     :raises REANAKubernetesMemoryLimitExceeded: If workflow job memory limits exceed the maximum memory limit that users can assign to their job containers.
     """
-    if not complexity:
+    k8s_memories = [mem for jobs, mem in complexity if jobs > 0]
+    if not k8s_memories:
         return None
-    max_job_memory = max(complexity, key=lambda x: x[1])[1]
+    max_job_memory = max(k8s_memories)
 
     if (
         REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT_IN_BYTES
@@ -41,14 +42,33 @@ def validate_job_memory_limits(complexity: List[Tuple[int, float]]) -> None:
 
 
 def get_workflow_min_job_memory(complexity):
-    """Return minimal job memory from workflow complexity.
+    """Return the smallest per-job memory requirement among initial Kubernetes steps.
 
-    :param complexity: workflow complexity list which consists of number of initial jobs and the memory in bytes they require. (e.g. [(8, 1073741824), (5, 2147483648)])
-    :return: minimal job memory (e.g. 1073741824)
+    :param complexity: list of (k8s_jobs, memory_bytes) tuples, one per initial
+        workflow step. k8s_jobs is 0 for steps that run on external backends
+        (HTCondor, Slurm) and >0 for steps that run on Kubernetes.
+    :return: minimum memory in bytes across initial Kubernetes steps, or 0 if
+        there are no initial Kubernetes steps (e.g. the workflow starts on an
+        external backend). A return value of 0 causes the Kubernetes node memory
+        check to be skipped entirely.
     """
-    if not complexity:
+    k8s_memories = [mem for jobs, mem in complexity if jobs > 0]
+    if not k8s_memories:
         return 0
-    return min(complexity, key=lambda x: x[1])[1]
+    return min(k8s_memories)
+
+
+def _build_estimator(workflow_type, reana_yaml):
+    if workflow_type == "serial":
+        return SerialComplexityEstimator(reana_yaml)
+    elif workflow_type == "yadage":
+        return YadageComplexityEstimator(reana_yaml)
+    elif workflow_type == "cwl":
+        return CWLComplexityEstimator(reana_yaml)
+    elif workflow_type == "snakemake":
+        return SnakemakeComplexityEstimator(reana_yaml)
+    else:
+        raise Exception("Workflow type '{0}' is not supported".format(workflow_type))
 
 
 def estimate_complexity(workflow_type, reana_yaml):
@@ -57,27 +77,25 @@ def estimate_complexity(workflow_type, reana_yaml):
     :param workflow_type: A supported workflow specification type.
     :param reana_yaml: REANA YAML specification.
     """
-
-    def build_estimator(workflow_type, reana_yaml):
-        if workflow_type == "serial":
-            return SerialComplexityEstimator(reana_yaml)
-        elif workflow_type == "yadage":
-            return YadageComplexityEstimator(reana_yaml)
-        elif workflow_type == "cwl":
-            return CWLComplexityEstimator(reana_yaml)
-        elif workflow_type == "snakemake":
-            return SnakemakeComplexityEstimator(reana_yaml)
-        else:
-            raise Exception(
-                "Workflow type '{0}' is not supported".format(workflow_type)
-            )
-
-    estimator = build_estimator(workflow_type, reana_yaml)
+    estimator = _build_estimator(workflow_type, reana_yaml)
     try:
         complexity = estimator.estimate_complexity()
     except Exception:
         return []
     return complexity
+
+
+def workflow_compute_backends(workflow_type: str, reana_yaml: dict) -> List[str]:
+    """Return the sorted set of step-level compute backends a workflow uses.
+
+    Scans all steps so that hybrid workflows are classified with every
+    backend they touch (e.g. ``["htcondor", "kubernetes"]``).
+    """
+    try:
+        backends = _build_estimator(workflow_type, reana_yaml).compute_backends()
+    except Exception:
+        return ["kubernetes"]
+    return sorted(backends) if backends else ["kubernetes"]
 
 
 class ComplexityEstimatorBase:
@@ -97,6 +115,10 @@ class ComplexityEstimatorBase:
         """Parse REANA workflow specification tree."""
         raise NotImplementedError
 
+    def compute_backends(self) -> set:
+        """Return the set of compute backends used across the whole workflow."""
+        raise NotImplementedError
+
     def estimate_complexity(self, initial_step="init"):
         """Estimate complexity in parsed REANA workflow tree."""
         steps = self.parse_specification(initial_step)
@@ -109,10 +131,13 @@ class ComplexityEstimatorBase:
             complexity += step["complexity"]
         return complexity
 
+    def _get_step_backend(self, step):
+        """Get the compute backend a step runs on, defaulting to Kubernetes."""
+        return step.get("compute_backend") or "kubernetes"
+
     def _get_number_of_jobs(self, step):
         """Get number of jobs based on compute backend."""
-        backend = step.get("compute_backend")
-        if backend and backend != "kubernetes":
+        if self._get_step_backend(step) != "kubernetes":
             return 0
         return 1
 
@@ -138,6 +163,13 @@ class SerialComplexityEstimator(ComplexityEstimatorBase):
             tree.append({name: {"complexity": complexity}})
         return tree
 
+    def compute_backends(self) -> set:
+        """Return the set of compute backends used across the workflow."""
+        steps = self.specification.get("steps", [])
+        if not steps:
+            return {"kubernetes"}
+        return {self._get_step_backend(step) for step in steps}
+
     def parse_specification(self, initial_step):
         """Parse and filter out serial workflow specification tree."""
         spec_steps = self.specification.get("steps", [])
@@ -149,6 +181,39 @@ class SerialComplexityEstimator(ComplexityEstimatorBase):
 
 class YadageComplexityEstimator(ComplexityEstimatorBase):
     """REANA Yadage workflow complexity estimation."""
+
+    @staticmethod
+    def _get_stage_resources(stage):
+        """Return the list of resources declared for a Yadage stage."""
+        resources = (
+            stage.get("scheduler", {})
+            .get("step", {})
+            .get("environment", {})
+            .get("resources", [])
+        )
+        return resources if isinstance(resources, list) else []
+
+    def _get_stage_compute_backend(self, stage):
+        """Return the compute backend resource dict for a Yadage stage.
+
+        When a stage declares ``compute_backend`` more than once, the last
+        declaration wins, matching how reana-workflow-engine-yadage builds its
+        job request (``_get_resources`` in ``externalbackend.py`` iterates all
+        resources and overwrites), so a stage is classified against the backend
+        it actually runs on.
+        """
+        return next(
+            filter(
+                lambda r: isinstance(r, dict) and "compute_backend" in r.keys(),
+                reversed(self._get_stage_resources(stage)),
+            ),
+            {},
+        )
+
+    @staticmethod
+    def _get_nested_stages(stage):
+        """Return the nested stages of a Yadage stage (empty if none)."""
+        return stage.get("scheduler", {}).get("workflow", {}).get("stages", [])
 
     def _parse_steps(self, stages, initial_step):
         """Parse and filter out Yadage workflow tree."""
@@ -165,19 +230,8 @@ class YadageComplexityEstimator(ComplexityEstimatorBase):
             return False
 
         def _get_stage_complexity(stage):
-            resources = (
-                stage.get("scheduler", {})
-                .get("step", {})
-                .get("environment", {})
-                .get("resources", [])
-            )
-            compute_backend = next(
-                filter(
-                    lambda r: isinstance(r, dict) and "compute_backend" in r.keys(),
-                    resources,
-                ),
-                {},
-            )
+            resources = self._get_stage_resources(stage)
+            compute_backend = self._get_stage_compute_backend(stage)
             k8s_memory_limit = next(
                 filter(
                     lambda r: isinstance(r, dict)
@@ -275,6 +329,26 @@ class YadageComplexityEstimator(ComplexityEstimatorBase):
 
         return _parse_stages(stages)
 
+    def compute_backends(self) -> set:
+        """Return the set of compute backends used across the workflow."""
+
+        def _collect_stages(stages):
+            backends = set()
+            for stage in stages:
+                nested = self._get_nested_stages(stage)
+                # A wrapper stage delegates its complexity to its nested stages
+                # (see ``_populate_complexity``), so its own backend is only
+                # relevant when there are no nested stages.
+                if nested:
+                    backends |= _collect_stages(nested)
+                else:
+                    compute_backend = self._get_stage_compute_backend(stage)
+                    backends.add(self._get_step_backend(compute_backend))
+            return backends
+
+        stages = self.specification.get("stages", [])
+        return _collect_stages(stages) if stages else {"kubernetes"}
+
     def parse_specification(self, initial_step):
         """Parse Yadage workflow specification tree."""
         steps = self._parse_steps(self.specification["stages"], initial_step)
@@ -286,6 +360,28 @@ class YadageComplexityEstimator(ComplexityEstimatorBase):
 class CWLComplexityEstimator(ComplexityEstimatorBase):
     """REANA CWL workflow complexity estimation."""
 
+    @staticmethod
+    def _get_step_hints(step):
+        """Return the effective hints of a CWL step, or an empty dict if none.
+
+        All hints are scanned and, for any given key, the first hint declaring
+        it wins. This mirrors how reana-workflow-engine-cwl resolves hints at
+        execution time (``ReanaPipeline._get_hint`` in ``cwl_reana.py``), so a
+        step is classified against the backend it actually runs on. Taking only
+        the last hint would misclassify the canonical shape, where a
+        ``class: reana`` hint carrying ``compute_backend`` is followed by, say,
+        a ``ResourceRequirement`` hint.
+
+        Robust to ``hints`` being absent, ``None`` or an empty list.
+        """
+        effective = {}
+        for hint in step.get("hints") or []:
+            if not isinstance(hint, dict):
+                continue
+            for key, value in hint.items():
+                effective.setdefault(key, value)
+        return effective
+
     def _parse_steps(self, workflow):
         """Parse CWL workflow specification tree."""
         tree = {}
@@ -294,7 +390,7 @@ class CWLComplexityEstimator(ComplexityEstimatorBase):
         for step in steps:
             name = step.get("id")
             run = step.get("run")
-            hints = step.get("hints", [{}]).pop()
+            hints = self._get_step_hints(step)
             # Parse scatter params
             scatter = step.get("scatter")
             scatter_params = None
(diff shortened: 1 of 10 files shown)


reana_server/complexity.py
```diff
@@ -41,14 +41,20 @@ def validate_job_memory_limits(complexity: List[Tuple[int, float]]) -> None:
...
+    :param complexity: list of (k8s_jobs, memory_bytes) tuples, one per initial
+        workflow step. k8s_jobs is 0 for steps that run on external backends
+        (HTCondor, Slurm) and >0 for steps that run on Kubernetes.
+    :return: minimum memory in bytes across initial Kubernetes steps, or 0 if
+        there are no initial Kubernetes steps (e.g. the workflow starts on an
+        external backend). A return value of 0 causes the Kubernetes node memory
+        check to be skipped entirely.
     """
-    if not complexity:
+    k8s_memories = [mem for jobs, mem in complexity if jobs > 0]
```
michaelbuchar, line 55: This fixes estimators that already mark external-backend steps with `jobs=0`, but Snakemake does not seem to do that today. Since Snakemake supports per-step `compute_backend`, external-backend Snakemake workflows may still end up with a non-zero `min_job_memory` and still hit the Kubernetes memory check. Should we handle Snakemake here as well, or treat this as a partial fix?

CameronMcClymont: Fixed to skip Kubernetes memory check for external Snakemake steps in line with other workflow languages.

reana_server/scheduler.py
```diff
@@ -71,7 +80,23 @@ def check_concurrent_workflows_limit() -> Optional[str]:
...
+                    func.array_length(Workflow.complexity, 1).is_(None),
+                    # Count only workflows that have at least one Kubernetes step.
+                    # ``complexity`` is a 2-D array of [job_count, memory] rows
+                    # (Postgres arrays are 1-indexed), so ``complexity[i][1]`` is
+                    # the job count of step ``i`` and is >0 for Kubernetes steps.
+                    text(
+                        "EXISTS ("
+                        "SELECT 1 FROM generate_series(1, array_length(complexity, 1)) AS i "
+                        "WHERE complexity[i][1] > 0"
+                        ")"
```
michaelbuchar: Hybrid workflows are still dropped from the running-workflow count after submission, so #780 can under-enforce the Kubernetes concurrency cap. The PR correctly adds all-step classification via workflow_uses_kubernetes() in reana_server/complexity.py:83-96, and publish_workflow_submission() even notes that stored complexity only reflects initial steps in reana_server/utils.py:399-413. But check_concurrent_workflows_limit() still counts already-running workflows only from Workflow.complexity in reana_server/scheduler.py:64-91. For a workflow that starts on HTCondor/Slurm and later runs on Kubernetes, those stored complexity rows can still be all jobs=0, so the workflow will not consume a concurrency slot once it is queued/running. I reproduced the underlying mismatch locally with a minimal serial hybrid spec: estimate_complexity() returned [(0, 2147483648.0)] for an external-first, later-Kubernetes workflow. The new scheduler tests in tests/test_scheduler.py:190-224 cover only K8S_COMPLEXITY vs EXTERNAL_COMPLEXITY; they do not cover the critical case uses_kubernetes=True with initial-step complexity=[(0, ...)].

CameronMcClymont: I have pushed a fix for this but haven't squashed it because a cleaner option would be to add a `uses_kubernetes` column in reana-db which would make the code in this PR a lot cleaner with less 'magic values' and commenting necessary. I will open the reana-db PR and push one more commit here and then we can choose which route to go down.

CameronMcClymont: Tbh, the reana-db route is much much cleaner and avoids forcing this 'does this workflow ever touch Kubernetes' boolean through the `complexity` column when it should really have its own. reana-db PR is here: https://github.com/reanahub/reana-db/pull/267

michaelbuchar (dismissed):
This looks good to me, only thing missing is fixing the CI checks!

reana_server/scheduler.py
```diff
@@ -61,8 +61,15 @@ def check_memory_availability(
...
-def check_concurrent_workflows_limit() -> Optional[str]:
-    """Check upper limit on running REANA batch workflows."""
+def check_concurrent_workflows_limit(uses_kubernetes: bool = True) -> Optional[str]:
+    """Check upper limit on running REANA batch workflows.
+
+    Workflows that run exclusively on external backends (HTCondor, Slurm) do not
+    consume Kubernetes resources: they bypass this check and are excluded when
+    counting already-running workflows.
+    """
+    if not uses_kubernetes:
```
tiborsimko: **🤖 RWC663-02: `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` is silently neutered for external workflows (HIGH)**

The PR turns the cluster-wide concurrency cap into an advisory cap that only applies to Kubernetes-job workflows. External-only workflows neither check the cap on their own admission nor count toward it for any future check.

`reana-server/reana_server/scheduler.py:64-86`:

```python
def check_concurrent_workflows_limit(uses_kubernetes: bool = True) -> Optional[str]:
    if not uses_kubernetes:
        return None                       # (a) external workflow → no cap check
    error = None
    try:
        running_workflows = (
            Session.query(func.count())
            .filter(or_(Workflow.status == RunStatus.pending,
                        Workflow.status == RunStatus.running),
                    Workflow.uses_kubernetes) # (b) external workflows don't count
            .scalar()
        )
        if running_workflows >= REANA_MAX_CONCURRENT_BATCH_WORKFLOWS:
            ...
```

A user submitting 10,000 HTCondor-only workflows admits 10,000 workflow-engine pods on Kubernetes. The engine always runs on k8s — it is the orchestrator pod that talks to HTCondor/Slurm — and lives for the entire wall-clock duration of the external jobs, which for HTCondor can be hours or days. The original cap was, among other things, protection against exactly this kind of orchestrator-pod explosion, and that protection is silently gone after upgrade. `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` is the only global workflow-count cap REANA has — there is no per-backend cap, no per-user concurrency cap, and quotas are tracked post-consumption rather than as admission control. (Per-workflow Dask caps exist — `REANA_DASK_CLUSTER_MAX_MEMORY_LIMIT`, `REANA_DASK_CLUSTER_MAX_NUMBER_OF_WORKERS` — but nothing caps the *number of concurrent Dask workflows*.)

Two further pieces of evidence the cap's semantics shifted in ways the PR does not acknowledge:

1. **Dashboard and scheduler disagree.** `ClusterHealth.get_workflow_health()` (`reana-server/reana_server/status.py:655-676`) reports `running + pending` against `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` and **does not** filter by `uses_kubernetes`. After this PR the dashboard will display "30/30, cluster full" while the scheduler is still admitting workflows. Operators reading the dashboard will be misled.
2. **Per-user priority and scheduler cap disagree.** `User.get_workflow_overload_priority()` (`reana-db/reana_db/models.py:329-351`) counts pending+running **without** filtering by `uses_kubernetes`, so external workflows still depress the user's priority. Two views of the same population, different filters.

Suggested fix — preserve the cap as a safety guarantee rather than removing it for external workflows. The right shape is **per-backend caps** along the compute-backend axis, optionally complemented by a **per-resource cap** along the Dask axis. Sketch:

```
REANA_MAX_CONCURRENT_K8S_BATCH_WORKFLOWS       = 30   # renamed-with-meaning of the
                                                       # current variable; counts
                                                       # workflows using the k8s backend
REANA_MAX_CONCURRENT_HTCONDOR_BATCH_WORKFLOWS  = 200  # new; bounds engine pods for
                                                       # HTCondor-using workflows
REANA_MAX_CONCURRENT_SLURM_BATCH_WORKFLOWS     = 200  # new; same for Slurm

REANA_MAX_CONCURRENT_DASK_WORKFLOWS            = 5    # new; orthogonal to the above,
                                                       # bounds workflows that request
                                                       # a Dask cluster regardless of
                                                       # their step backends
```

The defaults are illustrative, not prescriptive — operators pick numbers that match their cluster. Each cap is applied independently in `check_concurrent_workflows_limit()`; a hybrid k8s+HTCondor workflow counts against both backend caps, a Dask-on-HTCondor workflow counts against the HTCondor cap and the Dask cap. This composes naturally with the data-model change in [finding RWC663-03](https://github.com/reanahub/reana-db/pull/267#discussion_r3421389305) (`Workflow.compute_backends` ARRAY for the backend axis, the existing `workflow.services` relation for Dask).

Important: the per-DAG memory awareness must stay. `check_memory_availability()` already computes `workflow_min_job_memory` from the spec at submission time (`reana-server/reana_server/utils.py:417` → `complexity.py:get_workflow_min_job_memory`) and checks it against the largest node's free memory before admission. This is exactly the "is a DAG's smallest k8s step able to land on the cluster right now?" check that effective scheduling depends on, and the PR correctly keeps it. The cap rework should add per-backend slot accounting **on top of** the existing memory check, not replace it.

Whichever shape lands, three things must move together:

- `ClusterHealth.get_workflow_health()` updated so the dashboard reports against the same per-backend counts the scheduler uses (one tile per backend, or an aggregate that matches the scheduler's admission decision exactly).
- `User.get_workflow_overload_priority()` aligned so the priority calculation and the cap calculation see the same population.
- `REANA_WORKFLOW_SCHEDULING_READINESS_CHECK_LEVEL` documentation refreshed: the existing `concurrent` / `memory` / `all_checks` knob (`reana-server/reana_server/config.py:208-230`) currently implies a single concurrency check; it now needs to call out that "concurrent" is the OR of all per-backend caps.

A weighted-cap variant — count an external workflow as e.g. 0.1 of a slot, sum-based admission — would also preserve the original safety property and is mechanically simpler than multiple caps, but at the cost of operator legibility: the cap stops meaning "this many workflows" and starts meaning "this much weighted footprint", which is harder to set and reason about than a per-backend number that maps onto operator intuitions about HTCondor capacity, Dask capacity, etc.

reana_server/complexity.py
```diff
@@ -57,29 +77,28 @@ def estimate_complexity(workflow_type, reana_yaml):
...
-    estimator = build_estimator(workflow_type, reana_yaml)
+    estimator = _build_estimator(workflow_type, reana_yaml)
     try:
         complexity = estimator.estimate_complexity()
     except Exception:
         return []
     return complexity
 
 
+def workflow_uses_kubernetes(workflow_type: str, reana_yaml: dict) -> bool:
```
tiborsimko: **🤖 RWC663-01: Dask workflows with external step backends bypass the concurrent-workflow cap despite creating heavy Kubernetes resources (HIGH)**

A note on scope before the framing: this finding is **not** about the standard Dask demo. In a typical Dask workflow the steps omit `compute_backend` and therefore default to Kubernetes; the classifier sees k8s steps, marks the workflow `uses_kubernetes=True`, and the cap applies normally. The bypass triggers only when both conditions hold: `workflow.resources.dask` is set, **and** every step explicitly uses an external backend (HTCondor or Slurm). That combination is rarer than a vanilla Dask demo but it is valid configuration, and the workflows it admits are precisely the heaviest ones.

The framing that matters: a Dask-enabled workflow is not a "lighter" workflow that happens to escape a cap on a technicality. It is among the **heaviest** Kubernetes consumers REANA admits. Per-workflow defaults (`reana-server/reana_server/config.py:112-134` and the Helm chart) cap the requested Dask cluster at `REANA_DASK_CLUSTER_MAX_MEMORY_LIMIT = 16Gi` and `REANA_DASK_CLUSTER_MAX_NUMBER_OF_WORKERS = 20`. A perfectly honest user specifying `number_of_workers: 20, single_worker_memory: 800Mi` lands on 16 GiB of long-lived worker memory plus a scheduler pod plus the workflow-engine pod — all on Kubernetes, all for the wall-clock duration of the workflow. This is a much larger footprint than a typical batch workflow whose initial step requests a few GiB.

The cluster-wide `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` cap exists precisely to put a ceiling on this kind of resource commitment. After this PR, the specific class of Dask workflows described above is the one that escapes it. Direction of the bug matters: the cap stops applying along the axis where it is most needed.

The reason is a scope mistake in the classifier. `workflow_uses_kubernetes()` only inspects step-level `compute_backend`. Every estimator (`SerialComplexityEstimator`, `YadageComplexityEstimator`, `CWLComplexityEstimator`, `SnakemakeComplexityEstimator`) walks the workflow step tree and asks the same question — "is the step's compute backend `kubernetes`?" — and nothing else.

`reana-server/reana_server/complexity.py:88-99`:

```python
def workflow_uses_kubernetes(workflow_type: str, reana_yaml: dict) -> bool:
    """Return True if any step across the entire workflow runs on Kubernetes.
    ...
    """
    try:
        return _build_estimator(workflow_type, reana_yaml).uses_kubernetes()
    except Exception:
        return True
```

But Dask is declared at workflow-resource level, completely outside the step tree — `reana-server/reana_server/utils.py:813-817`:

```python
def workflow_uses_dask(reana_specification: Dict) -> bool:
    """Check whether a workflow specification requests a Dask service."""
    return bool(
        reana_specification.get("workflow", {}).get("resources", {}).get("dask")
    )
```

`DaskResourceManager` (`reana-workflow-controller/reana_workflow_controller/dask.py:51`, instantiated from `workflow_run_manager.py:407-410`) creates a `DaskCluster` CR with a scheduler pod plus N worker pods, optionally with an autoscaler.

So a workflow whose steps run on HTCondor or Slurm but whose `workflow.resources.dask` requests the maximum-allowed Dask cluster — e.g.:

```yaml
workflow:
  type: serial
  resources:
    dask:
      image: ...
      number_of_workers: 20
      single_worker_memory: 800Mi   # 20 × 800Mi = 16 GiB, the cluster max
  specification:
    steps:
      - compute_backend: htcondor
        commands: [...]
```

…classifies as external → the scheduler returns early from `check_concurrent_workflows_limit` (`reana-server/reana_server/scheduler.py:71`) → cap bypassed → ~16 GiB of Dask cluster memory plus a scheduler pod plus the engine pod spawn on the REANA cluster, unmetered, for each such workflow admitted. The cap that would otherwise limit how many of these accumulate at once is silent. The user does not need to "pretend" anything — straightforward use of documented Dask + documented external-backend features triggers this.

Suggested fix — short-circuit the classifier on any workflow-level resource that materialises as a Kubernetes object. Minimal patch:

```python
def workflow_uses_kubernetes(workflow_type: str, reana_yaml: dict) -> bool:
    # Workflow-level resources that materialise as k8s objects:
    if reana_yaml.get("workflow", {}).get("resources", {}).get("dask"):
        return True
    try:
        return _build_estimator(workflow_type, reana_yaml).uses_kubernetes()
    except Exception:
        return True
```

This is the narrowest patch that unblocks merge. Note that it still treats a Dask workflow as a single "k8s slot" even though it has a much larger k8s footprint than a regular workflow. A more honest accounting — a weighted cap, or a per-workflow k8s-resource profile that the scheduler can sum against the cap — is the natural follow-up and is the subject of separate findings on [the cap's semantics](https://github.com/reanahub/reana-server/pull/780#discussion_r3421401195) and on [the persisted data model](https://github.com/reanahub/reana-db/pull/267#discussion_r3421389305).

reana_server/scheduler.py
CameronMcClymont: Addressed by introducing per-backend concurrency limits rather than the weighted-cap variant.

reana_server/complexity.py
```diff
@@ -286,6 +352,14 @@ def parse_specification(self, initial_step):
 class CWLComplexityEstimator(ComplexityEstimatorBase):
     """REANA CWL workflow complexity estimation."""
 
+    @staticmethod
+    def _get_step_hints(step):
```
tiborsimko, line 356: **🤖 RWC663-04: Backend classification can disagree with what actually executes (CWL and Yadage) (MEDIUM)**

The new per-backend caps are only as good as the classifier that assigns workflows to backends, and for CWL and Yadage the classifier extracts a step's `compute_backend` with different semantics than both the validator and the workflow engines.

`reana_server/complexity.py:356`:

```python
@staticmethod
def _get_step_hints(step):
    """Return the last hint of a CWL step, or an empty dict if none.
    ...
    """
    return (step.get("hints") or [{}])[-1]
```

Four code paths, three distinct semantics for CWL:

- **Classifier (this PR)**: last hint only, `hints[-1]`.
- **Validator**: scans all hints for `class == "reana"` — reana-commons' `ComputeBackendValidatorCWL._get_reana_hints`.
- **Engine**: scans all hints, first hint containing the key wins — reana-workflow-engine-cwl's `_get_hint` in `cwl_reana.py`.
- The complexity estimator's job counting shares the classifier's last-hint assumption (as it did before this PR via `step.get("hints", [{}]).pop()`).

The canonical REANA representation is a `class: reana` hint, and nothing guarantees it is last. A step declared as:

```yaml
hints:
  - class: reana
    compute_backend: htcondorcern
  - class: ResourceRequirement
    ramMin: 4096
```

validates and executes as HTCondor, but the classifier reads `compute_backend` off the `ResourceRequirement` dict, gets nothing, and classifies the step as Kubernetes. The persisted `Workflow.compute_backends` is then wrong: the workflow is charged against the Kubernetes cap it does not use, and the HTCondor cap undercounts.

Yadage has the mirror-image problem: the classifier (`_get_stage_compute_backend`, `reana_server/complexity.py:195`) and the reana-commons validator both take the *first* `compute_backend` resource, while the engine's `_get_resources` in reana-workflow-engine-yadage's `externalbackend.py` iterates all resources with overwrite, so the *last* one wins. A stage declaring:

```yaml
resources:
  - compute_backend: htcondorcern
  - compute_backend: kubernetes
```

passes validation (first entry), is classified as external-only, and therefore bypasses both the Kubernetes concurrency cap and the Kubernetes memory check — yet the engine runs the job on Kubernetes. Nothing rejects duplicate declarations.

The new tests do not catch any of this because they exercise only the classifier, using a synthetic single-hint shape without `class: reana` (`hints: [{"compute_backend": "htcondor"}]`). On that shape the classifier and the engine happen to agree, while the validator's canonical REANA-hint semantics are never exercised.

On scope, to be fair: the CWL last-hint assumption predates this PR, so this is a pre-existing weakness rather than a regression — but this PR makes it load-bearing by persisting the classification and hanging admission control off it, so it is best fixed here.

Suggested fix — centralise "effective backend of a step" extraction so it matches the execution semantics: for CWL, scan all hints for the `class: reana` hint (matching the validator; the engine's first-key-wins rule is itself subtly different and worth reconciling in the same pass); for Yadage, either align on the engine's last-wins behaviour or, better, reject multiple `compute_backend` declarations per step at validation time in reana-commons. Add a realistic multi-hint CWL test (a `class: reana` hint followed by another hint) and a duplicate-resource Yadage test.

reana_server/status.py
```diff
@@ -653,25 +656,61 @@ def get_job_health(self):
...
+        binding = backends[bottleneck]
         return {
             "running": running,
             "queued": queued,
             "pending": pending,
-            "available": available,
-            "total": total,
-            "percentage": percentage,
-            "health": ClusterHealth.get_health_status(percentage),
+            "backends": backends,
```
tiborsimko, line 707: **🤖 RWC663-07: `/api/status` response grew fields the API contract does not describe (MEDIUM)**

`get_workflow_health()` now returns `backends`, `bottleneck` and `used` alongside the old keys (`reana_server/status.py:707-712`), and reana-ui PR #516 consumes them. But the OpenAPI docstring on the `/api/status` endpoint (`reana_server/rest/status.py`) still describes only the old single-counter `workflow` object (`available`/`running`/`queued`/`pending`/`percentage`/`health`/`sort`/`total`), and its `examples:` block still shows the old payload. Because `docs/openapi.json` is generated from that same docstring, the `--docs-openapi` check cannot detect the omission — the generated and committed specifications are both stale and still match — and the bundled copy in `reana-commons/reana_commons/openapi_specifications/reana_server.json`, the contract that bravado clients load at runtime, is equally stale. The headline feature of this PR set is thus invisible in the very API contract the new UI depends on.

Relatedly, nothing exercises the new response shape: `test_status.py` was not extended, so the per-backend tile assembly and the bottleneck selection — the only substantive new logic in `status.py` — have no test coverage.

Suggested fix — extend the docstring schema with the new `workflow` properties (`backends` as an object of per-resource `{available, health, percentage, total, used}` entries, plus `bottleneck` and `used`) and update the example payload to match, remembering that apispec emits schema fields in alphabetical order; regenerate `docs/openapi.json` via `./run-tests.sh --docs-openapi` and copy the result to `reana-commons/reana_commons/openapi_specifications/reana_server.json`. Add a `get_workflow_health()` unit test covering bottleneck selection (e.g. a saturated external backend against a roomy Kubernetes cap).

reana_server/status.py
```diff
@@ -653,25 +656,61 @@ def get_job_health(self):
...
 
+        active_counts = Workflow.count_active_per_backend()
+
+        # One tile per configured compute backend (so operators see headroom even
+        # at zero usage), plus any backend currently in use, plus the Dask cap.
+        resources = (
+            set(SUPPORTED_COMPUTE_BACKENDS) | {"kubernetes"} | set(active_counts)
+        )
+        resources.discard("dask")
+        resources = sorted(resources) + ["dask"]
```
tiborsimko, line 682: **🤖 RWC663-10: Dask capacity is reported even when Dask is disabled (LOW)**

`get_workflow_health()` unconditionally appends `"dask"` to the reported resources (`reana_server/status.py:681-682`):

```python
resources.discard("dask")
resources = sorted(resources) + ["dask"]
```

On a deployment with `DASK_ENABLED=false`, workflow specifications requesting Dask are rejected at validation (`reana_server/validation.py:192`), so no new Dask workflows can be admitted — yet the Workflows tile's backend breakdown and tooltips still show a Dask entry with, e.g., `0/10` slots. In steady state its count is 0, so it can never become the bottleneck, and the impact is confined to the breakdown misleading operators into thinking Dask capacity exists on a cluster where it does not. (The count is not *guaranteed* to be 0: if Dask is disabled while a Dask workflow is still active, that workflow legitimately shows up in the count until it finishes.)

Suggested fix — show the entry only when Dask is enabled or transitional usage is actually present, so nonexistent idle capacity is hidden without hiding real usage:

```python
from reana_server.config import DASK_ENABLED
...
resources.discard("dask")
resources = sorted(resources)
if DASK_ENABLED or active_counts.get("dask", 0):
    resources.append("dask")
```

reana-ui needs no matching change: it renders whatever `backends` contains.

reana_server/complexity.py
tiborsimko: The amended code resolves this finding: `publish_workflow_submission` now publishes `uses_dask` alongside the submission message, the scheduler applies the new orthogonal `REANA_MAX_CONCURRENT_DASK_WORKFLOWS` cap, and active Dask workflows are counted via the `Workflow.services` relation. `test_dask_on_external_workflow_submission_is_flagged` and `test_dask_workflows_are_capped` cover exactly the Dask-on-external scenario described here. Considering this addressed — thanks!

reana_server/complexity.py
CameronMcClymont: Very nice catch. Fixed now so both classifiers now resolve a step's backend the way the engine that runs it does, so the persisted `compute_backends` matches what actually executes. Also added new tests to cover the multi-hint CWL case and the duplicate-resource Yadage case in both orders, plus `tests/test_compute_backends.py` in reana-commons for the validator side.

reana_server/status.py
CameronMcClymont: Updated so the API contract now describes the response the UI actually consumes. `docs/openapi.json` was regenerated too and copied into `reana_commons/openapi_specifications/reana_server.json` in reanahub/reana-commons#541.

reana_server/status.py
CameronMcClymont: Fixed, using your suggested snippet.

PR #787, opened 2026-06-17
fix(config): enable Keycloak userinfo endpoint

tiborsimko:
Define the Keycloak userinfo endpoint switch expected by updated `invenio-oauthclient` when generic SSO login providers are configured.

The Keycloak helper reads `OAUTHCLIENT_KEYCLOAK_USER_INFO_FROM_ENDPOINT` directly during the OAuth callback. Without the setting, Keycloak SSO fails with a `KeyError` before account information can be loaded from the configured userinfo endpoint.

Closes #786


Sohn123 (approved):
LGTM :shipit: I checked it out locally and the fix worked as expected. Good to add the regression test 👍🏼

PR #791, opened 2026-06-23
fix(config): use a list for CSP default-src

Sohn123:
Allow Invenio-App to append its debug CSP source to default-src,
which it cannot do when the value is a string.

Normalise all CSP directive values to lists for a consistent
configuration shape.

Closes #790

### DIFF
--- a/reana_server/config.py
+++ b/reana_server/config.py
@@ -362,15 +362,15 @@ APP_DEFAULT_SECURE_HEADERS["strict_transport_security_include_subdomains"] = Tru
 APP_DEFAULT_SECURE_HEADERS["referrer_policy"] = "strict-origin-when-cross-origin"
 # NOTE: keep in sync with reana-ui nginx/reana-ui.conf content security policy headers
 APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {
-    "default-src": "'self'",
-    "script-src": "'self'",
+    "default-src": ["'self'"],
+    "script-src": ["'self'"],
     "style-src": ["'self'", "'unsafe-inline'"],
     "img-src": ["'self'", "data:"],
-    "font-src": "'self'",
-    "connect-src": "'self'",
-    "frame-ancestors": "'none'",
-    "object-src": "'none'",
-    "base-uri": "'self'",
+    "font-src": ["'self'"],
+    "connect-src": ["'self'"],
+    "frame-ancestors": ["'none'"],
+    "object-src": ["'none'"],
+    "base-uri": ["'self'"],
 }
 APP_DEFAULT_SECURE_HEADERS.update(
     json.loads(os.getenv("APP_DEFAULT_SECURE_HEADERS", "{}"))


michaelbuchar (dismissed):
Works great for me, let's merge this ASAP @tiborsimko!

reana_server/config.py
```diff
@@ -362,7 +362,7 @@ def _(x):
 APP_DEFAULT_SECURE_HEADERS["referrer_policy"] = "strict-origin-when-cross-origin"
 # NOTE: keep in sync with reana-ui nginx/reana-ui.conf content security policy headers
 APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {
-    "default-src": "'self'",
+    "default-src": ["'self'"],
```
tiborsimko, line 365: **🤖 Consider normalising CSP directive values to lists (NIT)**

`reana_server/config.py:364-374`

The `content_security_policy` dict mixes two value shapes — some directives are
lists, others bare scalar strings:

```python
"default-src": ["'self'"],                 # list
"script-src": "'self'",                    # string
"style-src": ["'self'", "'unsafe-inline'"],# list
"font-src": "'self'",                      # string
...
```

This inconsistency pre-dates this PR — the scalar directives were already
strings on master — so the change here does not introduce it, and it is not a
correctness bug. With the pinned flask-talisman 0.8.1 and invenio-app 3.0.0,
`_parse_policy` (`talisman.py:318`) only joins non-string values, so scalar
strings render verbatim and both forms emit the identical header; only
`default-src` is *required* to be a list, because invenio-app's DEBUG path
calls `csp["default-src"].append(...)` (`invenio_app/ext.py:142`). The rest are
equivalent under these versions, and the change is correct and complete as-is.

The remaining value is readability: uniform lists match invenio-app's own
default style (`{"default-src": ["'self'"], "object-src": ["'none'"]}`) and
pre-empt the same `AttributeError` footgun should invenio ever mutate another
directive.

Suggested fix — normalise every directive to a one-element list and squash it
into this commit (cosmetic only; the emitted header is byte-identical):

```python
APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {
    "default-src": ["'self'"],
    "script-src": ["'self'"],
    "style-src": ["'self'", "'unsafe-inline'"],
    "img-src": ["'self'", "data:"],
    "font-src": ["'self'"],
    "connect-src": ["'self'"],
    "frame-ancestors": ["'none'"],
    "object-src": ["'none'"],
    "base-uri": ["'self'"],
}
```

Commit message — if you reword while squashing, a headline that states the
change with the rationale in the body reads better than "match invenio
contract" (keep body lines ≤72 chars, and retain the ` (#791)` suffix):

```
fix(config): use a list for CSP default-src (#791)

Allow Invenio-App to append its debug CSP source to default-src,
which it cannot do when the value is a string.

Normalise all CSP directive values to lists for a consistent
configuration shape.
```

tiborsimko: Amended during merge.

PR #792, opened 2026-06-24
feat(launch): support launching workflows from repo folders

michaelbuchar:


reana_server/fetcher.py
```diff
@@ -229,6 +261,63 @@ def __init__(
...
+            f"refs/tags/{git_ref}",
+        ]
+        for ref_candidate in ref_candidates:
+            try:
+                repository.git.rev_parse("--verify", ref_candidate)
+                return True
+            except Exception:
+                continue
+
+        if re.fullmatch(r"[0-9a-f]{7,40}", git_ref):
```
CameronMcClymont: Should use the `_looks_like_git_sha` helper here?

reana_server/fetcher.py
```diff
@@ -229,6 +261,63 @@ def __init__(
...
+        remote_name = repository.remote().name
+        ref_candidates = [
+            f"refs/remotes/{remote_name}/{git_ref}",
+            f"refs/tags/{git_ref}",
+        ]
+        for ref_candidate in ref_candidates:
+            try:
+                repository.git.rev_parse("--verify", ref_candidate)
+                return True
+            except Exception:
```
CameronMcClymont: Could this exception scope be narrowed? Or at least a comment explaining what kind of exception(s) is/are expected.

reana_server/fetcher.py
```diff
@@ -229,6 +261,63 @@ def __init__(
...
+                continue
+
+        if re.fullmatch(r"[0-9a-f]{7,40}", git_ref):
+            try:
+                repository.commit(git_ref)
+                return True
+            except Exception:
```
CameronMcClymont: Could this exception scope be narrowed? Or at least a comment explaining what kind of exception(s) is/are expected.

tests/test_fetcher.py
```diff
@@ -161,6 +162,169 @@ def create_git_repository(repo_path, files, idx):
     assert os.path.isfile(expected_path)
 
 
+@pytest.mark.parametrize(
```
CameronMcClymont, line 161: Could add a test that expects a failure when attempting path-traversal (e.g. `tree/main/../../etc`)

reana_server/fetcher.py
```diff
@@ -229,6 +261,63 @@ def __init__(
...
+
+        return False
+
+    @staticmethod
+    def _looks_like_git_sha(git_ref: str) -> bool:
+        """Check whether the given git ref looks like a commit SHA."""
+        return bool(re.fullmatch(r"[0-9a-f]{7,40}", git_ref))
+
+    @classmethod
+    def _resolve_tree_path(
```
CameronMcClymont, line 320: After these changes, the ref resolution requires the ref to exist locally, which can cause issues with tags. For example, consider a repo with these commits:

main: A -> B -> C (HEAD)
tags: v1.0.0 -> A

Then say you try to launch `https://github.com/user/repo/tree/v1.0.0`

The repo clone uses depth=1, so it fetches only the branch tips (C, plus other branch tips) plus tags pointing at fetched commits. v1.0.0 points at A, which isn't fetched, so refs/tags/v1.0.0 is absent from the local clone.

Before this PR, `git_ref` = "v1.0.0" was passed straight to `repository.remote().fetch("v1.0.0", depth=1)` so git resolves the tag on the remote and fetches A which works.

But after this PR, `_resolve_tree_path` calls `_git_ref_exists(repo, "v1.0.0")`. But there is no local refs/tags/v1.0.0, and the name isn't a hex SHA, so this eventually raises 'Cannot checkout the given Git reference "v1.0.0"'
  
In summary, launching any release tag that isn't a branch HEAD now breaks. This is probably important to have working in case people want to e.g. "run v1.2.3 release of this workflow".

CameronMcClymont: So I think you would just need to change the `if` statement at line 309:

```
remote_refs = cls._remote_refs(repository)
...
if (
    f"refs/heads/{git_ref}" in remote_refs
    or f"refs/tags/{git_ref}" in remote_refs
    or cls._git_ref_exists(repository, git_ref)
):
```

CameronMcClymont (changes requested):

reana_server/fetcher.py
michaelbuchar: Good catch, thanks. I updated _git_ref_exists() to call _looks_like_git_sha() instead of duplicating the regex.

reana_server/fetcher.py
michaelbuchar: Agreed. I narrowed this to GitCommandError, since this branch only expects rev-parse --verify to fail when the candidate ref is absent.

reana_server/fetcher.py
michaelbuchar: Agreed. I narrowed this to GitCommandError, which is what GitPython raises when rev-parse --verify cannot resolve the candidate ref.

tests/test_fetcher.py
michaelbuchar: Added, thanks

reana_server/fetcher.py
michaelbuchar: Thanks for the detailed catch.

reana_server/fetcher.py
```diff
@@ -229,6 +262,84 @@ def __init__(
...
+    def _looks_like_git_sha(git_ref: str) -> bool:
+        """Check whether the given git ref looks like a commit SHA."""
+        return bool(re.fullmatch(r"[0-9a-f]{7,40}", git_ref))
+
+    @staticmethod
+    def _remote_refs(repository: Repo) -> set[str]:
+        """Get the remote heads and tags available in the repository."""
+        with repository.git.custom_environment(GIT_TERMINAL_PROMPT="0"):
+            remote_refs = repository.git.ls_remote(
+                "--heads", "--tags", repository.remote().url
```
CameronMcClymont: 🤖 The git ls-remote network call in _remote_refs is not wrapped, so failures become HTTP 500 instead of 400. Every other git
  operation in this module (clone, fetch, checkout) converts errors to REANAFetcherError, which launch() maps to a 400 with a useful message. If ls-remote
  fails between the successful clone and resolution (transient network error, rate limit, repo made private), the raw GitCommandError falls into launch()'s
  catch-all and returns a generic 500.

reana_server/fetcher.py
```diff
@@ -229,6 +262,84 @@ def __init__(
         """
         super().__init__(parsed_url, output_dir, spec)
         self._git_ref = git_ref
+        self._tree_path = tree_path
+
+    @staticmethod
+    def _looks_like_git_sha(git_ref: str) -> bool:
+        """Check whether the given git ref looks like a commit SHA."""
+        return bool(re.fullmatch(r"[0-9a-f]{7,40}", git_ref))
```
CameronMcClymont: 🤖 _looks_like_git_sha's lowercase-only regex [0-9a-f]{7,40} rejects uppercase/mixed-case SHAs that worked before this PR.
  Verified empirically: git accepts uppercase hex in fetch/checkout (it normalizes the oid), so pre-PR a tree/<UPPERCASE-SHA> URL launched fine; post-PR no
  resolution branch matches and the request fails with "Cannot checkout the given Git reference". Low likelihood (UI-generated URLs are lowercase) but a
  one-character fix (re.IGNORECASE).

reana_server/fetcher.py
```diff
@@ -229,6 +262,84 @@ def __init__(
...
+        return {
+            ref
+            for line in remote_refs.splitlines()
+            for ref in [line.split(None, 1)[1]]
+            if not ref.endswith("^{}")
+        }
+
+    @classmethod
+    def _git_ref_exists(cls, repository: Repo, git_ref: str) -> bool:
+        """Check whether the given git ref exists in the repository."""
```
CameronMcClymont, line 309: 🤖 Parts of _git_ref_exists are provably dead, and each dead check costs a git subprocess per candidate prefix in the 
  resolution loop. The refs/tags/<ref> rev-parse can never succeed when ls-remote membership already failed (ls-remote lists all remote tags), and the
  in-loop SHA branch returns exactly what the post-loop fallback returns. An invalid K-segment tree path spawns ~2·K dead rev-parse processes on the request
  path. Caveat from verification: the refs/remotes/.../HEAD rev-parse is load-bearing for /tree/HEAD URLs, and the ls-remote call itself is load-bearing for
  tags on non-tip commits — so trim the dead checks rather than the mechanism, and optionally try local refs first to skip the extra network round-trip for
  the common tree/<branch> case.

reana_server/fetcher.py
```diff
@@ -182,26 +185,55 @@ def _is_path_inside_output_dir(self, path: str) -> bool:
...
+        Otherwise, the file will be searched in the workflow root directory. This
+        method should be called after ``fetch``.
 
         :returns: Path of the workflow specification file.
         """
         if self._spec:
-            spec_path = os.path.abspath(os.path.join(self._output_dir, self._spec))
-            if not self._is_path_inside_output_dir(spec_path):
+            workflow_root = self.workflow_root_path()
+            spec_path = os.path.abspath(os.path.join(workflow_root, self._spec))
```
CameronMcClymont, line 240: 🤖 workflow_spec_path inlines a second copy of the realpath+commonpath containment check instead of generalizing 
  _is_path_inside_output_dir. This is a security-sensitive traversal guard that now exists in two hand-rolled copies; a future hardening fix (e.g.
  commonpath's ValueError edge cases) applied to the helper would silently miss the inline copy guarding the user-supplied spec path. A _is_path_inside(path,
  base) helper covers both.

reana_server/fetcher.py
```diff
@@ -255,6 +371,8 @@ def fetch(self) -> None:
                     f'Cannot checkout the given Git reference "{self._git_ref}"'
                 )
 
+        # Validate the selected workflow root after checkout.
+        self.workflow_root_path()
```
CameronMcClymont: 🤖 fetch() calls self.workflow_root_path() purely for side-effect validation and discards the result, and every later caller 
  re-runs the full abspath/realpath/commonpath/isdir sequence (3+ times per launch). Computing it once into self._workflow_root — after the .git rmtree,
  which matters because the current pre-rmtree validation misses roots under .git that the later re-validation catches — makes workflow_root_path() a plain
  accessor and removes the dead-looking bare call.

reana_server/fetcher.py
```diff
@@ -427,7 +546,11 @@ def _get_github_fetcher(
         return WorkflowFetcherZip(parsed_url, output_dir, spec, workflow_name)
     else:
         repository_url = ParsedUrl(f"https://github.com/{username}/{repository}.git")
-        return WorkflowFetcherGit(repository_url, output_dir, git_ref, spec)
+        if tree_path:
```
CameronMcClymont: 🤖 (and fetcher.py:596) — the if tree_path: / else branches in _get_github_fetcher and _get_gitlab_fetcher both construct the same 
  WorkflowFetcherGit. Since tree_path=None is the constructor default, each collapses to a single WorkflowFetcherGit(repository_url, output_dir, spec=spec, 
  tree_path=tree_path) — four call sites where two would do.

tests/test_fetcher.py
```diff
@@ -161,6 +162,238 @@ def create_git_repository(repo_path, files, idx):
...
+    ],
+)
+def test_fetcher_git_tree_path_with_workflow_directory(
+    tree_path, expected_root, expected_workflow_name, tmp_path
+):
+    """Test fetching workflows from repository folder URLs."""
+    repo_dir = os.path.join(tmp_path, "repo")
+    output_dir = os.path.join(tmp_path, "output")
+
+    repository = Repo.init(repo_dir, initial_branch="main")
```
CameronMcClymont: 🤖 The new tree-path tests hand-roll ~100 lines of Repo.init + makedirs + write + index.add + commit boilerplate, even though 
  this PR extended create_git_repository with os.makedirs specifically to support nested paths. Hoisting that helper to module level (it's currently nested
  inside test_fetcher_git) would shrink each new test to its distinctive assertions.

CameronMcClymont (changes requested):
Let Fable loose on this one. Nothing severe I would say, Fable seems to be quite nit-picky.

Also found this existing stale FIXME comment (I think launch.py:276): Snakemake chdir leak + subfolder cleanup poisoning the worker — verified empirically on both snakemake 7.x and 9.x paths: cwd is always restored; the FIXME in launch.py is stale (2022-era).

And one more pre-existing issue: on the snakemake v8+ path, launch.py passes workspace_path as a str, and
  snakemake 9.22.0's WorkdirHandler calls .exists() on it, raising AttributeError → "Snakemake specification is invalid" before any chdir — meaning
  allowlisted Snakemake launches likely fail outright with the pinned snakemake, independent of this PR.

reana_server/fetcher.py
```diff
@@ -219,6 +267,7 @@ def __init__(
         output_dir: str,
         git_ref: Optional[str] = None,
         spec: Optional[str] = None,
+        tree_path: Optional[str] = None,
```
CameronMcClymont: New `tree_path` param is missing a docstring.

reana_server/rest/launch.py
```diff
@@ -193,16 +195,21 @@ def launch(user, url, name="", parameters="{}", specification=None):
 
         # FIXME: locking will not be needed when the loading and validation of
         # specifications will be done inside an external sandbox
+        reana_spec_workspace_path = (
+            Path(workflow_root) if workflow_type == "snakemake" else workflow_root
+        )
```
CameronMcClymont: Just a note that the `reana_spec_workspace_path` conditional will be able to collapse back to `workspace_path=workflow_root` after this reana-commons fix: https://github.com/reanahub/reana-commons/pull/555

(although it's also harmless to keep even after 555 is merged)

reana_server/rest/workflows.py
```diff
@@ -607,8 +608,13 @@ def create_workflow(user):  # noqa
                 )
             elif workflow.type_ in ["cwl", "snakemake"]:
                 reana_yaml_path = os.path.join(workflow.workspace_path, "reana.yaml")
+                workspace_path = (
+                    Path(workflow.workspace_path)
+                    if workflow.type_ == "snakemake"
+                    else workflow.workspace_path
+                )
```
CameronMcClymont: Just a note that the `workspace_path` conditional will be able to collapse back to `workspace_path=workflow.workspace_path` after this reana-commons fix: https://github.com/reanahub/reana-commons/pull/555

(although it's also harmless to keep even after 555 is merged)

reana_server/fetcher.py
```diff
@@ -172,36 +176,80 @@ def _discover_workflow_specs(self, dir: Optional[str] = None) -> List[str]:
...
+        method should be called after ``fetch``.
 
         :returns: Path of the workflow specification file.
         """
         if self._spec:
-            spec_path = os.path.abspath(os.path.join(self._output_dir, self._spec))
-            if not self._is_path_inside_output_dir(spec_path):
+            workflow_root = self.workflow_root_path()
+            spec_path = os.path.abspath(os.path.join(workflow_root, self._spec))
+            if not self._is_path_inside(spec_path, workflow_root):
```
CameronMcClymont, line 276: Could add a regression test for this spec-path traversal guard (e.g. `spec="../reana.yaml"` or `spec="../../etc/passwd.yaml"` should both raise `Invalid path to the workflow specification`). The existing traversal test only covers the workflow root guard, not the spec one.

CameronMcClymont (changes requested):

CameronMcClymont:
Also needs a rebase.

reana_server/fetcher.py
michaelbuchar: Fixed

reana_server/rest/launch.py
michaelbuchar: Kept the conditional for compatibility

reana_server/rest/workflows.py
michaelbuchar: Yes, I left the conditional in place for compatibility

reana_server/fetcher.py
michaelbuchar: Fixed

michaelbuchar:
All issues should be solved and this PR is ready for re-review!

reana_server/fetcher.py
```diff
@@ -235,16 +282,119 @@ def __init__(
...
+        tree_path: str,
+        repository_path: Optional[str] = None,
+    ) -> tuple[str, Optional[str]]:
+        """Resolve a GitHub/GitLab tree path into git ref and workflow subdirectory.
+
+        Tree URLs have the form ``tree/<git_ref>[/path/to/workflow]``. Since Git refs can
+        themselves contain slashes, choose the longest path prefix that resolves to a
+        known git ref and treat the remaining suffix as a workflow subdirectory.
+        """
+        path_parts = tree_path.strip("/").split("/")
```
tiborsimko: **🤖 PR792-01: Percent-encoded tree paths are never decoded or validated segment-by-segment (MEDIUM)**

`ParsedUrl` keeps the URL-encoded path, and `_match_url()` returns it without
decoding. Consequently, a normal provider URL such as
`tree/main/workflows/my%20analysis` selects the literal directory
`workflows/my%20analysis`; the extracted repository contains
`workflows/my analysis`, so the launch fails with "Cannot find the given
workflow directory". The same applies to folder names containing spaces,
non-ASCII characters, `#`, and other characters that providers percent-encode.

Decoding the complete path without validating its segments would introduce a
different problem: encoded separators could become path separators, and the
current `tree/main//workflows` input already produces an absolute
`workflow_path`. The realpath containment check catches the latter today, but
malformed paths should be rejected before joining them to the output
directory.

Suggested fix — split the encoded tree path on literal `/`, strictly decode
each segment exactly once, and reject decoded segments that are empty, `.`,
`..`, contain `/` or `\`, or contain NUL. Rejoin only validated segments and
retain the existing realpath/commonpath containment check as a backstop. Add
production-shaped GitHub and GitLab tests covering a space or non-ASCII folder,
encoded traversal and separators, NUL, and a double slash.

reana_server/fetcher.py
```diff
@@ -235,16 +282,119 @@ def __init__(
...
+    @staticmethod
+    def _remote_refs(repository_url: str) -> set[str]:
+        """Get the remote heads and tags available in the repository."""
+        environment = dict(os.environ)
+        environment["GIT_TERMINAL_PROMPT"] = "0"
+        try:
+            process = subprocess.run(
+                ["git", "ls-remote", "--heads", "--tags", repository_url],
+                env=environment,
+                stdout=subprocess.PIPE,
```
tiborsimko: **🤖 PR792-02: Git reference discovery buffers unbounded remote output (MEDIUM)**

`_remote_refs()` runs `git ls-remote --heads --tags` with
`stdout=subprocess.PIPE`, then materialises the complete remote advertisement
as a string, a `splitlines()` list, and finally a set. The timeout limits how
long Git may run, but it does not limit how much output a user-selected
repository may return.

This bypasses the resource-bounding invariant enforced elsewhere in the
fetcher for remote files, archive entries, path depth, directory breadth, and
clone size. A repository with a very large number of refs can therefore cause
substantial API-process memory use before any bounded archive is downloaded.
The launch endpoint requires an authenticated session and the resolver only
contacts allowlisted provider hosts, so this is a departure from the module's
bounding invariant rather than an unauthenticated denial-of-service path.

Suggested fix — ask `ls-remote` only for the bounded set of candidate
`refs/heads/<prefix>` and `refs/tags/<prefix>` values derived from the tree
path, or stream its output with explicit byte and reference-count limits. Keep
the fully qualified `refs/heads/` and `refs/tags/` prefixes and separate
options from operands with `--` if the candidate-query approach is used, so no
user-derived argument can be parsed as an option. Preserve the current timeout
and `REANAFetcherError` translation, and add a test proving that excessive ref
output is rejected without being fully materialised.

reana_server/fetcher.py
```diff
@@ -262,6 +412,11 @@ def fetch(self) -> None:
                 "URL is correct and that the repository is publicly accessible."
             )
 
+        if self._tree_path:
```
tiborsimko: **🤖 PR792-03: Dead generic-Git tree machinery obscures missing production-path coverage (LOW)**

`get_fetcher()` sends GitHub and allowlisted GitLab URLs to
`WorkflowFetcherZip`; its only production construction of
`WorkflowFetcherGit` never supplies `tree_path`. The `if self._tree_path`
branch in `WorkflowFetcherGit.fetch()`, the `repository_path` argument to
`_resolve_tree_path()`, and the `_tree_path` name-generation branch therefore
cannot be reached through the public fetcher factory.

`_git_ref_exists()` is redundant even in that direct-construction path: it
only checks `refs/remotes/origin/<ref>`, while the preceding remote-ref query
already reports the corresponding branch as `refs/heads/<ref>`, and `HEAD` is
handled separately. Several new tests directly construct this non-production
configuration, while the shipping factory, archive extraction, selected root,
and specification lookup are only tested as separate halves.

Two smaller dead-code residues point in the same direction. The post-loop
`path_parts` truthiness check cannot fail because the single-segment case has
already returned, and `extract_streamed_zip_response()` assigns
`_workflow_root` from a fetcher constructed without `workflow_path`, so that
private assignment always writes the existing output directory back unchanged.

Suggested fix — remove the unused generic-Git tree-path plumbing and the
redundant local-ref fallback unless a supported generic-Git folder URL is
defined. Remove the redundant truthiness guard and streamed-fetch assignment,
and re-home the useful resolver and containment cases in tests that enter
through `get_fetcher()` and use a mocked provider archive download. Include a
slash-containing tag with a workflow folder so the `refs/tags/` resolution
path is exercised in the production configuration.

reana_server/fetcher.py
```diff
@@ -235,16 +282,119 @@ def __init__(
...
+
+        Tree URLs have the form ``tree/<git_ref>[/path/to/workflow]``. Since Git refs can
+        themselves contain slashes, choose the longest path prefix that resolves to a
+        known git ref and treat the remaining suffix as a workflow subdirectory.
+        """
+        path_parts = tree_path.strip("/").split("/")
+        if len(path_parts) == 1:
+            return path_parts[0], None
+
+        remote_refs = cls._remote_refs(repository_url)
```
tiborsimko: **🤖 PR792-04: Commit-SHA folder URLs unnecessarily depend on ref enumeration (LOW)**

For every multi-segment tree path, `_resolve_tree_path()` calls
`_remote_refs()` before considering the commit-SHA fallback. A URL such as
`tree/<full-commit-sha>/workflows/example` therefore fails if smart-Git ref
discovery is temporarily unavailable, even though its split is unambiguous
unless the same hexadecimal prefix is also an advertised branch or tag, and
the provider archive endpoint may still be able to fetch that commit.

Advertised refs should continue to take precedence when lookup succeeds,
because a hexadecimal-looking branch or tag can exist. The avoidable problem
is making a successful ref advertisement a hard prerequisite for every
canonical commit-folder URL.

This resolution inside `get_fetcher()` also changes a formerly local URL
selection helper into a network operation that can raise `REANAFetcherError`.
The launch endpoint maps that error correctly, but the helper's contract does
not mention the network dependency, and module-level provider functions call a
private classmethod on a different fetcher class to perform the work.

Suggested fix — preserve branch/tag precedence after a successful lookup, but
allow a well-defined SHA fallback when ref discovery fails, at least for a
canonical full-length commit SHA. Document how abbreviated SHAs and
hexadecimal-looking refs are disambiguated and that fetcher selection may
resolve remote refs. Prefer a module-level resolver shared by the provider
factories, and test the fallback with `_remote_refs()` failing while the
archive fetch succeeds.

reana_server/fetcher.py
```diff
@@ -188,36 +191,80 @@ def _discover_workflow_specs(self, dir: Optional[str] = None) -> List[str]:
...
+        """
+        return self._workflow_root
 
     def workflow_spec_path(self) -> str:
         """Get the path of the workflow specification file.
 
         If the path to the specification file was provided, only that will be used to
-        find the workflow specification. Otherwise, the file will be searched in the
-        output directory. This method should be called after ``fetch``.
+        find the workflow specification inside the selected workflow root directory.
```
tiborsimko, line 267: **🤖 PR792-05: Launch API docs omit folder-relative specification semantics (NIT)**

The launch API still describes `url` as the location where the REANA
specification file is hosted and `specification` only as a path to that file.
With this change, a GitHub or GitLab tree URL can select a repository folder,
and an explicit `specification` path is resolved relative to that selected
folder rather than the repository root. This is observable API behaviour that
clients and users cannot infer from the current contract.

Suggested fix — update the inline OpenAPI docstring on `launch()` in
`reana_server/rest/launch.py` to document repository-folder tree URLs and
state explicitly that `specification` is relative to the selected folder.
Regenerate `docs/openapi.json` and synchronise
`reana_commons/openapi_specifications/reana-server.json` in reana-commons.

tiborsimko (commented):
**🤖 PR792-06: Modified files omit the 2026 copyright year (NIT)**

Both modified files still contain only `Copyright (C) 2022 CERN`, although
their Git histories now contain modifications in 2022 and 2026. The project
copyright convention requires every modification year to remain listed.

Suggested fix — update the headers in `reana_server/fetcher.py` and
`tests/test_fetcher.py` to `Copyright (C) 2022, 2026 CERN`.


**🤖 PR792-07: Non-obvious launch-resolution design is undocumented in the commit and PR (NIT)**

The pull request description and the commit body are both empty, although the
change introduces longest-prefix ref resolution, a commit-SHA fallback, an
additional smart-Git network request, and folder-relative specification
semantics. These choices and their ambiguity trade-offs are not apparent from
the headline alone.

Recent substantive reana-server changes normally record their motivation and
design in the commit body. This is not required for changelog generation, but
without it future maintainers must reconstruct the resolution contract from
the implementation and review history.

Suggested fix — add a concise PR description explaining the accepted URL
forms, longest-prefix rule, SHA fallback, folder-relative specification scope,
and new `ls-remote` dependency. Add the same durable rationale to the commit
body before merge, keeping body lines within 72 characters.

reana_server/rest/launch.py
tiborsimko: **🤖 Follow-up (2026-08-23): obsolete after the 19 August rebase**

This PR no longer modifies `reana_server/rest/launch.py`, and the current head
does not contain this conditional. After the rebase onto the server-side
validation changes, launch stages the selected specification through the
sandboxed validator, and source scoping derives from the specification's
containing directory. No compatibility conditional remains in this code path,
so no further change is needed for this discussion.

reana_server/rest/workflows.py
tiborsimko: **🤖 Follow-up (2026-08-23): obsolete after the 19 August rebase**

This PR no longer modifies `reana_server/rest/workflows.py`, and the current
head does not contain this conditional. After the rebase onto the server-side
validation changes, workflow creation no longer loads the specification
through this conditional `workspace_path` branch; loading and source scoping
are handled by the staged validation path instead. No further change is needed
for this discussion.

reana_server/fetcher.py
michaelbuchar: Fixed!

reana_server/fetcher.py
michaelbuchar: Fixed!

reana_server/fetcher.py
michaelbuchar: Fixed

reana_server/fetcher.py
michaelbuchar: Fixed

reana_server/fetcher.py
michaelbuchar: Fixed!

reana_server/fetcher.py
```diff
@@ -722,6 +777,162 @@ def _match_url(parsed_url: ParsedUrl, rules: Sequence[str]) -> Mapping[str, Any]
...
+    subdirectory. Abbreviated SHAs are accepted only after successful remote ref
+    discovery proves no branch or tag takes precedence. Full-length SHAs can
+    fall back when remote ref discovery fails, leaving the archive endpoint to
+    resolve the commit.
+    """
+    path_parts = _decode_provider_tree_path(tree_path)
+    decoded_tree_path = "/".join(path_parts)
+    if len(path_parts) == 1:
+        return path_parts[0], None, decoded_tree_path
+    if path_parts[0] == "HEAD":
```
tiborsimko: **🤖 PR792-08: `HEAD` fast path bypasses longer branch and tag matches (LOW)**

`_resolve_provider_tree_path()` documents that ambiguous tree paths are
resolved by choosing the longest prefix advertised as a branch or tag. The
special case here bypasses that resolution whenever the first segment is
`HEAD`:

```python
if path_parts[0] == "HEAD":
    return "HEAD", "/".join(path_parts[1:]), decoded_tree_path
```

Git permits longer refs such as `refs/heads/HEAD/foo` and
`refs/tags/HEAD/foo`. For a URL such as
`tree/HEAD/foo/workflows/example`, an advertised `HEAD/foo` ref should
therefore win under the longest-prefix rule. Instead, the current code always
selects symbolic `HEAD` and treats `foo/workflows/example` as the folder. This
can either fail to find the workflow or launch the similarly named folder from
the repository's default branch rather than the requested ref. The earlier
iteration of this pull request considered longer advertised refs before
falling back to symbolic `HEAD`, so the early return introduces this edge-case
regression.

Suggested fix — let the existing longest-prefix loop consult the longer
`HEAD/...` candidates that `_remote_ref_candidates()` already emits. Remove
the premature return and restore symbolic `HEAD` as a post-loop fallback when
no longer advertised ref matches. Preserve the same fallback when remote
discovery fails; otherwise ordinary `tree/HEAD/<folder>` URLs acquire a new
network-availability dependency. `_remote_ref_candidates()` should continue
to omit only the exact `HEAD` candidate. Add tests for both
`tree/HEAD/<folder>` and an advertised `HEAD/foo` branch or tag followed by a
workflow folder, including a remote-discovery failure for the symbolic case.

reana_server/fetcher.py
tiborsimko: **🤖 PR792-02 follow-up (2026-08-25): candidate filtering does not bound `ls-remote` output (MEDIUM remains)**

Thanks for narrowing discovery to the candidate prefixes and filtering the
result back to exact ref names. Rejecting glob-bearing candidates prevents
user input from broadening the local display pattern, but the resource-bound
part of the original finding remains only partially addressed.

`git ls-remote` treats a fully qualified operand as a tail pattern. For
example, querying one candidate:

```console
git ls-remote --heads --tags -- <repo> refs/heads/main
```

also returns refs such as `refs/heads/x1/refs/heads/main` and
`refs/tags/t1/refs/heads/main`. A repository can advertise a large number of
such suffix matches. The exact post-filter correctly discards these lines, so
ref resolution is not affected, but it runs only after `subprocess.run()` has
buffered the complete stdout string and `splitlines()` has materialised it
again.

Under Git protocol v2 the candidate operands are not sent as exact ref
prefixes either. A packet trace for the command above sends only:

```text
ref-prefix refs/tags/
ref-prefix refs/heads/
```

These prefixes are derived client-side from the `--heads` and `--tags` flags,
so this behaviour does not depend on whether the repository is reached over
HTTPS or another transport. The remote therefore advertises all branches and
tags, and Git applies the tail patterns locally. The timeout bounds duration,
but neither the remote transfer nor the API process's buffered matching output
is size-bounded. The authenticated endpoint and provider allowlist continue
to limit exposure, so the original MEDIUM severity remains appropriate.

The current code comment says discarding glob characters keeps remote output
bounded. The filtering prevents pattern expansion, but it does not establish
that output bound.

Suggested fix — retain the glob guard and exact post-filter, but consume Git's
stdout incrementally with explicit byte and/or line limits, terminating and
reaping the subprocess if a limit is exceeded. The existing bounded-Git code
provides a useful pattern for deadline and process cleanup, although this path
needs its own stdout bound and cannot prevent the protocol's complete
heads/tags advertisement. Add a regression test with many suffix-matching
refs, and narrow the code comment to the guarantee the glob filter actually
provides.

tiborsimko (commented):
**🤖 PR792-09: Commit headline obscures the repository-subfolder scope (NIT)**

The current commit headline is:

```text
feat(launch): support launching workflows from repo folders (#792)
```

REANA already supports launching workflows from repositories, so “support
launching workflows” can be read as describing existing functionality. “Repo
folders” is also ambiguous about the important distinction introduced here:
a GitHub or GitLab tree URL can now select a repository **subfolder** as the
workflow root. This lets one repository contain several independently
launchable analyses, each with its own REANA specification and supporting
files. The PR title currently carries the same ambiguity, although the PR
description itself already explains the accepted `/tree/<ref>[/<folder>]`
forms clearly.

Suggested fix — make the subfolder scope explicit in the commit headline:

```text
feat(launch): launch workflows from repository subfolders (#792)
```

Update the PR title separately to the same text without `(#792)`; amending and
force-pushing the commit does not propagate its new headline to the GitHub PR
title. Replace the current commit body as part of the same amendment so that
it also drops the incorrect claim that filtering candidate patterns bounds
`ls-remote` output. For example:

```text
Treat a selected GitHub or GitLab repository subfolder as the workflow
root. Previously, provider tree URLs used the repository root and could
not select one of several analyses stored in the same repository, each
with its own REANA specification and supporting files.

Decode and validate every tree-path segment before use. Since Git refs
may contain slashes, query matching remote branch and tag refs and
select the longest matching prefix. Treat the remaining suffix as the
workflow folder and resolve the specification relative to it.

Preserve full commit SHA and symbolic HEAD fallbacks when remote ref
discovery is unavailable.
```

This example describes the intended final behaviour after the open
ref-resolution findings are addressed. In particular, its final paragraph
becomes accurate once symbolic `HEAD` is restored as a fallback after remote
discovery rather than bypassing discovery entirely. Finalise the commit body
together with those code changes so that it describes the implementation that
will actually be merged.

michaelbuchar:
Re-pushed with the new changes so feel free to re-review @tiborsimko !

reana_server/fetcher.py
```diff
@@ -722,6 +781,232 @@ def _match_url(parsed_url: ParsedUrl, rules: Sequence[str]) -> Mapping[str, Any]
...
+        candidates = (f"refs/heads/{git_ref}", f"refs/tags/{git_ref}")
+        ref_candidates.extend(
+            candidate
+            for candidate in candidates
+            if not _GIT_LS_REMOTE_GLOB_CHARACTERS.intersection(candidate)
+        )
+    return tuple(dict.fromkeys(ref_candidates))
+
+
+def _run_bounded_git_output(
```
tiborsimko, line 844: **🤖 PR792-10: Remote-ref success path is only tested across a mock boundary (LOW)**

The resource-limit rejection path exercises a real child process, but the
successful remote-ref path is split across a mock boundary.
`test_remote_refs_queries_only_candidate_refs()` replaces
`_run_bounded_git_output()` with a byte literal, while
`test_remote_refs_rejects_excessive_suffix_matches()` exercises only the
reader's limit-exceeded path. Consequently, no test runs the bounded reader to
EOF and then lets `_remote_refs()` parse its output. The EOF/unregister branch,
final timed wait, successful byte return, and non-zero child exit translation
remain uncovered (`reana_server/fetcher.py` lines 875–876, 888–892, and
907–911 at head `3c8b1b6`).

The existing tests do assert the constructed argv and the exact-match byte
filter separately, so the current implementation is not unverified wholesale.
The missing piece is their composition against Git's real output contract. The
post-discovery abbreviated-SHA acceptance branch is also untested; the current
abbreviated-SHA test covers only discovery failure (line 1001).

Suggested fix — create a real local repository with the module-level
`create_git_repository()` helper and pass its path to `_remote_refs()` or
`_resolve_provider_tree_path()`, since `git ls-remote` accepts local paths.
Assert a successful multi-candidate lookup, the abbreviated-SHA fallback after
successful discovery with no match, and the `REANAFetcherError` translation
for a missing repository path.

reana_server/fetcher.py
```diff
@@ -722,6 +781,232 @@ def _match_url(parsed_url: ParsedUrl, rules: Sequence[str]) -> Mapping[str, Any]
...
+                        raise REANAFetcherError(
+                            "Git reference discovery output exceeded its limit"
+                        )
+
+        remaining_time = deadline - time.monotonic()
+        if process.poll() is None:
+            if remaining_time <= 0:
+                raise subprocess.TimeoutExpired(command, FETCHER_REQUEST_TIMEOUT)
+            process.wait(timeout=remaining_time)
+    except (OSError, subprocess.TimeoutExpired, REANAFetcherError) as exc:
```
tiborsimko, line 897: **🤖 PR792-11: Unexpected exits can bypass Git child cleanup (NIT)**

The expected timeout, I/O, and output-limit failures all kill and reap the
`git ls-remote` child, but that cleanup lives inside an exception handler for
only three exception types:

```python
except (OSError, subprocess.TimeoutExpired, REANAFetcherError) as exc:
```

The sibling `WorkflowFetcherGit._run_bounded_git()` instead uses
`except Exception` and therefore kills and reaps its child for arbitrary
ordinary exceptions. The new reader narrows that cleanup guarantee to its
enumerated expected failures, even though both helpers own a Git child and a
deadline.

An unexpected exception from the selector/process plumbing, or a
`KeyboardInterrupt`/`SystemExit` during worker shutdown, skips the kill and
wait. The `finally` block closes stdout but can leave the Git child to notice
the broken pipe and exit asynchronously. No ordinary current path appears to
trigger this, so this is a defensive resource-cleanup inconsistency rather
than a demonstrated live leak.

The sibling's broad `except Exception` does not cover
`KeyboardInterrupt`/`SystemExit` either; moving cleanup into `finally` is what
would establish that stronger guarantee.

Suggested fix — make the existing `finally` block terminate and reap any
still-running process before closing stdout, so cleanup covers every exit
path. The `except` branch can then remain focused on translating the expected
errors. A small shared Git-environment or kill-and-reap helper could also avoid
duplicating this setup with `WorkflowFetcherGit._run_bounded_git()`.

reana_server/fetcher.py
tiborsimko: **🤖 Follow-up (2026-08-28): verified fixed at `3c8b1b6`**

`_run_bounded_git_output()` now consumes `git ls-remote` stdout
incrementally, enforces explicit byte and line limits before parsing, and
terminates and reaps the subprocess on the expected limit and timeout paths.
The code comment now correctly distinguishes the glob guard from the
independent stdout bound, and the fake-Git regression test exercises excessive
suffix-matching output. PR792-02 is resolved.

reana_server/fetcher.py
tiborsimko: **🤖 Follow-up (2026-08-28): verified fixed at `3c8b1b6`**

The premature symbolic-`HEAD` return is gone. Longer `HEAD/...` branch and tag
candidates now participate in longest-prefix resolution, while symbolic
`HEAD` remains a post-loop fallback and is preserved when remote discovery
fails. The amendment adds regression cases for the longer-ref and discovery-
failure paths. PR792-08 is resolved.

reana_server/fetcher.py
CameronMcClymont: `_remote_refs()` is now exercised against real `git ls-remote` output, using
local repositories created with the module-level `create_git_repository()`
helper (via a new `create_git_repository_with_refs()` wrapper):

- `test_remote_refs_matches_real_repository_refs()` runs a multi-candidate
  lookup against a repository advertising `main`, `feature/x`, an annotated
  tag `v1.0` (so the real `^{}` line is filtered) and a decoy
  `decoy/refs/heads/main` branch (so the real tail-match is filtered). This
  runs the bounded reader to EOF and lets `_remote_refs()` parse its bytes,
  covering the EOF/unregister branch and the successful byte return.
- `test_remote_refs_missing_repository_raises_fetcher_error()` points at a
  missing repository path, covering the non-zero child exit translation to
  `REANAFetcherError`.
- `test_resolve_provider_tree_path_uses_real_remote_refs()` asserts the
  composition through `_resolve_provider_tree_path()`: the `feature/x` and
  `v1.0` multi-segment lookups, the abbreviated-SHA fallback after successful
  discovery with no match, and the unresolvable-ref error.
- `test_provider_fetcher_abbreviated_sha_folder_after_successful_discovery()`
  covers the post-discovery abbreviated-SHA acceptance branch end to end
  through the fetcher, which previously only had the discovery-failure case.

Coverage of `_run_bounded_git_output()` after the change leaves only the
defensive `stdout is None` guard and the two timing-dependent deadline
branches uncovered; the reader's EOF, success and non-zero exit paths are now
reached by real child processes.

reana_server/fetcher.py
CameronMcClymont: Kill-and-reap moved out of the `except` branch and into the `finally` block of
`_run_bounded_git_output()`, so the Git child is terminated and reaped on every
exit path — unexpected exceptions from the selector/process plumbing and
`KeyboardInterrupt`/`SystemExit` included — before stdout is closed. The
`except` branch now only translates the expected `OSError` /
`TimeoutExpired` failures.

The duplicated setup is factored into two module-level helpers,
`_git_environment()` and `_kill_and_reap()`, now used by both
`_run_bounded_git_output()` and `WorkflowFetcherGit._run_bounded_git()`.

Added `test_remote_refs_reaps_git_on_unexpected_errors()`, which spawns a real
sleeping child, raises an unexpected `RuntimeError` from
`selectors.DefaultSelector`, and asserts the child was reaped. It fails on the
previous revision and passes on the amended one.

PR #793, opened 2026-06-26
feat(rest): support scoped user secret names

### LINKED ISSUE #978: Allow restricting which user secrets are exposed to workflow jobs and interactive sessions
After #805, user secrets can be propagated to interactive sessions. Today, secrets are generally exposed broadly to workflow job pods, Dask workers, and interactive sessions. This is convenient, but it also means pods receive secrets they do not actually need.

## Problem

There is currently no least-privilege way for a workflow author or an interactive-session caller to restrict which user secrets are exposed to a given pod/session.

We should keep the current default behaviour for backward compatibility, but provide an explicit opt-in mechanism to reduce the exposed secret set when desired.

## Proposed solution

Introduce an optional `secret_names` allowlist.

Behaviour:

- If `secret_names` is omitted: keep current behaviour and expose all user secrets.
- If `secret_names` is an empty list: expose no user secrets.
- If `secret_names` is provided: expose only the named secrets.
- If a requested secret name does not exist: reject the request with a clear validation error.

Implementation approach:

- Filter by secret name once in `UserSecrets`, then derive env refs and mounted file items from the filtered subset.
- Preserve the current behaviour for Kerberos/VOMS/Rucio-enabled paths for now, i.e. do not scope secrets there yet.

## User-facing scope

Interactive sessions:

- `reana-client open ... --secret-name NAME` (repeatable)
- `reana-client open ... --no-secrets`
- `reana-client-go open ... --secret-name NAME` (repeatable)
- `reana-client-go open ... --no-secrets`

Workflow authoring:

- Serial: step-level `secret_names: ["alpha", "beta"]` in `reana.yaml`
- Dask: `workflow.resources.dask.secret_names: ["alpha", "beta"]` in `reana.yaml`
- Snakemake: rule-level `resources: secret_names="alpha,beta"`
- Yadage: step `environment.resources` entry such as `- secret_names: ["alpha", "beta"]`

## Out of scope for this issue

- CWL support
- Secret scoping for Kerberos/VOMS/Rucio-enabled jobs

These can be addressed in follow-up work once the base scoping model is agreed.

## Acceptance criteria

- Existing workflows and interactive-session flows remain unchanged when `secret_names` is not provided.
- Interactive sessions can request either an explicit secret allowlist or no secrets at all.
- Serial, Snakemake, Yadage, and Dask authors can specify the secret subset for their jobs/workers.
- Empty allowlists result in no secrets being exposed.
- Unknown secret names produce a clear user-facing error.
- Kerberos/VOMS/Rucio-enabled paths preserve current behaviour.

## Affected components

- `reana-commons`
- `reana-job-controller`
- `reana-workflow-controller`
- `reana-client`
- `reana-client-go`
- `reana-workflow-engine-serial`
- `reana-workflow-engine-snakemake`
- `reana-workflow-engine-yadage`
- `reana-server`

michaelbuchar:
Document interactive-session secret precedence, reserve the internal sidecar manifest name, and validate uploaded secret names before they reach runtime secret reconstruction.

Closes reanahub/reana#978

### DIFF
--- a/reana_server/rest/secrets.py
+++ b/reana_server/rest/secrets.py
@@ -23,24 +23,31 @@ from marshmallow import Schema, validate
 from reana_server.decorators import signin_required
 
 blueprint = Blueprint("secrets", __name__)
+_RESERVED_USER_SECRET_NAMES = {"REANA_USER_SECRETS_TYPES"}
+
+
+def _validate_user_secret_name(secret_name: str):
+    """Reject reserved secret names used by REANA control variables."""
+    if secret_name in _RESERVED_USER_SECRET_NAMES:
+        raise marshmallow.ValidationError(
+            f"Secret name {secret_name} is reserved for internal use."
+        )
 
 
 class AddSecretsBodySchema(Schema):
     """Schema for add_secrets endpoint body."""
 
-    body = (
-        fields.Dict(
-            keys=fields.Str(),
-            values=fields.Nested(
-                {
-                    "value": fields.Str(required=True),
-                    "type": fields.Str(
-                        validate=validate.OneOf(Secret.types), required=True
-                    ),
-                }
-            ),
-            required=True,
+    body = fields.Dict(
+        keys=fields.Str(validate=_validate_user_secret_name),
+        values=fields.Nested(
+            {
+                "value": fields.Str(required=True),
+                "type": fields.Str(
+                    validate=validate.OneOf(Secret.types), required=True
+                ),
+            }
         ),
+        required=True,
     )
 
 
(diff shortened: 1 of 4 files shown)


reana_server/rest/secrets.py
```diff
@@ -153,7 +160,9 @@ def add_secrets(user, overwrite=False):
               }
     """
     json_body = request.json
-    AddSecretsBodySchema().validate({"body": json_body})
+    validation_errors = AddSecretsBodySchema().validate({"body": json_body})
```
tiborsimko, line 163: **🤖 PR793-01: The add-secrets OpenAPI contract omits the new HTTP 400 response (LOW)**

`add_secrets()` now returns HTTP 400 for reserved secret names and for schema errors such as an invalid secret type, but the endpoint's response declaration lists only 201, 403, 409, and 500. The generated server specification and the commons copy therefore omit a response that this PR deliberately introduces. Strict Bravado clients can raise `MatchingResponseNotFound` instead of returning the validation message to `reana-client secrets-add`.

Suggested fix — declare the 400 response and its message schema in the endpoint docstring, regenerate `docs/openapi.json`, propagate the synchronized specification to reana-commons, and add a client-level regression for a rejected secret name or type.

PR #797, opened 2026-08-19
feat(logs): prune expired workflow logs

### LINKED ISSUE #964: introduce system-wide retention policy for workflow engine, service and job logs
### Current behaviour

REANA preserves user runtime workflow logs forever. Every workflow run
accumulates:

- **Engine logs** — the workflow engine pod that orchestrates the run (Serial /
  CWL / Yadage / Snakemake).
- **Service logs** — supporting service pods (e.g. Dask) that run alongside the
  workflow.
- **Job logs** — per-step user job stdout/stderr.

Because there is no automated eviction, the total log volume may grow
unboundedly with the number of workflow runs. On busy instances this is
undesirable: logs are usually only consulted in the days immediately following a
run. They do not need to occupy the database space for the lifetime of the
deployment.

### Expected behaviour

Introduce a system-wide log retention policy. The operator-facing surface
mirrors the existing **workspace retention** feature
(`workspaces.retention_rules.*`) so that administrators encounter a familiar
shape, but the underlying mechanism is deliberately simpler: there is no
per-workflow log retention rule, only a single cluster-wide period.

```yaml
logs:
  retention_rules:
    maximum_period: forever # e.g. "60" days; "forever" preserves current behaviour
    cronjob_schedule: "0 3 * * *" # everyday at 3am (staggered from workspace cleanup)
```

After the configured retention period (e.g. 60 days), a periodic cronjob would
delete:

- engine logs
- service logs
- job logs

for all workflow runs older than the threshold, all the while preserving the
workflow record itself, its UUID, parameters, status (`finished`, `failed`,
`stopped`), start/end timestamps, etc.

GitHub Actions offers a good UX precedent here. Past runs remain listed forever
in the Actions tab with their full status, but opening an old run's logs
displays:

> The logs for this run have expired and are no longer available.

We should do the same in REANA. When a user requests logs for an expired run via
`reana-client logs`, the REANA UI, or the REST API, they should see a clear,
friendly message explaining that the logs have been pruned by the cluster's
retention policy.

#### Suggested scope

- **helm**: add `logs.retention_rules.maximum_period` (default `forever`) and
  `logs.retention_rules.cronjob_schedule`, mirroring the existing
  `workspaces.retention_rules.*` values. Render the period into the reana-server
  pod environment via
  `{{ tpl .Values.logs.retention_rules.maximum_period . | quote }}`, exactly as
  is already done for workspaces.
- **reana-db**: add a single `logs_pruned_at` timestamp column on the workflow
  model, set by the cronjob when it actually deletes a run's logs. No separate
  rule table and no `created → active → pending → applied` state machine: a
  cluster-wide policy has no per-workflow lifecycle to track. Storing the
  exact timestamp gives a clear audit trail and lets the UI tell users
  precisely when their logs were pruned.
- **reana-server**: add a cronjob that walks workflows older than the
  retention threshold whose `logs_pruned_at` is `NULL`, deletes their
  associated engine/service/job logs, and stamps `logs_pruned_at = now()`.
- **reana-ui** and **reana-client**: when fetching logs of a workflow whose
  `logs_pruned_at` is set, surface a clear "the logs for this run were pruned
  by the cluster's retention policy on `<date>` and are no longer available"
  message.

#### Notes

- The default of `forever` ensures this is a strictly opt-in behavioural change
  for existing deployments.
- Composes naturally with the workspace data retention feature: a workflow may
  have its logs pruned but its workspace kept, or vice versa.

michaelbuchar:
Closes reanahub/reana#964


reana_server/reana_admin/log_retention.py
```diff
@@ -0,0 +1,85 @@
...
+
+    service_ids = select(WorkflowService.service_id).where(
+        WorkflowService.workflow_id == workflow.id_
+    )
+    Session.query(ServiceLog).filter(ServiceLog.service_id.in_(service_ids)).delete(
+        synchronize_session="fetch"
+    )
+
+    workflow.logs = None
+    workflow.logs_pruned_at = pruned_at
```
tiborsimko, line 86: **🤖 PR797-01: Pruning ignores OpenSearch while the API hides the retained logs (MEDIUM)**

`prune_workflow_logs()` removes only the database copies: it clears
`Workflow.logs` and `Job.logs` and deletes `ServiceLog` rows. On an
OpenSearch-enabled cluster, however, reana-workflow-controller retrieves engine
and job logs from OpenSearch first through its `build_opensearch_log_fetcher`,
`fetch_workflow_logs`, and `fetch_job_logs` helpers, and falls back to those
database columns only when no OpenSearch result is available.

The pruning job has no corresponding OpenSearch deletion path or OpenSearch
configuration. It nevertheless stamps `logs_pruned_at`, after which
reana-server returns 410 before contacting the workflow controller. Enabling a
finite retention period on such a cluster therefore leaves the principal log
volume in OpenSearch while making those existing logs inaccessible and telling
the user that they were pruned.

Suggested fix — coordinate deletion of the workflow and job documents from
OpenSearch, using workflow UUID and backend job IDs, and stamp
`logs_pruned_at` only after every configured log store has been pruned
successfully. Make the operation idempotent so a partial external-store failure
can be retried safely. If OpenSearch retention is intentionally out of scope,
reject or clearly disable finite log retention when OpenSearch retrieval is
enabled rather than recording a successful prune.

reana_server/rest/workflows.py
```diff
@@ -1725,6 +1754,34 @@ def get_workflow_logs(workflow_id_or_name, user, **kwargs):  # noqa
...
+        try:
+            workflow = _get_workflow_with_uuid_or_name(
+                workflow_id_or_name,
+                str(user.id_),
+                include_shared_workflows=True,
+            )
+        except ValueError:
+            workflow = None
+
+        if workflow and workflow.logs_pruned_at:
```
tiborsimko, line 1769: **🤖 PR797-02: Permitted reana-db versions do not provide `logs_pruned_at` (MEDIUM)**

The new candidate query and REST response access `Workflow.logs_pruned_at`, but
`setup.py` still declares compatibility with `reana-db>=0.95.0a11`. That
released version does not define the column. A valid dependency resolution can
therefore install a model that makes these paths fail with:

```text
AttributeError: type object 'Workflow' has no attribute 'logs_pruned_at'
```

The immediate symptom is the current Python test job, which is the expected
consequence of the documented release ordering. The finding is the declared
lower bound itself: unless it is raised before this PR merges, a clean
installation of the released reana-server can still resolve a reana-db without
the column and fail the same way at runtime, outside CI.

Suggested fix — merge and publish the reana-db change first, then bump both
reana-db lower bounds in `setup.py` to the release containing
`logs_pruned_at`, regenerate `requirements.txt`, and rerun the complete Python
test suite against that published version before merging this PR.

reana_server/reana_admin/cli.py
```diff
@@ -774,6 +778,77 @@ def queue_consume(
...
 
+@reana_admin.command()
+@click.option(
+    "--dry-run",
+    is_flag=True,
+    default=False,
+    help="List workflows whose logs would be pruned without changing them.",
+)
+@click.option(
+    "--force-date",
```
tiborsimko, line 789: **🤖 PR797-03: `--force-date` bypasses the destructive-operation confirmation (MEDIUM)**

The new command accepts an arbitrary UTC `--force-date` and immediately uses it
to calculate the deletion cutoff. A mistaken future date can consequently erase
all persisted workflow, job, and service logs without a confirmation prompt.
The pruning command provides no undo or supported restore path.

The neighbouring `retention-rules-apply` command treats the same option as
dangerous: it prompts before a non-dry-run operation, supports
`--yes-i-am-sure` for automation, and reports the forced time. Omitting that
guard from the more irreversible operation makes the two administrative
interfaces unexpectedly inconsistent.

Suggested fix — mirror `retention-rules-apply`: add `--yes-i-am-sure`, prompt
when `--force-date` is used without it on a non-dry run, and echo the effective
forced time. Cover confirmation, abort, dry-run, and explicit-confirmation paths
in the CLI tests.

reana_server/reana_admin/cli.py
```diff
@@ -774,6 +778,77 @@ def queue_consume(
         consumer.run()
 
 
+@reana_admin.command()
```
tiborsimko, line 781: **🤖 PR797-04: `logs-prune` omits the established `reana_admin` token guard (LOW)**

The other mutating commands that are also run periodically, including
`retention-rules-apply` and `interactive-session-cleanup`, use
`@admin_access_token_option`. `logs-prune` is registered without that decorator,
and among the CronJobs that invoke `flask reana-admin` its job is the only one
that does not load `REANA_ADMIN_ACCESS_TOKEN`. The `resource-quota-update`
CronJob also omits the token, but it runs
`reana-db quota resource-usage-update` and is outside the `reana_admin`
interface.

The token is an administrative consistency and accidental-execution guard, not
a security boundary once a process already has database credentials. Even so,
making the irreversible log-deletion command the exception weakens the uniform
contract of the `reana_admin` interface and makes manual invocation easier to
perform unintentionally.

Suggested fix — add `@admin_access_token_option`, accept and validate the token
argument, and inject the existing admin-access-token secret into the logs-prune
CronJob as is done for `retention-rules-apply`. Add a CLI test showing that a
missing or invalid token prevents pruning.

reana_server/reana_admin/cli.py
```diff
@@ -774,6 +778,77 @@ def queue_consume(
...
+        except Exception as error:
+            Session.rollback()
+            failures += 1
+            click.secho(
+                f"Workflow {workflow_id}: failed to prune logs: {error}", fg="red"
+            )
+            logging.debug(error, exc_info=True)
+
+    action = "would be pruned" if dry_run else "pruned"
+    click.echo(f"{processed} workflow(s) {action}.")
```
tiborsimko: **🤖 PR797-05: The success summary includes workflows whose pruning failed (LOW)**

`processed` is incremented before pruning and is printed as the successful
count regardless of exceptions:

```python
processed += 1
...
except Exception:
    failures += 1
...
click.echo(f"{processed} workflow(s) pruned.")
```

For ten candidates with three transaction failures, the CronJob output reports
both `10 workflow(s) pruned` and `Failed to prune logs for 3 workflow(s)`. The
first statement is false and can mislead an operator or monitoring parser about
how much data was actually removed.

Suggested fix — maintain separate candidate and successful-prune counters, or
print `processed - failures` on the non-dry-run path. Add a test with one
injected pruning failure and assert that the eligible, successful, and failed
counts are internally consistent.

reana_server/reana_admin/log_retention.py
```diff
@@ -0,0 +1,85 @@
...
+    ServiceLog,
+    Workflow,
+    WorkflowService,
+)
+from sqlalchemy import func, select
+
+TERMINAL_WORKFLOW_STATUSES = (
+    RunStatus.finished,
+    RunStatus.failed,
+    RunStatus.stopped,
```
tiborsimko, line 27: **🤖 PR797-06: Deleted workflows are excluded from log retention (LOW)**

The candidate statuses contain only `finished`, `failed`, and `stopped`:

```python
TERMINAL_WORKFLOW_STATUSES = (
    RunStatus.finished,
    RunStatus.failed,
    RunStatus.stopped,
)
```

Deleting a workflow does not remove its database row, jobs, or logs;
reana-workflow-controller marks it as `RunStatus.deleted`. Such workflows can
therefore retain engine, job, and service logs indefinitely even when a finite
cluster-wide policy is enabled. In particular, deleting an old finished run
turns it from an eligible candidate into a permanently excluded one.

Suggested fix — include `RunStatus.deleted` in the candidate statuses. Add
coverage for a finished workflow subsequently marked deleted and for a deleted
workflow without a run-finished timestamp; the existing update/creation-time
fallback can provide the latter's retention age. Rename the constant as well if
including `deleted` makes `TERMINAL_WORKFLOW_STATUSES` misleading.

reana_server/reana_admin/log_retention.py
michaelbuchar: Thanks, OpenSearch deletion is intentionally deferred from this iteration per the agreed scope, and will be handled as a separate follow-up rather than expanding this database-backed retention series.

reana_server/rest/workflows.py
michaelbuchar: Agreed, after reana-db #272 is merged and published, I will bump both reana-db lower bounds, regenerate requirements.txt, and rerun the complete suite before #797 merges.

reana_server/reana_admin/cli.py
michaelbuchar: Added --yes-i-am-sure, interactive confirmation for destructive forced-date runs, effective-time output, and tests for confirmation, abort, dry-run, and explicit bypass

reana_server/reana_admin/cli.py
michaelbuchar: Added the established admin-token guard and wired the existing admin-access-token secret into the logs-prune CronJob, with missing/invalid-token and Helm-rendering coverage

reana_server/reana_admin/cli.py
michaelbuchar: Separated candidate and successful-prune counts so failures are excluded from the success summary, with a regression test covering a mixed success/failure run.

reana_server/reana_admin/log_retention.py
michaelbuchar: Included deleted workflows in retention eligibility, renamed the status constant, and covered both terminal-timestamp and legacy fallback-timestamp cases

reana_server/reana_admin/cli.py
```diff
@@ -774,6 +778,103 @@ def queue_consume(
...
+                    f"as if the current time were {current_time.isoformat()}.\n"
+                    "Are you sure you want to continue?",
+                    fg="red",
+                    bold=True,
+                ),
+                abort=True,
+            )
+        click.echo(f"The current time is forced to be {current_time.isoformat()}")
+
+    cutoff = (current_time - datetime.timedelta(days=LOG_RETENTION_PERIOD)).replace(
```
tiborsimko, line 838: **🤖 PR797-07: Log finalisation can race with permanent pruning (LOW)**

The sweep deletes the logs visible at that instant, stamps `logs_pruned_at`, and then excludes the workflow permanently through `Workflow.logs_pruned_at.is_(None)`. Several writers persist logs after the workflow row is already terminal, and none of them coordinates with that marker:

- `Workflow.update_workflow_status` commits internally, so reana-workflow-controller exposes the terminal status before `_get_workflow_engine_pod_logs` completes and before the caller's later commit persists those logs. The message is acknowledged before processing and terminal workflows are ignored on redelivery, so a controller crash in this window loses the engine logs rather than deferring them; this writer cannot produce an arbitrarily late write.
- reana-job-controller's Slurm monitor sets a job's terminal status before retrieving its final logs over SSH and calling `store_job_logs`. This window is likewise normally short, but retries can extend it.
- The Dask operator inserts `ServiceLog` rows asynchronously from pod-termination events. Its finalizer keeps a terminating pod alive across operator downtime, and the handler is also registered for resume with `deleted=True`, so this writer has no fixed completion bound.

The supported zero-day policy makes the ordinary race straightforward: a workflow terminating immediately before the CronJob cutoff can be selected while final log capture is still in flight. Longer policies greatly reduce the exposure, but do not eliminate the Dask case when an operator outage exceeds the retention period. A later write repopulates workflow, job, or service logs on an already-stamped workflow; subsequent sweeps skip it and the API keeps returning 410, leaving those logs hidden and retained with no re-prune path.

Suggested fix — clamp the cutoff so that it does not reach the present, for example by applying a grace interval of about an hour. This still honours the intent of a zero-day policy and covers the expected completion window of the controller and Slurm writers. Separately, make delayed Dask writes recoverable: either resolve the owning workflow and honour `logs_pruned_at` with suitable transaction synchronisation, or make the sweep idempotently reconcile stamped workflows that acquire logs again. If the Dask-side change is outside this series, please track it as a linked follow-up. Add coverage for a recently terminated workflow under a zero-day policy and for a service log arriving after the initial pruning pass.

reana_server/rest/workflows.py
tiborsimko: **🤖 PR797-02 follow-up (2026-08-30): remains open at f6e7722**

Thanks for confirming the intended release order. Waiting for reana-db #272 to be published before updating the dependency is the expected workflow; the resulting failures against the currently released model do not block review of the functionality itself and can be handled as part of the release sequence.

The declared lower bounds still permit `reana-db` 0.95.0a11, which does not provide `Workflow.logs_pruned_at`. The checked-in `requirements.txt` is already out of step with those bounds on master and currently resolves the older 0.95.0a10 build. That drift is pre-existing rather than introduced by this PR, but a released reana-server installation can still select an incompatible model and fail at runtime until both declarations are updated.

Suggested remaining fix — after reana-db #272 is published, raise both `reana-db` lower bounds in `setup.py`, regenerate `requirements.txt`, and rerun the Python suite before releasing this change.

reana_server/reana_admin/cli.py
tiborsimko: **🤖 PR797-05 follow-up (2026-08-30): verified fixed at f6e7722**

The command now tracks candidates, successful prunes, and failures separately. A real run reports only the successful count as pruned, while a dry run reports the candidate count.

The mixed-success regression test injects a pruning failure and verifies that the failed workflow is not included in the success summary. No residual concern remains for this finding.

reana_server/reana_admin/log_retention.py
tiborsimko: **🤖 PR797-01 follow-up (2026-08-30): remains open at f6e7722**

Thanks for clarifying that OpenSearch deletion is intentionally deferred from this series. Limiting this implementation to database storage is a defensible scope decision, but the user-visible inconsistency remains: the sweep stamps the workflow and the API returns 410 even though engine and job logs are still retained in OpenSearch and would otherwise be retrievable.

I could not find an operator-facing warning on the new Helm values, a configuration guard, or a linked public follow-up issue. As a result, enabling the policy on an OpenSearch-backed cluster still silently fails to reclaim the principal log storage while reporting that those logs were pruned.

Suggested remaining fix — either include OpenSearch deletion in this implementation, or document the limitation prominently alongside the new `logs.retention_rules.*` values and open a linked public follow-up issue. A Helm validation guard or runtime warning when log retention and OpenSearch are enabled together would further prevent operators from relying on a policy that does not cover their active log store.

reana_server/reana_admin/cli.py
tiborsimko: **🤖 PR797-03 follow-up (2026-08-30): verified fixed at f6e7722**

`logs-prune` now mirrors the established destructive-operation guard: `--force-date` prompts for confirmation unless the invocation is a dry run or supplies `--yes-i-am-sure`, and the effective forced time is shown to the operator.

The new test covers confirmation, abort, dry-run bypass, and explicit confirmation bypass. No residual concern remains for this finding.

reana_server/reana_admin/log_retention.py
tiborsimko: **🤖 PR797-06 follow-up (2026-08-30): verified fixed at f6e7722**

`RunStatus.deleted` is now included in the renamed `LOG_RETENTION_WORKFLOW_STATUSES`, so explicitly deleted workflows participate in the same retention sweep. The tests cover deleted workflows with both a terminal timestamp and the `updated` fallback.

Workflow restarts create a cloned workflow row, so the pruning marker on an earlier run is not carried into the restarted run. No residual concern remains for this finding.

reana_server/reana_admin/cli.py
tiborsimko: **🤖 PR797-04 follow-up (2026-08-30): verified fixed at f6e7722**

`logs-prune` now uses `@admin_access_token_option`, so direct invocations pass through the same token validation as the other periodically scheduled `flask reana-admin` maintenance commands. The corresponding CronJob at reana #996 head `a494100` loads `REANA_ADMIN_ACCESS_TOKEN` from the established Secret key.

The server tests cover missing and invalid tokens, and the Helm rendering test verifies the Secret reference. No residual concern remains for this finding.

PR #802, opened 2026-08-20
fix(workspace): preserve shared workspace permissions

### LINKED ISSUE #801: workspace: create server-seeded files with shared permissions
### Current behaviour

After server-side workflow creation was introduced in #794, files copied by
`seed_workspace()` are created with restrictive permissions:

- directories: `0700`
- files: `0600`

They are owned by `root:root`, while workflow-engine containers run as UID
`1000`, GID `0`. Consequently, launched workflows cannot read their workflow
specification files.

### How to reproduce

Launch the Snakemake H→4ℓ example:

```text
https://localhost:30443/launch?url=https%3A%2F%2Fgithub.com%2Freanahub%2Freana-demo-cms-h4l&name=reana-demo-cms-h4l&specification=reana.yaml
```

The workflow engine fails with:

```text
[Errno 13] Permission denied:
'/var/reana/users/.../workflows/.../workflow/Snakefile'
```

The workspace permissions are:

```text
drwx------ root root workflow/
-rw------- root root workflow/Snakefile
```

Running `chmod -R g+rwX <workspace>` and restarting the workflow fixes the
problem; the restarted workflow finishes successfully.

### Expected behaviour

Files and directories seeded by reana-server should use shared-workspace
permissions consistent with `REANA_WORKFLOW_UMASK=0002`, allowing runtime
containers in the shared group to access them.

### Suggested fix

Update `seed_workspace()` so that newly created directories and files use the
normal shared-workspace modes while retaining the existing symlink and
exclusive-creation protections.

Add tests that verify:

- seeded directory and file modes;
- accessibility to the runtime UID/GID;
- successful creation of a multi-file Snakemake workflow through `/launch`;
- the same permissions for other callers of `seed_workspace()`, such as GitLab
  workflow creation.

tiborsimko:
Create seeded directories and files with conventional base modes so that the process umask keeps them accessible to runtime containers.

Route uploaded specification bundles through the same validated-member seeding path. The bounded copy temporarily duplicates staged data while retaining exclusive, no-follow destination creation.

Add direct, create, and launcher-level regression coverage for the resulting shared workspace permissions.

Closes #801

### DIFF
--- a/reana_server/rest/workflows.py
+++ b/reana_server/rest/workflows.py
@@ -100,7 +100,6 @@ from reana_server.utils import (
     get_quota_excess_message,
     get_workspace_retention_rules,
     is_uuid_v4,
-    mv_workflow_files,
     prevent_disk_quota_excess,
     publish_workflow_submission,
 )
@@ -576,9 +575,9 @@ def validate_workflow_specification(user):  # noqa
                 jsonify({"message": "No specification bundle files were provided."}),
                 400,
             )
-        abs_dir, rel_path, _bundle_bytes, _legacy_parameters = _stage_validation_bundle(
-            request.files
-        )
+        bundle = _stage_validation_bundle(request.files)
+        abs_dir = bundle.absolute_path
+        rel_path = bundle.relative_path
         reana_yaml_path = next(
             (
                 os.path.join(abs_dir, name)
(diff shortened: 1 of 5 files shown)


CameronMcClymont (approved):
LGTM and tested on my local cluster 🦭

